Issue #95, opened 2021-11-04
[dev] Add Prettier to Project


Renji-XD:
Hi,

was playing a bit with extension + code and noticed the default formatter of my ide reformats basically everything. 

To encourage pr contributions it maybe makes sense if you could add prettier+ whatever config you think makes sense to all (sub)modules and recommit all (reformatted) src files.. otherwise either you on the review or the dev while working on changes will get crazy ;)

killergerbah:
Closed with https://github.com/killergerbah/asbplayer/pull/96 from @Renji-XD

PR #96, opened 2021-11-04
[dev] added prettier and base config

Renji-XD:
Simply added Prettier as described in #95 with some basic config changes.

width and quotes were basically completly mixed everywhere (even inside the same files) so i just choose one value for it


killergerbah:
Thanks for doing this. I've been foolishly using an editor with poor support for JavaScript projects which led to the mess that is asbplayer.

I'd like to set up prettier in my own development environment before merging this. Will come back in a bit.

Issue #39, opened 2021-05-13
Directly integrate with Netflix to obtain subtitle data


killergerbah:

killergerbah:
Closed with https://github.com/killergerbah/asbplayer/pull/107 from @Renji-XD

PR #107, opened 2021-12-13
Allow Sync of Subtitles from synced videos

Renji-XD:
Best Effort Implementation for #39 

Adds new Keybind, new Options and simple Event Mechanismum + UI Select Dialog for an Attempt of syncing Subtitles from a video to the asbplayer client (with page object dependant implementations for netflix and /watch path of youtube)

Client (requires repub to github pages):
- Propagation of theme without synced video (on extension installed callback)
- Propagation after tab was added to registry to trigger bind (and avoid errors in case the page loads quicker than the state update)

New Options:
- Preferred Language Codes: Comma separated list of language codes which can be used in conjunction with other options to preselect / auto sync subtitles for specific languages
- Autoselection of prefered Language Code (first track with matching code will be selected). Option can be useful if there are multiple tracks for a language so an User would not need to clear the language codes all the time
- Auto Sync: automatically triggers the subsync process without User Input
- Keep Pause: can be used to continue the video without the need of manually press anything or the play button

Example of manual selection:
https://imgur.com/a/pAWafEK

Example of full automatic sync:
https://imgur.com/a/cGn0Xgu

Version of extension not increased as i don't know if / when you want to release a new one


killergerbah:
Thanks for the contribution. I will need some time to review this.

client/src/components/App.js
```diff
@@ -448,7 +449,12 @@ function App() {
     const handleClipAudio = useCallback(
         async (item) => {
             try {
-                const clip = await audioClipFromItem(item);
+                const clip = await audioClipFromItem(
```
killergerbah: It would be great if you can rebase on killergerbah/main to pick up this change so your changes merge cleanly as one commit.

extension/src/background.js
```diff
@@ -89,3 +90,14 @@ chrome.commands.onCommand.addListener((command) => {
         }
     });
 });
+
+function sendMessageToCurrentVideoTab(tabs, message) {
```
killergerbah: I think a more accurate name might be `sendMessageToActiveVideoElements` as this function sends the message to all video elements currently in the tab registry. In particular, it's possible for the user to have more than one tab open which may each be synced to different asbplayers.

extension/src/pages/netflix-page.js
```diff
@@ -0,0 +1,137 @@
...
+                .map((track) => {
+                    return {
+                        label: `${track.bcp47} - ${track.displayName}${
+                            'CLOSEDCAPTIONS' === track.rawTrackType ? ' [CC]' : ''
+                        }`,
+                        language: track.bcp47.toLowerCase(),
+                        url: storedTracks.get(track.trackId),
+                    };
+                });
+            response.filename = `${titleId}.nfvtt`;
```
killergerbah: It would be great if we can use a human readable title for the file name so that the `Source` field for flashcards can continue to be useful.

extension/src/pages/youtube-page.js
```diff
@@ -0,0 +1,54 @@
+document.addEventListener(
+    'asbplayer-get-external-subtitles',
+    async () => {
+        const response = { error: '', filename: '', subtitles: [] };
+
+        try {
+            const playerContext = await fetch(window.location.href)
```
killergerbah, line 7: Why is this `fetch` necessary? It seems possible to produce `playerScript` by accessing `document.body` directly.

extension/src/pages/youtube-page.js
```diff
@@ -0,0 +1,54 @@
...
+                    }
+
+                    return new Function(`${playerScript.textContent}; return ytInitialPlayerResponse;`)();
+                });
+
+            if (!playerContext) {
+                throw new Error('YT Player Context not found...');
+            }
+
+            response.filename = `${playerContext.videoDetails?.title || 'youtube'}.ytxml`;
```
killergerbah: Is there any reason why we can't use `document.title` as a reasonable default?

extension/src/services/SubtitleSyncSelectContainer.js
```diff
@@ -0,0 +1,268 @@
...
+        this.lastError = error;
+        const client = await this.setUp();
+        return client.updateState({
+            alertOpen: true,
+            alert: this.lastError,
+            severity: 'error',
+        });
+    }
+
+    bufferToBase64(buffer) {
```
killergerbah: You can import this from `./Base64`

client/src/services/SubtitleReader.js
```diff
@@ -44,6 +61,48 @@ export default class SubtitleReader {
...
+                });
+            }
+
+            return subtitles;
+        }
+
         throw new Error('Unsupported subtitle file format');
     }
+
+    fixRTL(line) {
```
killergerbah: It's currently conventional in this codebase to prefix private methods with an underscore.

killergerbah (changes requested):
Thanks for your patience - I was traveling for the holidays and didn't have an opportunity to read through you work carefully. I appreciate your spending the time to work on this much-needed (and tedious) feature.

Before merging this I would like to propose the following high-level changes:

- Instead of creating a new `Select Subtitle` dialog, use the existing `Set Video Name` dialog that is currently used for syncing an empty subtitle file. Currently, both of these dialogs are used to sync _something_ to an asbplayer tab. Therefore I think it makes sense to simply add the new functionality to the existing dialog to simplify UX and to reduce the number of keyboard shortcuts that asbplayer needs to support. Obviously, we will want to change the name of the existing dialog to something else.

- I think the new `tabs-updated` message may not be necessary. I realize there is a race condition between the `TabRegistry` being updated and the asbplayer tab actually being ready to receive a sync request, but I believe this can be fixed in a way that is invisible to the video content script, in a separate change.

- The feature doesn't seem to work unless an asbplayer tab is already open. At first glance this seems to happen because the video needs to receive a `tabs-updated` message first. As mentioned above, I propose removing `tabs-updated` for now.

Please let me know what you think.

Renji-XD:
hi,

no worries i am out for the next ~1,5 week by myself. Will leave first high level comments and then come back when i find some time in the new year.

- Merged Elements: can make sense i guess - name may even be reused for the subtitle filename instead. Need to check the implementation and if there are issues with the select one - could maybe be renamed to something like VideoDataSyncX etc. Only thing functional wise which comes to my mind right now is the fact it may not work well with a full automatic way as i thought of it. First there is a user interaction required to select the element (which fades out) which can be missed if triggered automatically / can be confusing for a user and second the dialog itself would be skipped with respective found languages (here the automatic naming could be used but it means if someone wants to have a manual name he could not use auto sync). Think also the error messages are currently always displayed on opening

- tabs event: On Auto / quick keybind use it was basically always running into the tab not found error. If the event is not applicable i need to recheck this part of the implementation / may need guidance if you already found a way to avoid the race condition as i didn't go that deep into that part of the code

> The feature doesn't seem to work unless an asbplayer tab is already open. 
Ya i think you are right.. didn't had this case in mind.. needs to work similiar to the video name element which opens the tab with passed data or similiar.

Have a good start in to the next year - will come back later to your points when things settled down ;)

extension/src/pages/netflix-page.js
Renji-XD: if i remember correctly they don't pass a human readable name around (and window title is also just netflix). Alternative would be to intercept their metadata api call / fireing one manually which would introduce some "longer" waittime and required logic to make sure both - subtitle and name are capture before dialog is shown... up to you - let me know. Maybe partially solved if the dialaogs are merged and a video name prop is passed

extension/src/pages/youtube-page.js
Renji-XD: As best effort i didn't deep dived into their player or api stuff.. but this script is only updated on the first opened video.. if you go an other one they just replace the video source but everything else stays with data from the old video.. guess they do some stuff internally in their player lib.. therefore the request to get the current video data

extension/src/services/SubtitleSyncSelectContainer.js
Renji-XD: if i remember correctly i initially did it and it brought a rat tail of webpack / build stuff and errors due to the fact it is not really touched from it yet. as it was 5 lines i just copied it so i i wont need to mess around with changes on the build .. can recheck if i maybe did something wrong

extension/src/pages/netflix-page.js
killergerbah: I see, yeah, hooking into an API call seems kind of heavy...

I hunted around the subadub repo for a way to do this and found this: https://github.com/rsimmons/subadub/issues/32
Looks like the necessary data is hidden kind of deep inside the Netflix player API. I was able to get something close to what subadub does in the developer console:
```
// Assuming 'track' is already defined above
session = netflix.appContext.state.playerApp.getAPI().getSessionDataBySessionId('watch-c10a86cf-2bb0-4d71-b3b6-9eeff4dc9a5f')
videoId = String(session.titleVideoId)
video = netflix.appContext.state.playerApp.getAPI().getVideoMetadataByVideoId(videoId).getCurrentVideo();
`${video.getTitle()}.S${video.getSeason()._season.seq}.E${String(video.getEpisodeNumber()).padStart(2, '0')}.${track.bcp47}`
```

extension/src/services/SubtitleSyncSelectContainer.js
killergerbah: Ah interesting, okay. In theory I think `Base64` isn't importing anything else so it shouldn't be dragging unwanted dependencies with it but I could be wrong.

killergerbah:
Cool, enjoy your time off!

Responding to your thoughts:

> Merged Elements: can make sense i guess - name may even be reused for the subtitle filename instead. Need to check the implementation and if there are issues with the select one - could maybe be renamed to something like VideoDataSyncX etc. Only thing functional wise which comes to my mind right now is the fact it may not work well with a full automatic way as i thought of it. First there is a user interaction required to select the element (which fades out) which can be missed if triggered automatically / can be confusing for a user and second the dialog itself would be skipped with respective found languages (here the automatic naming could be used but it means if someone wants to have a manual name he could not use auto sync). Also the messages maybe in some cases not optimal anymore

- Currently, a user interaction is required to select an element if there is more than one video element on the page. We would probably need to improve this logic for pages where we know what the video element is supposed to be. However, it seems that Netflix and YouTube only have one video element so we may not need to do anything. It should be noted that with the current approach we will may an issue for pages that have multiple video elements as the it looks like the auto-syncing will attempt to sync all of the video elements on the page (and I would guess the sync requests may compete for the same asbplayer tab).
- You're right, some messages will not be optimal. In fact I think all of them will become unnecessary, namely the new command `sync-external-subtitles` will and as mentioned above the `tabs-updated` should no longer  be necessary. I may be missing others.

> tabs event: On Auto / quick keybind use it was basically always running into the tab not found error. If the event is not applicable i need to recheck this part of the implementation / may need guidance if you already found a way to avoid the race condition as i didn't go that deep into that part of the code

It makes sense you were running into that issue very frequently as tabs are published to asbplayer tabs automatically only every 1 second, and I think there's some extra time needed for the tab data to be received if the asbplayer tabs have been backgrounded for a while. I would guess that retrieving the subtitle data may often take less than that amount of time. This is actually already an issue when dragging-and-dropping subtitles (or when syncing empty subtitle files) as there is a small window of time after auto-opening an asbplayer tab where the asbplayer instance isn't ready to receive the tab data. We try to work around this by force-publishing the data when auto-opening a tab but it doesn't seem to always work.

Since this issue is breaking auto-sync so often we could fix it in this pull request. However, since this race-condition is already a user-facing bug, I think it would be an easy enough change to make in a separate commit, as well.

For reference, these are my thoughts on a fix that would be isolated to `TabRegistry` and the client:
Once the `sync` message is dispatched, we can leave the forwarding of the `sync` message to an appropriate tab entirely to the background handler. Therefore, in theory there should be a way to avoid this problem with some changes only to `TabRegistry`. At a high level, I think we could do something like this:
1. Keep the `tabs-updated` message but create a background handler that receives the message instead of forwarding it all the way to the video elements.
2. Modify `TabRegistry.videoElements` values to include the `src` strings that each video element has actually synced.
3. Modify `TabRegistry.findAsbplayerTab` to include a check for the `src` of the video element requesting the sync.

killergerbah:
@Renji-XD Let me know if you want to tackle the tab data race condition problem. I can also fix this issue in the main branch, if you prefer, as it's already a bug.

Renji-XD:
@killergerbah if you feel like it and have time go for it - will be anways a couple of days before i am back and you also requested to rebase already the forked repo.. i can then check later what you have changed and how to use it for this change (hopefully without much merge conflicts / effort : > )

killergerbah:
I've merged the fix for the race condition into `main`. It should be a clean rebase for you but you will still to remove the `tabs-updated` message publishing and handling manually. You should be able to trigger the subtitle fetch without waiting and it _should_ work because of the fix.

Renji-XD:
Ya, rebase done without anything on top required.

So related to your change - tbh i did not checked the old vs new implementation etc. in detail but i just quickly wanted to test a rebuild before starting any further changes (maybe next week)... 

so apparently the error is gone but now it simply opens every time a new asbplayer tab in case there is an early sync request due to the fact that the received tabs is not yet updated at the time of passing this line https://github.com/killergerbah/asbplayer/blob/main/extension/src/services/TabRegistry.js#L77

not sure if i miss sth or if that was sth to be somehow handled by me?

killergerbah:
> so apparently the error is gone but now it simply opens every time a new asbplayer tab in case there is an early sync request due to the fact that the received tabs is not garantueed to be updated yet by the a heartbeat at the time of passing this line https://github.com/killergerbah/asbplayer/blob/main/extension/src/services/TabRegistry.js#L67

Okay, so this is likely a new bug that comes up because of the nature of the auto-sync.

- The original fix worked for the case when the page loads and the user drags-and-drops a subtitle file, or tries to sync an empty one. Usually this manual dragging-and-dropping process takes more than a few seconds, during which time it is likely the video elements of the page will be published to all of the asbplayers.

- With your change I am guessing that the subtitles are being retrieved so quickly that it's possible they are being sent before the video elements are synced, as this can take up to a second or so (some amount of time to the heartbeat, and up to a second to publish to the video elements to the asbplayers).

- Just to cover all bases, there is another reason a new tab may open even if an asbplayer tab already exists. If a tab is backgrounded for a long time, Chrome stops running the JS as frequently (or at all) in the tab. Thus the heartbeat stops running and the tab disappears from the tab registry. So far I haven't been able to find a solution around this, but this is a separate issue and probably shouldn't be dealt with in this change as this problem has always existed.

So, assuming that the issue is occurring even when the background tab is sending a heartbeat, I believe we will want to fix the race condition described two bullet points above in your change. I'm not immediately sure what a good solution is right now, but if it's possible it would be great to keep all of the tab management/finding logic in the background scripts to avoid any special-case logic.

Renji-XD:
a potential workaround could may be to move the sync message from synchandler to the tabregistry and just queue the tab, src and subtitle data to the registry in case the tab was not found (while opening a new asbplayer tab in case there is none to start the heartbeat) .. then call a flush method in the heartbeat handler which goes over the queue items and recall the findAsbplayerTab method (up to e. g. 5 ticks before the item is removed with an error)

Under the Assumption that for all normal cases the tab is found immediately and that there will not be that many asbplayer tabs and queued Items otherwise i don't think it would be some heavy footprint even with the 1s interval execution.. not sure if that may bring implications or race conditions.. i quickly tested it and at least the issue i noticed is gone

killergerbah:
@Renji-XD Is this a correct interpretation of your solution?
1. Content script sends sync request.
2. Sync request handler calls `findAsbplayerTab`.
  a. If `findAsbplayerTab` fails and `TabRegistry.asbplayers` map is NOT empty, queue request in background.
  b. If `findAsbplayerTab` fails and `TabRegistry.asbplayers` map IS empty, open a new tab.
  c. If `findAsbplayerTab` succeeds, stop here.
3. asbplayer heartbeat handler will call  `findAsbplayerTab` for all the requests in the queue.

I'm not sure if I interpreted this part of your solution correctly:

>  (while opening a new asbplayer tab in case there is none to start the heartbeat)

as I'm assuming we do not want to open an asbplayer tab every single time.

Renji-XD:
yes i think you got what i mean. that was the code i was using to play around

image

flush is the iteration function called by the asbplayer heartbeat handler on respective interval tick

killergerbah:
Okay thanks for clarifying.

I'd like to suggest the following modifications to your approach:
1. Instead of using `syncQueue`, which is shared state, use the recursive `_anyAsbplayerTab` which I believe already implements the behavior you're proposing and keeps all required state local to the function call.
2. Since `_anyAsbplayerTab` will recurse on its own every second, you don't need to change the asbplayer heartbeat handler to do anything new.

I believe with this approach the diff will be small. It should look something like this (untested):
```
    async findAsbplayerTab(videoTab, videoSrc) {
        let chosenTabId = null;
        const now = Date.now();
        let min = null;

        for (const tabId in this.asbplayers) {
            const asbplayer = this.asbplayers[tabId];

            if (this._asbplayerReceivedVideoTabData(asbplayer, videoTab, videoSrc)) {
                const elapsed = now - asbplayer.timestamp;

                if (min === null || elapsed < min) {
                    min = elapsed;
                    chosenTabId = tabId;
                }
            }
        }

        if (chosenTabId) {
            return chosenTabId;
        }

        if (Object.keys(this.asbplayers).length > 0) {
            return new Promise((resolve, reject) => {
                this._anyAsbplayerTab(videoTab, videoSrc, resolve, reject, 0, 5);
            });
        }

        return this._createTab(videoTab, videoSrc);
    }
```

Renji-XD:
yes you are right.

do you mind if i remove the min/elapsed part for now (looks like it is anyways always passing the condition) because then the initial lookup could also reuse the _anyAsbplayerTab method

<details>
<summary>Example</summary>

```javascript
    async findAsbplayerTab(videoTab, videoSrc) {
        const chosenTabId = await new Promise((resolve, reject) =>
            this._anyAsbplayerTab(videoTab, videoSrc, resolve, reject, 0, 1)
        ).catch(() => {});

        return (
            chosenTabId ??
            new Promise(async (resolve, reject) => {
                if (!Object.keys(this.asbplayers).length) {
                    await new Promise(async (resolve) => {
                        chrome.tabs.create(
                            {
                                active: false,
                                selected: false,
                                url: (await this.settings.get(['asbplayerUrl'])).asbplayerUrl,
                                index: videoTab.index + 1,
                            },
                            resolve
                        );
                    });
                }
                this._anyAsbplayerTab(videoTab, videoSrc, resolve, reject, 0, 5);
            })
        );
    }
```
</details>

killergerbah:
> do you mind if i remove the min/elapsed part for now (looks like it is anyways always passing the condition) because then the initial lookup could also reuse the _anyAsbplayerTab method

I think in it's beneficial to keep the original for loop even if it seems to repeat some code. Trying to reuse `_anyAsbplayerTab` changes the behavior of the code so that it's calling an async function to perform a calculation that can be done synchronously.

Alternatively, if you want to reduce duplicated code in that class then you could extract the loop into its own function and instead reuse it from `_anyAsbplayerTab`.

Renji-XD:
ok, so i will try to do the changes this week.

To multiple Videos per Site i think it is a fair assumption that this feature will only work on single video pages (which should cover most of platforms). It is anways not garantueed you will have a site implementation which is able to identify which video binding belongs to which data etc. and also it should not be that big pain for an user to go to the main page of the video platform in case of embeddables or similar in my opinion. Worst case he can still drag&drop subtitle files there.

To sum up changes:
- code changes as per pr comments
- small fix to the tabregistry
- remove subsync/video-name container / ui elements and merge them to a new VideoDataSync one which allows to set name and select subtitle for sync (which is set to off on non supported sites). Any Error Message will be shown as last element above buttons

I will remove the alert and make the dialog nonblocking. What is your preference in terms of behavior of data loading on subsync supported sites:
1. Display Dialog Title and Spinner + disabled Elements until Subtitle Data is loaded (worst case name cannot be set independantly in case of missing response - timeout is not a good alternative i think as it cannot be predicted how long the subtitle will be loaded e. g. on a slow vpn connection)
2. Spinner besides disabled subtitles Input with off as selected value (as long as loading), other elements enabled, name will be replaced by returned data (can be confusing if user started to type before response returned and then his input is gone)
3. Spinner besides disabled subtitles Input with off as selected value (as long as loading), other elements enabled, name will be only replaced by returned data if nothing entered yet (some users may never know there is an autofill when they always start typing)

think that was all we discussed so far

killergerbah:
> To multiple Videos per Site i think it is a fair assumption that this feature will only work on single video pages (which should cover most of platforms). It is anways not garantueed you will have a site implementation which is able to identify which video binding belongs to which data etc. and also it should not be that big pain for an user to go to the main page of the video platform in case of embeddables or similar in my opinion. Worst case he can still drag&drop subtitle files there.

Sure, I agree that we don't need to address the multiple-videos-per-site possibility for now. However, I think it would be ideal if subtitle fetching is triggered at a page level rather than at a `Binding` (per-video-element) level as the extension could start behaving badly if YouTube suddenly decided to add a second video element (for example, a separate video element for ads).

> code changes as per pr comments
small fix to the tabregistry
code removal for new keybind (will be the existing toggle-video-select)
remove subsync/video-name container / ui elements and merge them to a new VideoDataSync one which allows to set name and select subtitle for sync (which is set to off on non supported sites). Any Error Message will be shown as last element above buttons

In the interest of simplifying UX, I would like to propose:
- Remove "keep pause" option. We already have behavior for this that is implemented by the Anki dialog. That behavior is: put video back whatever state it was in before the dialog was requested. I believe this is the most reasonable behavior and dialogs should consistently implement it. Currently, the video-select dialog does NOT implement that behavior (bug) so implementing the correct one could be a separate change.
- Keep the style of English used throughout the software consistent. I noticed in documentation and help you are capitalizing nouns - please stay consistent with the rest of the software.
- (P2) Simplify the new subtitle sync options. I believe we only new option we need is the "auto sync" toggle. The selection of the language does not need to be configured as this can be inferred from what the user actually chooses to sync. The dialog can be displayed if the last synced language code is not available for some reason. If the user for some reason wants to change the language code they always have the option of opening the dialog manually through `Ctrl+Shift+F`, but I think it's safe to assume that this case should be very rare.
- (P2) As mentioned above, it would be ideal if subtitle fetching is triggered at a page level rather than at a `Binding` (per-video-element) level as this could protect us from bugs and save us time in case we ever want to implement auto sync on pages that have multiple video elements (like Amazon Prime).

> I will remove the alert and make the dialog nonblocking. What is your preference in terms of behavior of data loading on subsync supported sites (think 3 is potentially the best):

Yes I agree 3 sounds the most reasonable.

Renji-XD:
i added one more change to how bindings are added / removed which you may check if you can think of implications. 

I had my tests running with bookmarked playlists / videos but after updating the pr i kept the extension running and later i noticed an odd case on youtube. When you enter via the main page and hover a video they actually insert an video element. When unhovering they keep the element (with unset src) and carrying it across all navigations which lead to the odd behavior that on triggering a sync you had to select the video element manually again as there were now multiple bindings on one page - also it occurs that the image sometimes overlapped each other so that you could not interact with it etc.

I added the condition that the video needs to have an actual value for the src or a source children with a src

Renji-XD:
i added one more change to how bindings are added / removed which you may want to 
check if you can think of implications. 

I had my tests running with bookmarked playlists / videos but after updating the pr i kept the extension running and later i noticed an odd case on youtube. When you enter via the main page and hover a video they actually insert an video element. When unhovering they keep the element (with unset src) and carrying it across all navigations which lead to the odd behavior that on triggering a sync you had to select the video element manually again as there were now multiple bindings on one page - also it occurs that the image sometimes overlapped each other so that you could not interact with it etc. When entering youtube via a video / playlist / search or reloading the tab this behavior was back to normal.

I added the condition that the video needs to have an actual value for the src or a source children with a src in order to be valid for having a binding

README.md
```diff
@@ -51,7 +56,7 @@ The below information can also be found in asbplayer by clicking on the question
 |Ctrl+Shift+Q|Copy current subtitle and open Anki export dialog|
 |Ctrl+Shift+Z|Copy current subtitle (streaming video in another tab)|
 |Ctrl+Shift+X|Copy current subtitle and open Anki export dialog (streaming video in another tab)|    
-|Ctrl+Shift+F|Select video element to mine with no subtitle file (streaming video in another tab)|      
+|Ctrl+Shift+F|Select video element to mine with no subtitle file / sync data for supported sites (streaming video in another tab)|
```
killergerbah: "Select video element to mine with no subtitle file, or available subtitles for supported sites" might be a bit more clear.

client/src/services/SubtitleReader.js
```diff
@@ -10,16 +10,33 @@ export default class SubtitleReader {
     }
 
     async _subtitles(file, track) {
+        const tagRegex = RegExp('</?([^>]*)>', 'ig');
```
killergerbah: Can you move this to the top level of the file or instantiate and cache lazily? Otherwise the regex is being compiled every time `_subtitles` is called

client/src/services/SubtitleReader.js
```diff
@@ -44,6 +61,48 @@ export default class SubtitleReader {
...
+        } else if (index2 > -1) {
+            newLine = line.substring(0, index2) + '\u202b' + line.substring(index2 + 5) + '\u202c';
+            return this._fixRTL(newLine);
+        }
+
+        return line;
+    }
+
+    _decodeHTML(text) {
+        const e = document.createElement('div');
```
killergerbah: Can this element be cached?

extension/src/handlers/video/SyncHandler.js
```diff
@@ -12,7 +12,10 @@ export default class SyncHandler {
     }
 
     async handle(request, sender) {
-        let chosenTabId = await this.tabRegistry.findAsbplayerTab(sender.tab, request.src);
+        let chosenTabId = await this.tabRegistry.findAsbplayerTab(sender.tab, request.src).catch((error) => {
```
killergerbah: Nitpick: you use `try` `catch` instead of calling `catch` on the returned promise?

extension/src/popup.html
```diff
@@ -1,75 +1,89 @@
...
+    </div>
+    <hr>
+    <div>
+        <input class="checkbox-color" id="subsDragAndDropInput" type="checkbox">
+        <label id="subsDragAndDrop" for="subsDragAndDropInput">Dropping subs into video syncs with asbplayer</label>
+    </div>
+    <hr>
+    <div>
+        <input class="checkbox-color" id="autoSyncInput" type="checkbox">
+        <label id="autoSync" for="autoSyncInput">Enable automatic sync</label>
```
killergerbah: "Automatically sync any available subtitles" might be more concise.
"Enable" is implied since the option is a toggle.

extension/src/popup.js
```diff
@@ -103,7 +110,7 @@ document.addEventListener('DOMContentLoaded', async (e) => {
                 } else {
                     help.push(
                         c.shortcut +
-                            " enables selection of a video to mine without a subtitle file. Once a video is selected, either of the 'copy' shortcuts will start and stop recording."
+                            " enables selection of a video to mine without a subtitle file / sync data. Once a video is selected, either of the 'copy' shortcuts will start and stop recording."
```
killergerbah: "... to mine with no subtitle file, or available subtitles for supported sites. When mining with no subtitle file, either of the 'copy' shortcuts..."

extension/src/services/TabRegistry.js
```diff
@@ -79,15 +79,20 @@ export default class TabRegistry {
...
-                {
-                    active: false,
-                    selected: false,
-                    url: (await this.settings.get(['asbplayerUrl'])).asbplayerUrl,
-                    index: videoTab.index + 1,
-                },
-                (tab) => this._anyAsbplayerTab(videoTab, videoSrc, resolve, reject, 0, 5)
-            );
+            if (!Object.keys(this.asbplayers).length) {
+                await new Promise(async (resolve) => {
```
killergerbah: Can you extract this promise into an async private function for readability?

extension/src/services/TabRegistry.js
```diff
@@ -104,7 +109,7 @@ export default class TabRegistry {
             }
         }
 
-        setTimeout(() => this._anyAsbplayerTab(videoTab, videoSrc, resolve, attempt + 1, maxAttempts), 1000);
+        setTimeout(() => this._anyAsbplayerTab(videoTab, videoSrc, resolve, reject, attempt + 1, maxAttempts), 1000);
```
killergerbah, line 116: Nice catch

extension/src/ui/components/VideoDataSyncDialaog.js
```diff
@@ -0,0 +1,97 @@
...
+    return (
+        <Dialog
+            disableEnforceFocus
+            fullWidth
+            maxWidth="sm"
+            open={open}
+            onBackdropClick={onCancel}
+            onEscapeKeyDown={onCancel}
+        >
+            <DialogContent>
```
killergerbah: Can you add a `DialogTitle` for consistency?
Maybe "Choose Subtitles"

extension/src/ui/components/VideoDataSyncDialaog.js
```diff
@@ -0,0 +1,97 @@
...
+                        value={name}
+                        onChange={(e) => setName(e.target.value)}
+                    />
+                    <div
+                        style={{
+                            display: showSubSelect ? 'block' : 'none',
+                            marginTop: '20px',
+                        }}
+                    >
+                        <InputLabel id="track-select-label">Subtitle Sync</InputLabel>
```
killergerbah: Can you use "Subtitle Tracks" for clarity?
Can you also make the select form take the full width of the dialog for consistency with the form element above it? This will take up the space where the spinner currently is. To deal with this you can move the spinner into the select menu itself. If that looks bad we can consider taking the spinner out entirely.

Also, I believe in Material UI design for select dialogs includes the input label as part of the dialog as in the samples found at https://mui.com/components/selects/ .

extension/src/ui/components/VideoDataSyncDialaog.js
```diff
@@ -0,0 +1,97 @@
...
+                <form>
+                    <TextField
+                        variant="filled"
+                        color="secondary"
+                        fullWidth
+                        label="Video Name"
+                        value={name}
+                        onChange={(e) => setName(e.target.value)}
+                    />
+                    <div
```
killergerbah: I think there's a more standard Material UI way to implement a form like this so that it doesn't need so many magic constants for arranging the elements around.
`SettingsDialog.js` almost does this but it has some special requirements so the form isn't completely standard. I think you can probably infer how to write a form with a select element in the standard MUI way from that example, though.

extension/src/ui/components/VideoDataSyncDialaog.js
```diff
@@ -0,0 +1,97 @@
...
+                                {subtitles.map((subtitle) => (
+                                    <MenuItem value={subtitle.url} key={subtitle.url}>
+                                        {subtitle.label}
+                                    </MenuItem>
+                                ))}
+                            </Select>
+                            {isLoading && <CircularProgress size={20} />}
+                        </div>
+                    </div>
+                    {error && <div style={{ color: '#ff1f62', marginTop: '20px' }}>Error occured: {error}</div>}
```
killergerbah: Material UI has a way to show errors related to form elements. I think the `error` property does this.

extension/src/ui/components/VideoDataSyncUi.js
```diff
@@ -0,0 +1,86 @@
...
+
+import { createTheme } from './theme';
+import VideoDataSyncDialog from './VideoDataSyncDialaog';
+
+export default function VideoDataSyncUi({ bridge }) {
+    const [open, setOpen] = useState(false);
+    const [isLoading, setIsLoading] = useState(true);
+    const [suggestedName, setSuggestedName] = useState('');
+    const [showSubSelect, setShowSubSelect] = useState(true);
+    const [subtitles, setSubtitles] = useState([{ language: '', url: '-', label: 'Off' }]);
```
killergerbah: Can you use 'None' for clarity?

extension/src/ui/components/VideoDataSyncDialaog.js
killergerbah: Can you use the secondary colors for the new form elements and spinner to match the colors of existing ones?

killergerbah (changes requested):
Thank you so much for the nice work! I did a first pass. Let me know if you need any clarity on anything.

I also found a bug: 
Open YouTube and then go to a video that has no subtitles. Immediately try to use `Ctrl+Shift+F`. Sometimes, the subtitle track selector has an infinite spinner. I'm guessing you have to use `Ctrl+Shift+F` very quickly.
asbplayer-pr-autosync-bug-20220113

README.md
Renji-XD: ya i think had "subtitles" first but after the merge now also the name is prefilled out (and maybe also new things in the future) so i thought maybe a generic "synced data" is potentially more accurate (why i also e. g. changed event name, option etc.). Let me know if i should change it or not (same for other instances) or use "available data"

extension/src/ui/components/VideoDataSyncDialaog.js
Renji-XD: ya i think for now except of the netflix api check there are only errors for sub sync so makes sense. If there are in furher updated more generic errors it maybe can be moved to an own element

extension/src/ui/components/VideoDataSyncDialaog.js
Renji-XD: ya i think for now the only errors are related to sub sync (except of the netflix player check) so makes sense. maybe in further updates and generic errors it may be an own element

Renji-XD:
i think your assumption is correct - i also had this on rare occasions when i was really speedy and pressed the keybind while elements / video were still building up / after navigation .

I think it was related to https://github.com/killergerbah/asbplayer/blob/main/extension/src/services/FrameBridgeClient.js#L72 (not sure if you looked at errors and can confirm your case). As the frameid is asynchronously set in a resolved promise and is awaited in the sync class i was a bit confused on what is the cause.. but as it was not happening often and only in a time window of couple of seconds after page load i did not spent time on it xD

Let me check again if i see other causes

README.md
killergerbah: I agree that strictly speaking "data" is probably more accurate, but in this case I think "subtitles" is really the main thing that users care about and would easily recognize. Using "data" everywhere would probably only make sense to users who have already used the feature already.

If we add more syncable data then we can reconsider what words to use again.

extension/src/handlers/video/SyncHandler.js
```diff
@@ -12,18 +12,22 @@ export default class SyncHandler {
...
-            chrome.tabs.sendMessage(Number(chosenTabId), {
-                sender: 'asbplayer-extension-to-player',
-                message: {
-                    command: 'syncv2',
-                    subtitles: request.message.subtitles,
-                },
-                src: request.src,
-                tabId: sender.tab.id,
-            });
+            if (chosenTabId) {
```
killergerbah: Another nitpick: This `if` is unnecessary, since at this point in the function, `findAsbplayerTab` has succeeded. I don't know if `chosenTabId` can be 0 or not, which is the only case that would cause bugs, but anyway the check can be removed.

extension/src/popup.html
```diff
@@ -1,75 +1,89 @@
...
+    </div>
+    <hr>
+    <div>
+        <input class="checkbox-color" id="subsDragAndDropInput" type="checkbox">
+        <label id="subsDragAndDrop" for="subsDragAndDropInput">Dropping subs into video syncs with asbplayer</label>
+    </div>
+    <hr>
+    <div>
+        <input class="checkbox-color" id="autoSyncInput" type="checkbox">
+        <label id="autoSync" for="autoSyncInput">Automatically sync available data</label>
```
killergerbah: As mentioned before, for now I think "available subtitles" will be clearer to the user.

extension/src/services/FrameBridgeClient.js
```diff
@@ -1,70 +1,75 @@
 export default class FrameBridgeClient {
     constructor(frame) {
         this.frame = frame;
+        this.bindPromise;
```
killergerbah: I'm not a JS expert - does this line do anything? Does it make sense to change the line to `this.bindPromise = undefined` if we only care about declaring the field?

extension/src/services/FrameBridgeClient.js
```diff
@@ -1,70 +1,75 @@
...
     }
 
     async bind() {
-        return new Promise((resolve, reject) => {
-            let ready = false;
+        this.bindPromise =
+            this.bindPromise ||
+            new Promise((resolve, reject) => {
```
killergerbah: Can you extract the `Promise` into a private function for readability?

extension/src/ui/components/VideoDataSyncDialaog.js
```diff
@@ -0,0 +1,119 @@
...
+import MenuItem from '@material-ui/core/MenuItem';
+import TextField from '@material-ui/core/TextField';
+import { makeStyles } from '@material-ui/styles';
+import React, { useEffect, useState } from 'react';
+
+const createClasses = makeStyles((theme) => ({
+    relative: {
+        position: 'relative',
+    },
+    spinner: {
```
killergerbah, line 17: Sorry for being so OCD, but on my computer the spinner seems a little bit off center. The spinner should be in line with the arrow icon, and the arrow icon should be the same distance away from the spinner as it is from the edge of the box.
asbplayer-pr-offcenter-20220115
:

extension/src/ui/components/VideoDataSyncDialaog.js
```diff
@@ -0,0 +1,119 @@
...
+    return (
+        <Dialog
+            disableEnforceFocus
+            fullWidth
+            maxWidth="sm"
+            open={open}
+            onBackdropClick={onCancel}
+            onEscapeKeyDown={onCancel}
+        >
+            <DialogTitle>Choose Data</DialogTitle>
```
killergerbah: As mentioned before, can we change this to "Select Subtitles?"

extension/src/video.js
```diff
@@ -2,16 +2,41 @@ import Binding from './services/Binding';
...
 
     const interval = setInterval(() => {
         const videoElements = document.getElementsByTagName('video');
 
         for (const v of videoElements) {
             const bindingExists = bindings.filter((b) => b.video.isSameNode(v)).length > 0;
 
-            if (!bindingExists) {
-                const b = new Binding(v);
+            if (!bindingExists && (v.src || [...v.children].some((elm) => 'SOURCE' === elm.tagName && elm.src))) {
```
killergerbah: A little bit of a micro-optimization, but I think it could be worth it since this loop runs every second.
Could you extract this test to a function? 

``` (v.src || [...v.children].some((elm) => 'SOURCE' === elm.tagName && elm.src)```

And instead of using `[...v.children].some`, just do a loop that checks each element for a source.

**Review by killergerbah — CHANGES_REQUESTED** [review]
Thank you for your patience and great work. I think this can be merged very soon.

File: extension/src/pages/netflix-page.js [hunk 28]
```diff
@@ -0,0 +1,147 @@
...
+                    return {
+                        label: `${track.bcp47} - ${track.displayName}${
+                            'CLOSEDCAPTIONS' === track.rawTrackType ? ' [CC]' : ''
+                        }`,
+                        language: track.bcp47.toLowerCase(),
+                        url: storedTracks.get(track.trackId),
+                    };
+                });
+            response.basename = `${title}${
+                season ? ` S${`${season}`.padStart(2, '0')}E${`${ep}`.padStart(2, '0')}` : ''
```
killergerbah: Nitpick: Maybe this won't ever happen but if season is zero and episode is not, then this string would end up being just the title.

Renji-XD:
Hi,

one more unfortunate issue i found with automatic syncing and fullscreen. When enabled and you use a (slower) connection like vpn it seems the time between video changes pass the time in which chrome clears the user activity state.. 

as the dialog is skipped the request to put the element back in fullscreen will throw with an error ("...API can only be initiated by a user gesture...").. you can catch it but there will always be a warning printed in the error section of the extension page...

so what is your preference here:
1. New Option to restore Fullscreen which needs to be manually enabled/disabled by user (most likely unclear to user what it is doing / when he should use it)
2. Simply catch and ignore the warning (fullscreen will not work on slow connections, can produe a lot of entries in the error section i guess)
3. Always skip to restore fullscreen on auto sync (trying to track any user interactivy seems odd - user need to retrigger fullscreen by himself in this case with autosync)
4. Simply skip the need of executing the keybind and always show the dialog (similiar to if the last synced language is not available).

killergerbah:
Hey @Renji-XD , why is there a need to exit fullscreen in the first place? Can we just let the sync happen without exiting full screen?

killergerbah (approved):
Thanks for your time and excellent work.

Issue #106, opened 2021-12-03
Regex filter for "sentence" text when exporting to Anki


killergerbah:
For example, it would be nice to be able to remove the name of the speaker that's commonly found in parentheses at the beginning of some subtitles. See https://github.com/killergerbah/asbplayer/issues/103

fr1edrik:
Not really sure if its according to this issue or the mentioned below, but I added a button for hiding the speaker when exporting (so its includes the regex).

https://github.com/killergerbah/asbplayer/pull/129

Issue #103, opened 2021-11-27
hide speaker from subtitles and/or Anki export


AsuraAce:
would be nice if you could add an option to hide the speaker from subtitles and the Anki export.

killergerbah:
What do you mean by speaker?

AsuraAce:
> What do you mean by speaker?

the name of the person that speaks at any given moment (the name in parenthesis in the image). 

![image](https://user-images.githubusercontent.com/13394741/143817102-1cd946e1-79c1-4845-b58e-8cf510397f7c.png)

AsuraAce:
Was able to solve this with Yomichan. Yomichan already provides this functionality.   

"You can use the regexReplace template helper in your Anki field templates."
https://github.com/FooSoft/yomichan/blob/master/docs/templates.md#regexreplace

killergerbah:
Closing this issue as I've created an issue specifically for the regex filter that would solve this problem.

PR #129, opened 2022-03-17
Regex filter for "sentence" text when exporting to Anki

fr1edrik:
https://github.com/killergerbah/asbplayer/issues/106

Added now a button where you can hide the speaker.

grafik

grafik

It removes the speaker (along with the spaces). The regex currently targets japanese and english parenphesis. I Couldnt find an example with japanese brackets (if someone provides an example I can add it)

If the user edits the text, the toggle will be disabled
grafik

It also checks if the pattern is reckognized. If it wasnt found it hides the toggle completely.
grafik

### DIFF
--- a/client/src/components/AnkiDialog.js
+++ b/client/src/components/AnkiDialog.js
@@ -19,6 +19,8 @@ import Tooltip from '@material-ui/core/Tooltip';
 import Typography from '@material-ui/core/Typography';
 import ZoomInIcon from '@material-ui/icons/ZoomIn';
 import TagsTextField from './TagsTextField';
+import { speakerRegex, getSubtitleWithoutSpeaker } from '../services/Util.js';
+import { Checkbox, FormControlLabel } from '@material-ui/core';
 
 const useStyles = makeStyles((theme) => ({
     root: {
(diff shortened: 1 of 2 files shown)


killergerbah:
Thanks! I'll be able to give this a closer look soon.

killergerbah (changes requested):
Thanks again for your time and work on this project!
Here are my thoughts:
- This change implements one of the feature requests (https://github.com/killergerbah/asbplayer/issues/103), but for a few reasons, some of which I mentioned on that issue, I believe this isn't the optimal way to solve the overarching problem.
  - The "speaker" text is highly specific to the subtitles being inspected.
  - Hiding this specific "speaker" text is important to relatively few users.
  - It's almost no effort anyway to delete the "speaker" text manually.
- For these reasons I would prefer that this problem is solved in a general way that provides more value to more users. For example: providing a way for the user to configure a regex filter themselves. A design for this might be:
  - Add a "Sentence Regex Filter" in the settings. The regex should be validated so that only correct regexes can be saved. Capture groups inside the regex should map to the parts of the text we want to remove.
  - For both the extension and app Anki dialog, try to automatically apply the "Sentence Regex Filter" to the sentence, if the sentence was prefilled by asbplayer.
  - If the filter was applied, provide an obvious button (probably in the sentence field) to remove the filter, as well as a way to quickly see what the regex was.

Feel free to let me what you think. Thanks!

killergerbah:
Use case is covered by subtitle regex filter now implemented

Issue #120, opened 2022-01-23
Way to download subtitles as srt 


killergerbah:
Now that asbplayer can produce subtitles on its own it would be good to have a way to get them onto disk.

PR #131, opened 2022-03-18
Download subtitles as srt

fr1edrik:
https://github.com/killergerbah/asbplayer/issues/120

Added feature to download currently loaded subtitles

When subtitles are loaded right click on the Subtitles Player.
grafik

After that click on "Download Subtitles"

grafik
File dialog opens with the current loaded title as name, but is changeable.

grafik

## Sidenote:

Couldnt test with the mentioned "asbplayer can produce subtitles on its own", because I dont know how to do it.

### DIFF
--- a/client/src/components/App.js
+++ b/client/src/components/App.js
@@ -909,6 +909,7 @@ function App() {
                                         )}
                                         <DragOverlay dragging={dragging} loading={loading} />
                                         <Player
+                                            title={fileName || 'asbplayer'}
                                             subtitleReader={subtitleReader}
                                             settingsProvider={settingsProvider}
                                             onCopy={handleCopy}
--- a/client/src/components/ContextMenu.js
+++ b/client/src/components/ContextMenu.js
@@ -1,0 +1,47 @@
+import { Menu, MenuItem } from '@material-ui/core';
+import React, { useState } from 'react';
+
+export default function ContextMenuWrapper({ menuItems, children }) {
+    const [anchorEl, setAnchorEl] = useState(null);
+    const [styles, setStyles] = useState(undefined);
+    const open = Boolean(anchorEl);
+
+    const screenH = window.innerHeight;
+
+    const handleClose = () => {
+        setAnchorEl(null);
+    };
+    const handleContexMenu = (e) => {
+        e.preventDefault();
+        setAnchorEl(null);
+
+        const anchor = e.currentTarget;
+        setStyles({ top: e.pageY - screenH / 2, left: e.pageX });
+        setAnchorEl(anchor);
+    };
+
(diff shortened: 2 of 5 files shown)


killergerbah (changes requested):
Thanks for taking the time to contribute to this project!

Here are a couple thoughts I had on this change:
- Implementing a new right-click menu on the subtitle player makes it impossible to use the right-click menu to do anything else, such as copy-and-paste. It would be great if we could move the download button somewhere else. Maybe next to the open-file button?
- I would suggest choosing the filename automatically. Using a prompt to ask for the filename is inconsistent with the styling and UX of the rest of the app. For example, the app already allows you to download image and audio things and in those cases the filename is always chosen automatically.
- You left some empty comments on some new functions. You should either write something there if it's necessary or not write anything at all to keep the code style somewhat consistent.

Feel free to let me know what you think. Thanks!

fr1edrik:
Hey, thanks for reviewing!

Regarding your points:

1. Yeah I see your point. I used this method because this was the most straighforward. We could theoretically place it next to "open file". But how its currently implemented outside the "Subtitle Player" scope no component has access to parsed "subtitles". With some store management like redux it would be easy, did you have something for the project already in mind? Also an option would be to have it behave like Youtube so you can rightclick twice(the first gets intercepted the second not)
2. I currently use the current title as suggestion. As I see it "download image" uses kind of the same(just from the file itself) along with an timestamp. Though yeah I may remove the file dialog and just take the title with the timestamp.
3. I only used the comments to have some typing as we dont use TS. Not sure what I should write there, as the function name seemed selfexplanatory to me.

killergerbah:
> Yeah I see your point. I used this method because this was the most straighforward. We could theoretically place it next to "open file". But how its currently implemented outside of the "Subtitle Player" scope no component has access to parsed "subtitles". With some store management like redux it would be easy, did you have something for the project already in mind? Also an option would be to have it behave like Youtube so you can rightclick twice(the first gets intercepted the second not)

I see, placing the button on the app bar would not be super easy. No plans for anything like redux right now - I think if we wanted subtitles to be accessible from the App component we could just lift the state up, but that would need to be done very carefully since the initialization logic for the Player is kind of a mess.

About the rightclick-twice approach - that seems pretty cool but I think I would still prefer to make right click behave normally since the normal right click menu provides the most value, as opposed to the custom one which relatively few people would need on some occasions.

Another approach might be just to put a download button in the controls that are overlaid on top of the SubtitlePlayer.

> I only used the comments to have some typing as we dont use TS. Not sure what I should write there, as the function name seemed selfexplanatory to me.

Yeah in that case you can just delete the empty comments.

killergerbah:
Closing this as this feature is now implemented in a different change.

PR #309, opened 2023-11-10
Added option to select multiple subtitle tracks at the same time from the track selector

Zyphdoz:
The subtitle track selector now has three dropdown menus instead of one, allowing users to load up to three tracks at once.


extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -357,6 +361,33 @@ export default class VideoDataSyncController {
...
+            
+            for (let i = 0; i < data.length; i++) {
+                const { name, extension, subtitleUrl, m3U8BaseUrl } = data[i];
+                const subtitleFiles = await this._subtitlesForUrl(name, extension, subtitleUrl, m3U8BaseUrl);
+                if (subtitleFiles !== undefined) {
+                    subtitles.push(...subtitleFiles);
+                }
+            }
+
+            if (subtitles === undefined) {
```
killergerbah: This check can be removed since `subtitles` is always defined

extension/src/ui/components/VideoDataSyncDialog.tsx
```diff
@@ -113,7 +117,99 @@ export default function VideoDataSyncDialog({
...
+    }, [suggestedName, selectedSubtitles[0], subtitles]);
+
+    function handleOkButtonClick() {
+        const selectedSubtitleTracks: ConfirmedVideoDataSubtitleTrack[] = allSelectedSubtitleTracks();
+
+        // Remove all but one empty track in order to intentionally send one empty track if no tracks are selected
+        const uniqueTracks: ConfirmedVideoDataSubtitleTrack[] = filterByUniqueUrl(selectedSubtitleTracks);
+
+        // If track length > 1 and we have unique tracks, then at least one language must have been selected and it is safe to remove the remaining empty track
+        uniqueTracks.length > 1 ? onConfirm(removeEmptyTracks(uniqueTracks)) : onConfirm(uniqueTracks);
```
killergerbah: Can you rewrite this using `if` ... `else` ... instead of a ternary operator? Ternary operator are used to return values and here no value is being returned.

killergerbah (commented):
Great work.
Can you run `prettier` on the changed files to make the formatting consistent?

killergerbah (approved):
Thank you!

PR #313, opened 2023-11-26
Toggle for remembering subtitle tracks and remember all tracks instead of only the first.

Zyphdoz:
- Subtitle track selector remembers all three tracks instead of just one.
- Added toggle for remembering subtitle tracks.
By default, the toggle is checked when all preferred tracks were auto-selected. It defaults to unchecked if one or more preferred languages were not available. This prevents preferences from being overwritten with empty in instances where the preferred languages were not available whilst still giving the user the option to overwrite them if they want to.


common/src/message.ts
```diff
@@ -423,6 +423,7 @@ export interface AnkiUiBridgeRerecordMessage extends Message {
 export interface VideoDataUiBridgeConfirmMessage extends Message {
     readonly command: 'confirm';
     readonly data: ConfirmedVideoDataSubtitleTrack[];
+    readonly shouldRememberTrackChoices: Boolean;
```
killergerbah: Use the primitive `boolean` instead

extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -158,30 +165,16 @@ export default class VideoDataSyncController {
...
-                      url: '-',
-                      label: i18n.t('extension.videoDataSync.emptySubtitleTrack'),
-                      extension: 'srt',
-                  }
-                : subtitleTrackChoices.find((subtitle) => subtitle.language === this.lastLanguageSynced);
+        const subs = this._matchLastSyncedWithAvailableTracks();
+        const selectedSub: VideoDataSubtitleTrack[] = subs.autoSelectedTracks;
+        const defaultCheckboxState: boolean = subs.completeMatch;
 
         if (selectedSub !== undefined && !userRequested && !this._syncedData?.error) {
```
killergerbah: With the current logic we will never get into the `else` block. The `else` block is used to show the track selector in case the user's preferred tracks aren't found. Right now, we are auto-loading tracks even when there are no matching tracks found.

killergerbah (changes requested):

killergerbah (approved):

killergerbah:
Thanks!!

PR #317, opened 2023-12-02
Fix bug with empty subtitle names showing up on the screen when user has 'Open subtitle list when loading subtitle' enabled in the settings.

Zyphdoz:
Get rid of isEmptySubtitle and detect empty tracks in binding.ts instead. That way it's handled regardless of which path it takes to get there.


killergerbah:
Thanks you so much for your contributions!

PR #323, opened 2023-12-05
Fast forward between subtitles

artjomsR:
I'm trying to figure out a feature and was wondering if you could help me out. I'm trying to do a variant of condensed play mode where it speeds up between subtitles rather than completely skipping them, as it does currently. I've been using the auto pause play mode for my testing: as you can see, I simply change the speed when the subtitle starts/ends. 

This works, but only sometimes: either the player keeps playing at 2.3x when the next subtitle is displayed or doesn't apply the 2.3x speed when the subtitle ends. I was wondering if you had thoughts on why it might and if it's possible to fix it


artjomsR:
Ok, thank you very much for your time! That definitely makes sense. I've tried lower values, but that didn't work at all. I figured we'd need a different approach with constant polling anyway. This is because there's always a risk of being in the middle of the subtitle to trigger. 

For now, I've put this in `src\controllers\subtitle-controller.ts:this.subtitlesInterval = setInterval`:
```
if (slice.showing.length === 0) {
    this.video.playbackRate = 2.3
} else {
    this.video.playbackRate = 1
}
```

I've done a quick a test, and it seems to work

killergerbah:
That makes sense. If you want to merge this in I would suggest creating a new boolean setting that controls the behavior of condensed mode, as I believe a lot of users would prefer completely skipping through gaps instead of fast forwarding through them.

extension/src/services/binding.ts
```diff
@@ -331,7 +368,7 @@ export default class Binding {
 
             chrome.runtime.sendMessage(command);
 
-            if (this._synced) {
+            if (this._synced && this._playMode !== PlayMode.fastForward) {
```
artjomsR, line 386: I think the constant speed notification is annoying so I've disabled it for the new mode

artjomsR:
>  creating a new boolean setting that controls the behavior of condensed mode
I thought about doing it originally but ended up implementing as a new shortcut. I made sure that both condensed and FF mode can't be on at the same time. The reason is simple - more options to the user. I like condensed mode as well, but sometimes when it switches between scenes, I lose context. So I think it's useful to have both since they're useful in slightly different contexts

common/locales/en.json
```diff
@@ -91,6 +91,7 @@
         "toggleAsbplayerSubtitleTracks": "Toggle subtitle track in asbplayer",
         "toggleAutoPause": "Toggle auto-pause",
         "toggleCondensedPlayback": "Toggle condensed playback",
+        "toggleFastForwardPlayback": "Toggle fast forward playback",
```
killergerbah, line 94: If you run `node merge-loc-from-en.js` inside `scripts` then you can copy this line to the loc files for all of the other languages
With the way asbplayer works right now, all of the keys need to exist in all of the loc files.

extension/src/services/binding.ts
```diff
@@ -183,14 +197,36 @@ export default class Binding {
...
+                    disableCondensedMode();
+                }
+                this.subtitleController.fastForwardBetweenSubtitles = async (slice) => {
+                    const subtitleIsSufficientlyOffsetFromNow = (subtitleEdgeTime: number | undefined) => {
+                        return subtitleEdgeTime &&
+                            Math.abs(subtitleEdgeTime - this.video.currentTime * 1000) > this.fastForwardPlaybackMinimumGapMs
+                    }
+                    if (
+                        slice.showing.length === 0 &&
+                        subtitleIsSufficientlyOffsetFromNow(slice?.lastShown?.[0]?.end) &&
```
killergerbah: You should use the largest `end` value

extension/src/services/binding.ts
```diff
@@ -183,14 +197,36 @@ export default class Binding {
...
+                        subtitleIsSufficientlyOffsetFromNow(slice?.nextToShow?.[0]?.start)
```
killergerbah: Similarly you should use the smallest `start` value

extension/src/services/binding.ts
```diff
@@ -183,14 +197,36 @@ export default class Binding {
                 };
                 this.subtitleController.notification('info.enabledCondensedPlayback');
                 break;
+            case PlayMode.fastForward:
+                if (this._playMode === PlayMode.condensed) {
+                    disableCondensedMode();
+                }
+                this.subtitleController.fastForwardBetweenSubtitles = async (slice) => {
```
killergerbah: It would be more consistent with the naming of the other hooks to call this something like `onSlice` which isn't specific to fast forwarding.

killergerbah (changes requested):
Can you also make sure that `yarn run verify` from the project root passes?

killergerbah (changes requested):
Looks good, if you just could run prettier on the files you changed I'll merge this in. That's the only check that doesn't pass in `yarn run verify`.

killergerbah:
Thank you!!

PR #328, opened 2023-12-09
Show hover effect on checkboxes and radio buttons when the label is hovered.

Zyphdoz:


common/components/SettingsForm.tsx
```diff
@@ -26,6 +26,7 @@ import { AsbplayerSettings, KeyBindName, SubtitleListPreference } from '@project
 import { computeStyles, download } from '@project/common/util';
 import { CustomStyle, supportedLanguages, validateSettings } from '@project/common/settings';
 import { useOutsideClickListener } from '@project/common/hooks';
+import { styled } from '@material-ui/core';
```
killergerbah: Can you import it like this instead?
```import styled from '@material-ui/core/styles/styled';```
This will reduce the amount of code imported as the entire `@material-ui/core` module won't be imported.

File: common/components/SettingsForm.tsx [hunk 2]
```diff
@@ -120,6 +121,15 @@ const useSelectableSettingStyles = makeStyles<Theme>((theme) => ({
     },
 }));
 
+export const LabelWithHoverEffect = styled(FormControlLabel)(({ theme }) => ({
```
killergerbah: Can you put this in its own file and export this function as the default export? Then import it as needed here and in `VideoDataSyncDialog.tsx`. This will allow just that component to be imported without importing the code in this file as well.

extension/src/ui/components/VideoDataSyncDialog.tsx
```diff
@@ -14,6 +13,7 @@ import Toolbar from '@material-ui/core/Toolbar';
 import Typography from '@material-ui/core/Typography';
 import makeStyles from '@material-ui/styles/makeStyles';
 import Switch from '@material-ui/core/Switch';
+import { LabelWithHoverEffect } from '@project/common/components/SettingsForm';
```
killergerbah: As mentioned above, import `LabelWithHoverEffect` from its own file rather than pulling in `SettingsForm` here.

killergerbah (changes requested):

killergerbah (approved):

killergerbah:
Thanks!!

PR #334, opened 2023-12-15
fix width of rememberTrackPreference label in track selector

Zyphdoz:
The clickable area was too wide

image


killergerbah (approved):

killergerbah:
Sweet, thanks as always

PR #345, opened 2024-01-04
Repeat mode

iam6lake:
Inconsistently working at the moment, seems to clash with other modes. Opening this draft pull request for visibility and feedback from those more familiar with the code.

### DIFF
--- a/common/app/components/Player.tsx
+++ b/common/app/components/Player.tsx
@@ -226,6 +226,20 @@ export default function Player({
     const classes = useStyles({ appBarHidden, appBarHeight });
     const calculateLength = () => trackLength(channelRef.current, subtitlesRef.current);
 
+    const seek = useCallback(
+        async (time: number, clock: Clock, forwardToMedia: boolean) => {
+            clock.setTime(time);
+            forceUpdate();
+
+            if (forwardToMedia) {
+                await mediaAdapter.seek(time / 1000);
+            }
+
+            autoPauseContextRef.current?.clear();
+        },
+        [forceUpdate, mediaAdapter]
+    );
+
     const handleSubtitlePlayerResizeStart = useCallback(() => setSubtitlePlayerResizing(true), []);
     const handleSubtitlePlayerResizeEnd = useCallback(() => setSubtitlePlayerResizing(false), []);
 
@@ -241,17 +255,27 @@ export default function Player({
         pause(clock, mediaAdapter, true);
     }, [playMode, clock, mediaAdapter, videoFileUrl, settings]);
 
-    const handleOnWillStopShowingSubtitle = useCallback(() => {
-        if (
-            playMode !== PlayMode.autoPause ||
-            settings.autoPausePreference !== AutoPausePreference.atEnd ||
-            videoFileUrl // Let VideoPlayer do the auto-pausing
-        ) {
-            return;
-        }
+    const getCurrentSubtitle = (currentTime: number) => {
+        const subtitlesAtCurrentTime = subtitleCollection.subtitlesAt(currentTime);
+        return subtitlesAtCurrentTime.showing.length > 0 ? subtitlesAtCurrentTime.showing[0] : null;
+    };
 
-        pause(clock, mediaAdapter, true);
-    }, [playMode, clock, mediaAdapter, videoFileUrl, settings]);
+    const handleOnWillStopShowingSubtitle = useCallback(() => {
+        const currentTime = clock.time(calculateLength());
+        const currentSubtitle = getCurrentSubtitle(currentTime);
+
+        if (playMode === PlayMode.repeat && currentSubtitle) {
+            // If in repeat mode, seek to the start of the current subtitle
+            seek(currentSubtitle.start, clock, true);
+        } else if (
+            playMode === PlayMode.autoPause &&
+            settings.autoPausePreference === AutoPausePreference.atEnd &&
+            !videoFileUrl // Ensure not to interfere with VideoPlayer's auto-pausing
+        ) {
+            // Handle auto-pause logic
+            pause(clock, mediaAdapter, true);
+        }
+    }, [playMode, clock, mediaAdapter, videoFileUrl, settings, seek, getCurrentSubtitle, calculateLength]);
 
     const autoPauseContext = useMemo(() => {
         const context = new AutoPauseContext();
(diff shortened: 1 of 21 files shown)


killergerbah:
Hi @iam6lake , thanks for contributing.
At first glance you might be running into issues with `Player.tsx` and `VideoPlayer.tsx` being in different documents. The `VideoPlayer` is in an iframe and might not always receive key events, so I would implement the keyboard bind inside of `Player.tsx` as well.
There is also a separate question of whether the actual loop behavior needs to be in `VideoPlayer.tsx` at all. `Player.tsx` implements the logical behavior of many of the other play modes, allowing `VideoPlayer.tsx` to do its own auto-pausing only because there's a need to have the video pause precisely while the subtitle is still on screen.

extension/src/services/binding.ts
```diff
@@ -235,7 +237,26 @@ export default class Binding {
                 };
                 this.subtitleController.notification('info.enabledFastForwardPlayback');
                 break;
+            case PlayMode.repeat:
+                if (!this.repeatIntervalId) {
```
killergerbah: Instead of scheduling and saving a new interval I would suggest just subscribing to the existing onWillStopShowing event instead. This would reduce the amount of additional state you need to manage, as currently you are having to schedule and unschedule the interval.

common/app/components/VideoPlayer.tsx
```diff
@@ -1025,6 +1025,41 @@ export default function VideoPlayer({
...
+                }
+            },
+            () => false
+        );
+
+        return () => unbindRepeat();
+    }, [keyBinder, playMode, setPlayMode]);
+
+    useEffect(() => {
+        if (playMode !== PlayMode.repeat || showSubtitlesRef.current.length === 0) {
```
killergerbah: `repeat` mode currently will only work when a video is open. To maintain consistency with the other play modes I would implement this logic inside `Player.tsx` instead.

killergerbah (changes requested):

iam6lake:
OK, made those chagnes

common/app/components/Player.tsx
```diff
@@ -683,6 +683,31 @@ export default function Player({
         return () => clearInterval(interval);
     }, [updatePlaybackRate, subtitleCollection, clock, subtitles, playMode, settings.fastForwardModePlaybackRate]);
 
+    useEffect(() => {
+        if (playMode === PlayMode.repeat) {
+            autoPauseContext.onWillStopShowing = () => {
```
killergerbah: I don't think this is causing any actual bugs right now but this particular `onWillStopShowing` assignment is overriding the assignment made on line 259. Can you move this logic into the existing subscription `handleOnWillStopShowingSubtitle`?

common/app/components/VideoPlayer.tsx
```diff
@@ -1025,6 +1025,22 @@ export default function VideoPlayer({
         );
     }, [keyBinder, playerChannel]);
 
+    useEffect(() => {
+        const unbindRepeat = keyBinder.bindToggleRepeat(
+            (event) => {
+                event.preventDefault();
+                if (playMode === PlayMode.repeat) {
```
killergerbah: You'll want to use `togglePlayMode` instead, otherwise the keyboard shortcut won't appear to work when the `VideoPlayer`'s `iframe` is in focus.

killergerbah (changes requested):
One more general comment: I think it would be cleaner if the 'toggle repeat' shortcut only does something when there's a subtitle on screen. It feels counter-intuitive that the mode gets enabled and nothing happens in the case when there is no subtitle on screen.

I know we've gone through quite a bit of back-and-forth, so thanks for your patience working with me.

common/app/components/Player.tsx
```diff
@@ -241,17 +255,27 @@ export default function Player({
...
+        return subtitlesAtCurrentTime.showing.length > 0 ? subtitlesAtCurrentTime.showing[0] : null;
+    };
+
     const handleOnWillStopShowingSubtitle = useCallback(() => {
-        if (
-            playMode !== PlayMode.autoPause ||
-            settings.autoPausePreference !== AutoPausePreference.atEnd ||
-            videoFileUrl // Let VideoPlayer do the auto-pausing
+        const currentTime = clock.time(calculateLength());
+        const currentSubtitle = getCurrentSubtitle(currentTime);
```
killergerbah: The `currentSubtitle` calculation here is not necessary (sorry it sort of triggers my OCD that this calculation is being done again). The subtitle can be passed through from `AutoPauseContext` if you modify the signature of the `willStopShowingSubtitle` hook.

common/settings/settings-provider.ts
```diff
@@ -65,6 +65,7 @@ export const defaultSettings: AsbplayerSettings = {
         decreasePlaybackRate: { keys: isMacOs ? '⇧+⌃+[' : 'ctrl+shift+[' },
         increasePlaybackRate: { keys: isMacOs ? '⇧+⌃+]' : 'ctrl+shift+]' },
         toggleSidePanel: { keys: '`' },
+        toggleRepeat: { keys: 'R' },
```
killergerbah: Can this shortcut be `Shift + R` instead so that it's harder to trigger by accident?

extension/src/services/binding.ts
```diff
@@ -237,7 +239,20 @@ export default class Binding {
...
+                    };
+                    this.subtitleController.notification('info.enabledRepeatPlayback');
+                }
+                break;
             case PlayMode.normal:
+                if (this._playMode === PlayMode.repeat) {
+                    this.subtitleController.autoPauseContext.onWillStopShowing = undefined;
+                    this.subtitleController.notification('info.disabledRepeatPlayback');
+                }
                 if (this._playMode === PlayMode.autoPause) {
```
killergerbah: Can use `else if` here

killergerbah (changes requested):

common/app/components/Player.tsx
iam6lake: Changed

common/settings/settings-provider.ts
iam6lake: Changed

extension/src/services/binding.ts
iam6lake: Changed

iam6lake:
Addressed all current comments

killergerbah:
Thanks!

Issue #349, opened 2024-01-08
Automatically play media when invoking 'Seek to beginning of current'


artjomsR:
I've mentioned this one in discord. Implementation-wise, it's as simple as adding `context.play();` to `extension\src\services\key-bindings.ts:_unbindSeekToBeginningOfCurrentSubtitle`.

You mentioned that you'd prefer this to be a setting, and this is where I got stuck - that file doesn't seem to be using any settings at the moment. And I'm not sure what's the best way to import them.

killergerbah:
The approach I would take is to add the new setting, and then cache it as an instance variable in `Binding._refreshSettings`. Then in `_unbindSeekToBeginningOfCurrentSubtitle` you can access the cached variable through `context` to decide whether to `play()` or not.

artjomsR:
Closed as completed, thanks for the help with implementation tips :)

PR #350, opened 2024-01-09
Autoplay after sub repeat option

artjomsR:
Implements https://github.com/killergerbah/asbplayer/issues/349


killergerbah:
Thanks!

PR #352, opened 2024-01-11
Seek to beginning of current subtitle: Play previous subtitle if current is not found

artjomsR:
I'm not sure what to do with the existing translations though


artjomsR:
Ok, I've updated them 👍

common/settings/settings.ts
```diff
@@ -113,7 +113,7 @@ export interface KeyBindSet {
     readonly seekForward: KeyBind;
     readonly seekToPreviousSubtitle: KeyBind;
     readonly seekToNextSubtitle: KeyBind;
-    readonly seekToBeginningOfCurrentSubtitle: KeyBind;
+    readonly seekToBeginningOfCurrentOrPreviousSubtitle: KeyBind;
```
killergerbah: Although it will be inconsistent with the behavior of the bind, could you keep the previous name of the KeyBind? Changing the key without some kind of migration will cause users to lose their setting.

artjomsR:
Ok, I've reverted most of the renames because I think the discrepancy between the names would only add to confusion

killergerbah:
Awesome, thanks as always!

PR #365, opened 2024-01-30
Add Emby Support

bpwhelan:
I'm expecting a bit of pushback, but I'm mostly happy with this as of right now.

- Couldn't figure out a config, made the "isVideoPlayerPage()" logic quite a bit more complicated, the host regex is overly simple as is...
- path doesn't work since emby uses the same path for everything
- Might be able to just go through ApiClient Directly? But i already had the code written to hit the RESTApi directly.

examples of urls where we notice the video player and attempt to get subtitles.

http://ip:8096/web/index.html#!/videoosd/videoosd.html

https://emby.custom.url/web/index.html#!/videoosd/videoosd.html


extension/src/pages/emby-page.ts
```diff
@@ -0,0 +1,40 @@
+import { VideoDataSubtitleTrack } from '@project/common';
+import { VideoData } from '@project/common';
+
+declare const ApiClient: any | undefined;
+
+document.addEventListener(
+    'asbplayer-get-synced-data',
+    async () => {
+        const deviceID = ApiClient?._deviceId;
+        const apikey = ApiClient?._userAuthInfo.AccessToken;
```
killergerbah: If for some reason `ApiClient` doesn't exist then the script should send back an error as in some of the other page scripts. As it is right now it is possible to `fetch` with `apiKey` being `undefined`.

extension/src/services/pages.ts
```diff
@@ -74,6 +75,9 @@ export class PageDelegate {
     }
 
     isVideoPage() {
+        if (this.config.hash) {
```
killergerbah, line 80: Now that both `hash` and `pathname` are possible I think it would make sense to treat them both as optional and test each setting if it is present. Right now `pathname` is ignored if `hash` is present.

extension/src/pages.json
```diff
@@ -79,6 +79,15 @@
             "autoSync": {
                 "enabled": true
             }
+        },
+        {
+            "host": "emby*|:8096*",
```
killergerbah, line 84: Is there a specific reason why `emby` would be in the host, or is that just a likely user preference?

killergerbah (changes requested):
Thanks for your work on this @bpwhelan . Regarding the problem with the `host` configuration, how likely do you think it is for an Emby user to use a non-default port? If it is not likely then I am thinking it might be reasonable to requires users to use the default port if they want to use asbplayer with Emby.

extension/src/pages.json
bpwhelan: User preference. If they run emby under a custom domain I think it's likely that they will have "emby" somewhere in dns. I personally do. 

Also, I think it's safe to assume most emby users are using the default 8096 for emby. The only case where they wouldn't be is if they are also running jellyfin (what emby was forked from) on the same server.

extension/src/services/pages.ts
bpwhelan: Makes sense, updated to include both in [df1e15f](https://github.com/killergerbah/asbplayer/pull/365/commits/df1e15ffd19897c24ab4a0c383bd707726b83fbc)

extension/src/pages/emby-page.ts
bpwhelan: Updated to check `!ApiClient` and send back error in [df1e15f](https://github.com/killergerbah/asbplayer/pull/365/commits/df1e15ffd19897c24ab4a0c383bd707726b83fbc)

killergerbah (approved):
Nice work, thanks!

killergerbah:
Ah crap, can you run prettier on the changed files? Will merge it in after that.

bpwhelan:
> Ah crap, can you run prettier on the changed files? Will merge it in after that.

Done.

killergerbah:
Thank you!

PR #383, opened 2024-03-10
feature: option to change name of tab

m-edlund:
This adds the option to change the name of the player tab. 
The use case for this is being able to use the document-title field in yomitan with asbplayer. Say for example you're watching several episodes of a show, simply set the title of the show as the tab name and all anki cards can then be tagged with the name of the show. 

All translations were done with DeepL, and the German, English, Spanish and Japanese translations were double checked by a human.

### DIFF
--- a/common/app/components/App.tsx
+++ b/common/app/components/App.tsx
@@ -1127,6 +1127,10 @@ function App({ origin, logoUrl, settings, extension, fetcher, onSettingsChanged 
         );
     }, [keyBinder, handleOpenCopyHistory, ankiDialogOpen]);
 
+    useEffect(() => {
+        document.title = settings.tabName;
+    }, [settings]);
+
     const { initialized: i18nInitialized } = useI18n({ language: settings.language });
 
     if (!i18nInitialized) {
--- a/common/app/components/App.tsx
+++ b/common/app/components/App.tsx
@@ -1129,7 +1129,7 @@ function App({ origin, logoUrl, settings, extension, fetcher, onSettingsChanged 
 
     useEffect(() => {
         document.title = settings.tabName;
-    }, [settings]);
+    }, [settings.tabName]);
 
     const { initialized: i18nInitialized } = useI18n({ language: settings.language });
 
--- a/common/app/services/video-channel.ts
+++ b/common/app/services/video-channel.ts
@@ -531,6 +531,7 @@ export default class VideoChannel {
             clickToMineDefaultAction,
             language,
             lastSubtitleOffset,
+            tabName,
         } = settings;
         const message: MiscSettingsToVideoMessage = {
             command: 'miscSettings',
@@ -550,6 +551,7 @@ export default class VideoChannel {
                 clickToMineDefaultAction,
                 language,
                 lastSubtitleOffset,
+                tabName,
             },
         };
         this.protocol.postMessage(message);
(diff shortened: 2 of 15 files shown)


killergerbah:
Thanks for contributing. I have a couple of questions before merging:
- Did you explore using asbplayer's tag setting instead?
- How are you using this diff to actually get the behavior you want? I'm not seeing a way to configure the new setting from the UI.

m-edlund:
Sorry, I worded my explanation badly :sweat_smile: 

The aim of this feature has nothing to do with tagging. What I'm aiming to achieve is to give the option to essentially freely write the field of a card. So by setting the name of the tab you can then use the `document-title` handlebar to set a field as you wish.
While on the other hand with tags you can only add to the already exiting tags.

My personal use case for this is having the source of the content I mined directly baked into the card. So if I mined something from a wikipedia article it will automatically add the name of the article (name of the tab). This would just be a continuation of that, but with the difference you can choose what you add, since asbplayer has no way of automatically knowing what is being watched.

One though I had, was that this could maybe be solved with the "Add Custom Field" option (creating a custom field for the anki field that would usually be your `document-title`),  but I did not see a way to actually set the content that is written into the custom field. If there is a way to control the content that is written there, this feature would be superflous.

killergerbah:
Got it, I missed the part in your initial explanation about using `document-title` in the card template itself. There are some future plans to implement fancier custom Anki fields but those changes are likely not coming in the short-term.

common/app/components/App.tsx
```diff
@@ -1127,6 +1127,10 @@ function App({ origin, logoUrl, settings, extension, fetcher, onSettingsChanged
         );
     }, [keyBinder, handleOpenCopyHistory, ankiDialogOpen]);
 
+    useEffect(() => {
+        document.title = settings.tabName;
+    }, [settings]);
```
killergerbah: Use `settings.tabName` instead of `settings` in the dependency array, so that the effect will run less often.

killergerbah (commented):
One last question:
Why is the `Source` from asbplayer not sufficient for your purposes? In general, the `Source` field should be populated with the thing you're consuming in asbplayer. While the field is not directly accessible from Yomitan, if you're updating cards with asbplayer-supplied information then you could simply assign the `Source` field to a field in your Anki note type.

common/app/components/App.tsx
m-edlund: Done.

m-edlund:
> One last question: Why is the `Source` from asbplayer not sufficient for your purposes? In general, the `Source` field should be populated with the thing you're consuming in asbplayer. While the field is not directly accessible from Yomitan, if you're updating cards with asbplayer-supplied information then you could simply assign the `Source` field to a field in your Anki note type.

The issue is that it takes the source from the file name. So if say you're watching a series and want the value of the source field to be the same for all the cards mined from any episode, you would have to edit the field on every export. 

One solution I could see that uses the source field instead of the name of the tab, is adding manual overrides to fields that will be used if the user has defined them. 

So while you're watching a series called Foo Bar, you can set the override to "FooBar" and it won't take the name of the current subtitle file "fooBar-episode21.srt", but "FooBar" instead. And once you're done with the series you can remove the manual override again. 

If you want, I can implement that instead.

killergerbah:
Hey can you make `yarn run verify` pass? Then I'll merge. Thanks!

m-edlund:
I've never used jest before, and I'm getting an error while trying to run the client tests that I can't seem to fix. However I updated the settings test and test schema with the new setting, which should hopefully resolve the failed tests.

killergerbah:
The final prettier check is failing, so you just need to run prettier on `common/settings/settings-import-export.test.ts`

PR #413, opened 2024-04-30
Add setting to blur and unblur subtitle tracks on mouse hover

nekorushi:
Added setting that allows for blurring of selected subtitle tracks. When mouse hovered, they unblur.


killergerbah:
@nekorushi Thanks for taking the time to contribute your idea and implementation. The only major feedback I have is that I would prefer if the data model for the settings future-proofed itself against the possibility of more track-specific settings. This would make it more trivial to implement more of these types of settings for more than 3 tracks, with less code duplication, and with less hardcoding of keys that look like `foo1, foo2, foo3, ...`.  For example, I have received a feature request to make subtitle alignment track-specific as well, and the implementation for this would be simplified if the data model already readily supported track-specific settings. This is my suggestion for the settings data model:

```
subtitleTracks: [{"blur": true/false}]
```

common/locales/en.json
```diff
@@ -295,6 +295,11 @@
         "subtitleAlignment": "Subtitle Alignment",
         "subtitleAlignmentBottom": "Bottom",
         "subtitleAlignmentTop": "Top",
+        "subtitleBlur": "Subtitle blur",
+        "subtitleBlurTrack1": "Track 1",
+        "subtitleBlurTrack2": "Track 2",
+        "subtitleBlurTrack3": "Track 3",
+        "subtitleBlurDescription": "Hides selected subtitle tracks by blurring them, can be unblurred on mouse hover",
```
killergerbah: Nitpick: `Hides selected subtitle tracks by blurring them. Can be un-blurred on mouse hover.`

killergerbah:
Thanks @nekorushi for your contribution. It is close to merge, can you just make `yarn run verify` pass? You can update the loc keys for non-English languages by running `node merge-loc-from-en.js` inside of `scripts`.

nekorushi:
Hey @killergerbah sorry for the delay. I've updated the locale and some missing config found with `yarn run verify`

killergerbah:
Thank you @nekorushi !

pawel-soltys:
@Kellen-wq It doesn't animate. Originally I added such CSS on my side. But for the regular use it felt very annoying due to delaying the reveal. Degree is also not adjustable. I've set the blur to 10px so the text is surely not readable.

The settings surely could be added but for the usability of the feature I am not sure if it's really needed. If you insist though, the earliest I could take a look would be the end of the next month due to personal plans.

nekorushi:
@Kellen-wq It doesn't animate. Originally I added such CSS on my side. But for the regular use it felt very annoying due to delaying the reveal. Degree is also not adjustable. I've set the blur to 10px so the text is surely not readable.

The settings surely could be added but for the usability of the feature I am not sure if it's really needed. If you insist though, the earliest I could take a look would be the end of the next month due to personal plans.

PR #415, opened 2024-05-05
PostMinePlayback preference

artjomsR:
As discussed, 3 options to choose from for the new setting


common/settings/settings.ts
```diff
@@ -39,6 +39,7 @@ export interface AnkiSettings {
     readonly maxImageHeight: number;
     readonly surroundingSubtitlesCountRadius: number;
     readonly surroundingSubtitlesTimeRadius: number;
+    readonly clickToMineDefaultPlayback: PostMinePlayback;
```
killergerbah: Looking at this now the settings categorization does seem somewhat arbitrary but I think I would still prefer for this setting to be under `MiscSettings`.

killergerbah: I would name this setting something else since it applies even when not clicking to mine.

killergerbah (changes requested):

killergerbah (approved):

killergerbah:
Thanks as always @artjomsR

artjomsR:
Always a pleasure, thanks for your hard work on this beautiful project!

PR #420, opened 2024-05-10
Add speed change step

artjomsR:
`0.1` was too slow so figured I'd make it customisable :) 

***
I've also added a small improvement in FF mode: as of master it would block increase/decrease speed shortcuts, so I've made it work: i.e. if there's a subtitle active, and you increase the speed (assuming defaults), it'll go up to 1.0 + 0.1 = 1.1 or if there's no subtitle: 2.7 + 0.1 = 2.8

***
I've tried to make the above changes work for the video player, here's my best attempt in `common\app\components\VideoPlayer.tsx`:
```        
return keyBinder.bindAdjustPlaybackRate(
            (event, increase) => {
                event.preventDefault();
                const video = videoRef.current;

                if (!video) {
                    return;
                }

                const currentSpeed = video.playbackRate;
                togglePlayMode(event, PlayMode.normal);

                if (increase) {
                    updatePlaybackRate(Math.min(5, currentSpeed + settings.speedChangeStep), true);
                } else {
                    updatePlaybackRate(Math.max(0.1, currentSpeed - settings.speedChangeStep), true);
                }
            },
            () => false
        );
```
I've had several problems with this:
1. the speed step remains as 0.1 until the value is actively changed in the settings, then the new value is being picked up by the shortcut
2. ~when FF mode is on, it still blocks shortcuts, despite me putting the call to switch to Normal playback mode~ probably too niche anyway, let's not worry about this for now


killergerbah:
> `0.1` was too slow so figured I'd make it customisable :)
> 
> I've also added a small improvement in FF mode: as of master it would block increase/decrease speed shortcuts, so I've made it work: i.e. if there's a subtitle active, and you increase the speed (assuming defaults), it'll go up to 1.0 + 0.1 = 1.1 or if there's no subtitle: 2.7 + 0.1 = 2.8
> 
> I've tried to make the above changes work for the video player, here's my best attempt in `common\app\components\VideoPlayer.tsx`:
> 
> ```
> return keyBinder.bindAdjustPlaybackRate(
>             (event, increase) => {
>                 event.preventDefault();
>                 const video = videoRef.current;
> 
>                 if (!video) {
>                     return;
>                 }
> 
>                 const currentSpeed = video.playbackRate;
>                 togglePlayMode(event, PlayMode.normal);
> 
>                 if (increase) {
>                     updatePlaybackRate(Math.min(5, currentSpeed + settings.speedChangeStep), true);
>                 } else {
>                     updatePlaybackRate(Math.max(0.1, currentSpeed - settings.speedChangeStep), true);
>                 }
>             },
>             () => false
>         );
> ```
> 
> I've had several problems with this:
> 
> 1. the speed step remains as 0.1 until the value is actively changed in the settings, then the new value is being picked up by the shortcut
> 2. when FF mode is on, it still blocks shortcuts, despite me putting the call to switch to Normal playback mode

1. Can you clarify a bit more what the problem is? I am failing to understand how that behavior differs from what is expected. I would expect onlyy the setting value to be used.
2. It's not intuitive but `Player.tsx` controls a lot of the logical player state, via a broadcast channel. It looks like the behavior of fast forward mode is implemented in this interval: https://github.com/killergerbah/asbplayer/blob/e2bd2469099f744b702bb46de033e5a201c32d32/common/app/components/Player.tsx#L674

killergerbah:
To explain the behavior you are observing where the setting value doesn't take effect immediately, these are the premises:
- There are a couple sources of settings changes that need to be handled in `VideoPlayer.tsx`:
  - `settings` prop which is updated from `RootApp.tsx`.
  - `miscSettings` (and others) state whose initial state comes from `settings` but where future updates may come from `playerChannel`. 
- `settings` has the following counter-intuitive behavior: In `WebsiteApp.tsx` the `settingsStorage` starts out as the local storage settings but switches to extension storage settings if the extension is installed.
- `useEffect` behaves like this: The function passed to `useEffect` will only run if a change is detected in one of the members of its dependency array.

Consequently:
- To get the latest value of any misc setting you need to refer to `miscSettings`
- `miscSettings` needs to be passed to the `useEffect` dependency array.

artjomsR:
Thank you so much for your detailed response! Should be resolved now but let me know if there's anything else :)

killergerbah:
Thanks again @artjomsR

Issue #378, opened 2024-03-01
'Enter' should confirm subtitle track selection from the subtitle track selector


killergerbah:
From Kayano on Discord:
> Gave up trying to make this sound like a feature request:
I, when I open the select subtitle window on Youtube and the auto-generated subs get chosen, frequently try to press Enter to confirm that selection

PR #445, opened 2024-06-23
Enter confirms subtitle track selection

pooky-programs:
Implementation of https://github.com/killergerbah/asbplayer/issues/378
VideoDataSyncDialog now listens for enter key press when subtitle drop downs aren't in focus.
Can be significantly improved if I can figure out how to shift focus from the side panel to the dialog when it opens on click.
Currently it only submits on enter when the dialog is focused after a click, which I understand is not ideal for what the original issue intended.

### DIFF
--- a/extension/src/ui/components/VideoDataSyncDialog.tsx
+++ b/extension/src/ui/components/VideoDataSyncDialog.tsx
@@ -15,7 +15,7 @@ import makeStyles from '@material-ui/styles/makeStyles';
 import Switch from '@material-ui/core/Switch';
 import LabelWithHoverEffect from '@project/common/components/LabelWithHoverEffect';
 import { ConfirmedVideoDataSubtitleTrack, VideoDataSubtitleTrack } from '@project/common';
-import React, { useEffect, useState } from 'react';
+import React, { useRef, useEffect, useState, FormEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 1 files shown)


killergerbah:
Hi @pooky-programs , sorry for taking a long time to look at this. I thought that your changes were still in-progress since the PR is in draft.

extension/src/ui/components/VideoDataSyncDialog.tsx
```diff
@@ -186,6 +204,8 @@ export default function VideoDataSyncDialog({
                                     return newSelectedSubtitles;
                                 })
                             }
+                            onFocus={() => setIsDropdownActive(true)}
+                            onBlur={() => setIsDropdownActive(false)}
```
killergerbah, line 205: As you pointed out, with this implementation it still takes another click for the dialog to respond to the `Enter` key. I think for this change to be really useful the dialog should close on `Enter` as long as none of the drop-downs are open. This would reduce the total clicks required to select subtitles.

killergerbah (changes requested):
Thanks for your contribution. As mentioned in the comments, I think this change will really be effective when it reduces the clicks necessary for users to select subtitles.

pooky-programs:
@killergerbah no problem, and yeah this is still WIP! I just haven't had the time to sit down and work on it again, sorry for not telling you sooner. I'll be attempting to fix it after July ends.

Issue #451, opened 2024-06-25
Blur shortcuts improvements


artjomsR:
**Is your feature request related to a problem? Please describe.**
With the current implementation, `toggleAsbplayerBlurTrack` blur shortcuts persistently enable / disable the blur. This requires 2 key presses to unblur and then blur the subtitle, which is equivalent to using a shortcut to hide & then display the subtitle

**To distinguish the shortcuts, I suggest one of the following:**
- The `toggleAsbplayerBlurTrack` shortcut unblurs the subtitle but doesn't persist the blur setting. As soon as the subtitle changes, the track becomes blurred again (I think this option is easier and more intuitive. This is also how Language Reactor does this. I'm not aware of other players that provide subtitle blur) 
- `toggleAsbplayerBlurTrack` shortcut unblurs the subtitle as long as the subtitle is held. The subtitle is blurred again after the key is released

tagging @nekorushi in case you're interested in this discussion :)

nekorushi:
That's a cool improvement. I am not sure if that should be a separate set of shortcuts or should it replace the current ones (although i feel that your idea would be utilised more often, so it could replace current ones).
Personally I won't be able to look into it myself due to personal responsibilities, but definitely +1 from me.

killergerbah:
In this scenario, would the settings toggle specify the initial blur state when subtitles load? And the shortcuts only control the blur state during playback?

artjomsR:
Yes, exactly. 

If track blur is **off** in the settings:
blur is permanently off for that track and mouse hover / blur shortcut doesn't change the subtitle appearance

If track blur is **on** in the settings:
- shortcut removes blur for the respective track, but blur is reintroduced when subtitles change
- mouse hover works as is, no changes (unblurs on hover, resumes blur when cursor leaves the sub track)

artjomsR:
>  I am not sure if that should be a separate set of shortcuts or should it replace the current ones

I think it is fine to replace the existing set of shortcuts. Whoever will use the blur subtitles will be using them regularly as part of their workflow, and I'm not expecting them to use the shortcuts to change the type of the flow that they use

In other words, if they're using blur, I'm expecting them to continue using the blur and not use the shortcuts to periodically toggle that setting on and off

killergerbah:
I see, makes sense. I agree this would be better.

artjomsR:
Ok I'll take a look at this then

artjomsR:
[...]

PR #461, opened 2024-07-09
Blur shortcuts improvements

artjomsR:
Implement https://github.com/killergerbah/asbplayer/issues/451

atm I've been only focusing on the extension. Current implementation works, but it's not immediate to unblur / reblur (on subtitle change). I suspect this is due to refreshing the caching, but I've not figured out a way to update it otherwise. So, not sure the current implementation can be improved

Happy if somebody else wants to update / pick this up, as I'm feeling a bit stuck with this as of now

(I've also realised that the renaming of the shortcut broke the untoggling, would need to investigate why. But the PR works as of the 1st commit `3df421e`)


common/locales/de.json
```diff
@@ -103,7 +103,7 @@
         "toggleAsbplayerSubtitleTrack2": "Untertitel 2 in asbplayer umschalten",
         "toggleAsbplayerSubtitleTrack3": "Untertitel 3 in asbplayer umschalten",
         "toggleAsbplayerSubtitleTracks": "Untertitel in absplayer umschalten",
-        "toggleAsbplayerBlurTrack": "Toggle subtitle track {{trackNumber}} blur in asbplayer",
```
killergerbah, line 111: You should keep the old key `toggleAsbplayerBlurTrack` as well since the same loc file is used for all versions of the extension.

extension/src/controllers/subtitle-controller.ts
```diff
@@ -310,21 +312,25 @@ export default class SubtitleController {
                 this.onNextToShow?.(slice.nextToShow[0]);
             }
 
+            let shouldUnblur = false;
+            if (this.subtitleSettings) {
```
killergerbah: This is relatively expensive code to be running ~10 times a second. I think there is a more performant approach.

extension/src/services/key-bindings.ts
```diff
@@ -184,20 +183,12 @@ export default class KeyBindings {
...
-                    sender: 'asbplayer-video',
-                    message: {
-                        command: 'blur-subtitles',
-                        track: track,
-                        trackCount: Math.max(...context.subtitleController.subtitles.map((s) => s.track)) + 1,
-                    },
-                    src: context.video.src,
-                };
-                chrome.runtime.sendMessage(command);
+                context.subtitleController.unblurredSubtitleTracks[track] =
```
killergerbah: Instead of modifying the state of `subtitleController` I suggest implementing a function to trigger the CSS class changes since a function would give you the opportunity to run some code (vs changing internal state and polling for it in a loop).

killergerbah (changes requested):
I think you are close, but I would like to suggest a different overall approach.
As mentioned in the comments, your current approach is:
1. On key-press change internal state.
2. Check state in loop and recalculate classes as necessary.

I would suggest this instead:
1. On key-press call `classList.remove('asbplayer-subtitles-blurred')` on the appropriate subtitle track. Record `unblurredSubtitleTracks[track] = true` only if the class was actually removed.
2. In polling loop add back the class 'asbplayer-subtitles-blurred' for every marked track in `unblurredSubtitleTracks` in the `if (subtitlesAreNew)` block.
3. Empty out `unblurredSubtitleTracks` in `setSubtitleSettings`.

This way you only modify the CSS classes at the moments when they need to be modified and avoid having to recalculate the classes over and over again in the polling loop.

In general, calculating the full state in a loop can be simpler and more bug free but in this case I think the performance implications merit the more event-based approach which isn't actually that much more complex.

common/locales/de.json
artjomsR: reverted the old key

killergerbah:
I pushed what I described as the solution. I took it for granted that you could query the displaying subtitle elements and also know their track number. I think this is where the difficulty was.

artjomsR:
That's great, thank you so much for looking into it! From my point of view, it's functionally complete now, the only thing now that comes to my mind is the translation strings. I've tried to cherry-pick `f232dc80` again, but unblurring still stops working with that commit added, so there's still something missing..

killergerbah:
Thinking about it we should probably add completely new loc strings for this "unblur" behavior. Since it is significantly different from "toggle." But we will need to keep the old loc keys/values for older versions of the extension.

artjomsR:
Do you mean having shortcuts for both toggle and unblur actions? I think _Toggle Unblur_ is too similar to Hide/Show Subtitle Track shortcut (in terms of workflow), but then again nothing's preventing us from having both

killergerbah:
> Do you mean having shortcuts for both toggle and unblur actions? I think _Toggle Unblur_ is too similar to _Hide/Show Subtitle Track_ shortcut (in terms of workflow), but then again nothing's preventing us from having both

Nope, not shortcuts, just new translation (localization) keys. So that we change the labels for the existing shortcuts, while keeping the existing labels for older versions of the extension that have the old toggle behavior.

artjomsR:
I've checked and blur shortcuts worked for me after the rename so should be all good now

killergerbah (approved):

killergerbah:
Thanks again!

Issue #459, opened 2024-07-05
Yomitan not picking up the full sentence with a normal "add". Related to: {cloze-prefix} {cloze-suffix}


Viterkim:
I like having a setup in yomitan with (anki connect)
```
{cloze-prefix} ({cloze-body}) {cloze-suffix}
```
when i mine stuff i have the full sentence and the word in `()`.
However with asbplayer, when there's "2 lines of subtitles" only the line with the actual word in will get mined, this means that the audio will have the full sentence but the sentence is only "half".

Using the asbplayer anki settings doesn't help as i like modifying the content and not just pop in the entire sentence.

Would it be possible to make both sentence lines the same HTML part or something like that? Or is there a way i can make yomitan pickup both sentences?

Thanks!

edit: this used to work fine in https://github.com/animebook/animebook.github.io :D
also would it be possible to disable double click to full screen, or click to play/pause?

edit 2:
disabling double click can be done with a simple violent monkey / tamper monkey script:
```
// ==UserScript==
// @name        Disable oduble click for fullscreen for killergerbah.github.io
// @namespace   Violentmonkey Scripts
// @match       https://killergerbah.github.io/asbplayer/*
// @grant       none
// @version     8.5
// @author      Good old chat gippity
// @description 7/6/2024, 1:33:00 PM
// ==/UserScript==

(function() {
    'use strict';

    document.addEventListener('dblclick', function(event) {
        if (event.target.tagName.toLowerCase() === 'video') {
            event.stopPropagation();
        }
    }, true);
})();
```

killergerbah:
@Viterkim If you 
1. Configure both Yomitan and asbplayer to supply the sentence.
2. And use the "create card with Yomitan, then update with asbplayer" flow.

Then, the asbplayer-supplied sentence should override the Yomitan one, BUT also inherit any HTML markup in the Yomitan sentence. If that's not what is happening then there might be a bug.

Viterkim:
Here's a subtitle file 
[Non Non Biyori-S3 E12-Episode 12_ja.zip](https://github.com/user-attachments/files/16314964/Non.Non.Biyori-S3.E12-Episode.12_ja.zip)

killergerbah:
@Viterkim Got it, I think I understand the problem a bit better now. Yomitan picks up the entire sentence on animebook, I believe due to a combination of how animebook renders the subtitle and how Yomitan detects boundaries between sentences. I can suggest two possible workarounds:

[...]

PR #474, opened 2024-07-25
VideoPlayer.tsx: Don't newline subtitles. (Help yomitan get the entire subtitle)

Viterkim:
Related to the issue i had, after still struggling and not being able to do anything with yomitan, i decided to just fix it here.
https://github.com/killergerbah/asbplayer/issues/459#issuecomment-2242794235

image
image

Works!

### DIFF
--- a/common/app/components/VideoPlayer.tsx
+++ b/common/app/components/VideoPlayer.tsx
@@ -166,7 +166,7 @@ const showingSubtitleHtml = (
     }
 
     const lines = subtitle.text.split('\n');
-    const wrappedText = lines.map(line => `<p class="subtitle-line">${line}</p>`).join('');
+    const wrappedText = lines.map((line) => `<p class="subtitle-line">${line}</p>`).join('');
     return `<span style="${subtitleStyles}" class="${subtitleClasses}">${wrappedText}</span>`;
 };
 
--- a/common/app/components/VideoPlayer.tsx
+++ b/common/app/components/VideoPlayer.tsx
@@ -198,7 +198,12 @@ const ShowingSubtitle = ({
             />
         );
     } else {
-        content = <span style={subtitleStyles}>{subtitle.text}</span>;
+        const lines = subtitle.text.split('\n');
+        content = lines.map((line, index) => (
+            <p key={index} className="subtitle-line" style={subtitleStyles}>
+                {line}
+            </p>
+        ));
     }
 
     return (
--- a/common/app/components/VideoPlayer.tsx
+++ b/common/app/components/VideoPlayer.tsx
@@ -70,7 +70,7 @@ const useStyles = makeStyles({
         paddingRight: 20,
         textAlign: 'center',
         whiteSpace: 'normal',
-        lineHeight: 'normal',
+        lineHeight: 'inherit',
     },
 });
 
--- a/common/app/components/VideoPlayer.tsx
+++ b/common/app/components/VideoPlayer.tsx
@@ -69,7 +69,7 @@ const useStyles = makeStyles({
         paddingLeft: 20,
         paddingRight: 20,
         textAlign: 'center',
-        whiteSpace: 'pre-wrap',
+        whiteSpace: 'normal',
         lineHeight: 'normal',
     },
 });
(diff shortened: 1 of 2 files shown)


Viterkim:
Also this has been driving me crazy, trying to run it locally.
image
The extension doesn't seem to work/thinks its a different site even though I changed this. 

image

the streaming video options also dissapears.

There's also hardcoded things, preventing you from simply running it on localhost, it has to be at localhost/asbplayer
```
    GET http://localhost/asbplayer/static/js/main.62a96a7c.js net::ERR_ABORTED 404 (Not Found)
script.js:1 Ignoring Event: localhost
```
 
Was thinking it would be convenient to run it/test everything + run it locally if internet goes down :)

killergerbah:
Have you tried fixing this problem using the regex-based text replacement feature instead?
Under misc settings you can configure asbplayer to replace all `\n` characters with a space ` `.

About the extension and locally-hosted asbplayer website - prod versions of the extension (like from Chrome web store) do not recognize localhost due to restrictions on the webstore. You can build a dev version of the extension instead if you want it to work with a locally-hosted asbplayer website.

Viterkim:
That makes it into 1 long line yea, but that's definitely worse to read. (I think basically all long subtitles are stacked). So not ideal :/
image


And i guess for making it work offline, it would be way easier to just add a serviceworker and add "progressive web app" support, that way the browser would cache the app even if you don't have internet.

killergerbah:
Have you tried this setting on Yomitan?
image

I am hesitant to make changes just to workaround specific problems with other software.

Viterkim:
Yes i tried that, it adds random text from other sentences to the line.


I'm not really sure what the bad thing about adding it would be :D
I have another friend who has the same issue but who doesn't use github. I think it's valid to be able to get the full sentence without using the extension? (And i think for most people, the only reason why they're using a browser media player, is for pop up dictonaries).

killergerbah:
Okay I see your point.
I'll come back to this - I want to be reasonably sure this change is safe to make.

Viterkim:
> Okay I see your point. I'll come back to this - I want to be reasonably sure this change is safe to make.

Sure! Also don't want to pressure anything! 
Maybe make it an option in the misc settings to be 100% sure? (So maybe default on, then `Don't turn newlines into tags in subtitles` or something

common/app/components/VideoPlayer.tsx
```diff
@@ -164,7 +164,9 @@ const showingSubtitleHtml = (
 `;
     }
 
-    return `<span style="${subtitleStyles}" class="${subtitleClasses}">${subtitle.text}</span>`;
+    const lines = subtitle.text.split('\n');
```
killergerbah, line 168: This change should be repeated on line 198 to get the same behavior when DOM caching is disabled.

killergerbah (changes requested):

common/app/components/VideoPlayer.tsx
Viterkim: image
image

killergerbah (changes requested):
It looks like there's a bug: multiple subtitle tracks are rendered with extra space between them now.

Viterkim:
> It looks like there's a bug: multiple subtitle tracks are rendered with extra space between them now.

Do you have a subtitle file i could try out and see?

Viterkim:
Here's my branch, and master. (With & without the dom option). (No difference)
image
image
image
image

Viterkim:
Or is it this? foun some subtitle online with dual, but it seems really weird
image
seems to only happen on my branch, with `Pre-cache Subtitle DOM` on
image

I'll try to fix it

Viterkim:
image
```
    line-height: normal;
    display: inline;
```
seems to do it

Viterkim:
Actually seems like you can get away with just setting
```
.subtitle-line {
    display: inline;
}
```

I don't know what's best :)

Viterkim:
image
ok should work now

Viterkim:
image

killergerbah:
Can you run Prettier on VideoPlayer.tsx? I can merge after that.

killergerbah:
Thanks!

PR #489, opened 2024-08-08
add default policy if one doesn't exist

s-cork:
This seems to fix the issue reported on discord

I don't know if this is a chrome bug or something else
It seems without the default policy that chrome is treating the trustedScript as a string
would be nice to know why
With the default policy added, then the default policy is used


s-cork:
It looks like it's a chrome bug:
https://wpt.live/trusted-types/eval-function-constructor.html

In my browser the last test fails

killergerbah:
Thanks @s-cork !

Issue #442, opened 2024-06-19
[Feature Request] Add automatic subtitle detection for Jellyfin


jonathanluepfert:
Jellyfin has some advantages over Emby. Maybe it would not be that difficult to implement subtitle detection for it since it is quite similar to Emby.

PR #505, opened 2024-08-25
WIP: Add Jellyfin support

shekhirin:
Closes https://github.com/killergerbah/asbplayer/issues/442

The API itself is very similar to Emby, with minor changes to `ApiClient`.

TODO:
- When the subtitle selection dialog is opened before subtitles are loaded, entries will have a spinner until you re-open the dialog
- Auto-pause works, but auto-play doesn't. Video element detection seems to work (because pause works), but hover event listener does not.
- Subtitles are hovering on top of media progress and controls bar image
- Jellyfin `pages.json` entry shouldn't conflict with Emby, because `hash` is different, but I didn't test it.
- I did a hacky timeout with five retries to fetch the subtitles, but I'm not sure if that's the best way to do it. I see that `netflix-page.ts` has the same approach.

Other than that, it works pretty well 🙂
image

### DIFF
--- a/extension/src/manifest.json
+++ b/extension/src/manifest.json
@@ -108,6 +108,7 @@
                 "pages/viki-page.js",
                 "pages/unext-page.js",
                 "pages/emby-page.js",
+                "pages/jellyfin-page.js",
                 "pages/osnplus-page.js",
                 "pages/bilibili-page.js",
                 "pages/nrk-tv-page.js",
--- a/extension/src/pages.json
+++ b/extension/src/pages.json
@@ -91,6 +91,14 @@
             }
         },
         {
+            "host": "jellyfin*|:8096*",
+            "script": "jellyfin-page.js",
+            "hash": "#/video",
+            "autoSync": {
+                "enabled": true
+            }
+        },
+        {
             "host": "emby*|:8096*",
             "script": "emby-page.js",
             "hash": "#!/videoosd/videoosd.html",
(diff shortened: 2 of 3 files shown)


killergerbah:
Thanks for your work! Just some general comments on the TODOs you listed:

> When the subtitle selection dialog is opened before subtitles are loaded, entries will have a spinner until you re-open the dialog

This is a problem in general with the subtitle track selector dialog. Not necessary to fix with the functionality you're adding but definitely a welcome fix it you'd like to attempt it.

> Auto-pause works, but auto-play doesn't. Video element detection seems to work (because pause works), but hover event listener does not.

Auto-play works by detecting hover on top of the video element, in order to deal with the case where users are mousing inside a popup dictionary like Yomitan (don't want to auto-play in that case). So I suspect there is something about Jellyfin making the video hover detection not work.

> Subtitles are hovering on top of media progress and controls bar 

This is a side effect of how subtitles are rendered in general in order to keep the subtitles text selectable and I don't think it's necessary to fix. Users have the option to adjust the position of the subtitles.

bpwhelan:
@shekhirin Hey, would you consider this still WIP? I added Emby support to asbplayer and all the things you listed are "technically" issues in Emby too.

shekhirin:
> @shekhirin Hey, would you consider this still WIP? I added Emby support to asbplayer and all the things you listed are "technically" issues in Emby too.
> 
> 
> 
> 

I see, then it's ready for review 😄

bpwhelan:
One quick question, was the timeout thing a response to a problem you were having? Sometimes subs don't load on emby and I have to go back and restart the stream.

killergerbah (approved):

killergerbah:
Thanks!

PR #508, opened 2024-09-02
Top / bottom setting for individual subtitles

artjomsR:
Mostly working for the extension and still needs cleaning up, but I wanted to ask first: is there any penalty for always having 2 subtitle containers (one at top and one at bottom), even if one of them is not used? E.g., all subtitles are displayed at the bottom?


killergerbah:
> is there any penalty for always having 2 subtitle containers (one at top and one at bottom), even if one of them is not used?

Sorry for the really late response on this. I probably clicked through this PR and then forgot to reply.
I don't think there should be a huge penalty. My guess is that the DOM cache will be the biggest drain on resources, but also probably not a big deal. But it would be good to avoid rendering the subtitles unnecessarily. E.g. if only top or only bottom then only render top or only render bottom.

artjomsR:
no problem, thanks for getting back to me! I've updated the PR now, let me know what you think :)

extension/src/controllers/subtitle-controller.ts
```diff
@@ -113,36 +118,46 @@ export default class SubtitleController {
...
         }
 
-        this._resetOverlays(this._subtitleAlignment, preCacheDom);
+        this._resetOverlays(preCacheDom);
         this._preCacheDom = preCacheDom;
     }
 
     cacheHtml() {
-        if (!(this.subtitlesElementOverlay instanceof CachingElementOverlay)) {
+        if (!(this.bottomSubtitlesElementOverlay instanceof CachingElementOverlay) || !(this.topSubtitlesElementOverlay instanceof CachingElementOverlay)) {
```
killergerbah: Currently, the check assumes that both overlays are always the same type, which may be true and enforced internally by this class.

But since it does not hurt, I think to be safe the `instanceof` check should be done independently for `bottomSubtitlesElementOverlay` and `topSubtitlesElementOverlay` before calling `uncacheHtml`.

extension/src/controllers/subtitle-controller.ts
```diff
@@ -154,6 +169,22 @@ export default class SubtitleController {
             this.cacheHtml();
         }
 
+        const newAlignments = allTextSubtitleSettings(newSubtitleSettings).map((s) => s.subtitleAlignment);
+        if (newAlignments !== this.subtitleTrackAlignments) {
```
killergerbah: This check will always return `true` because JavaScript will use object identity for comparison. This class already implements `_arrayEquals`, you can use that instead.

extension/src/controllers/subtitle-controller.ts
```diff
@@ -169,21 +200,8 @@ export default class SubtitleController {
...
-        this._applyElementOverlayParams(this.subtitlesElementOverlay, subtitleOverlayParams);
-        this._applyElementOverlayParams(this.notificationElementOverlay, notificationOverlayParams);
-        this.subtitlesElementOverlay.hide();
-        this.notificationElementOverlay.hide();
-        this._subtitleAlignment = value;
-    }
-
-    get subtitleAlignment() {
-        return this._subtitleAlignment;
+    getSubtitleTrackAlignment(trackIndex: number) {
```
killergerbah: Can be `private`

extension/src/controllers/subtitle-controller.ts
```diff
@@ -451,7 +473,10 @@ export default class SubtitleController {
             this.notificationElementOverlayHideTimeout = undefined;
         }
 
-        this.subtitlesElementOverlay.dispose();
+        if (this.shouldRenderBottomOverlay)
```
killergerbah: Skip the `should*` checks - just dispose both overlays.

common/settings/settings.ts
```diff
@@ -149,7 +150,8 @@ const subtitleSettingsKeysObject: { [key in keyof SubtitleSettings]: boolean } =
     subtitleCustomStyles: true,
     subtitleBlur: true,
     imageBasedSubtitleScaleFactor: true,
-    subtitlePositionOffset: true,
+    bottomSubtitlePositionOffset: true,
```
killergerbah: Use `subtitlePositionOffset` instead of `bottomSubtitlePositionOffset` and add a comment saying it's the bottom offset. The naming is less clear but without a system to migrate the settings structure it's safer to preserve older names, since the extension and app share settings.

killergerbah (changes requested):
As always thank you. It took me a while to get to this because of work and also trying to get 1.5.0 out first.

artjomsR:
No problem for the delay, there's no rush :) 

I've updated the PR, should've addressed all your comments. You didn't address my previous comment, not sure if you have an opinion about this?

> I'm also thinking if we should remove the if checks from shouldRenderBottomOverlay (and remove it altogether), leaving only shouldRenderTopOverlay: boolean;: I think users would use either only bottom or bottom + top subtitles. But I don't see why someone would use only top subtitles

killergerbah:
> No problem for the delay, there's no rush :)
> 
> I've updated the PR, should've addressed all your comments. You didn't address my previous comment, not sure if you have an opinion about this?
> 
> > I'm also thinking if we should remove the if checks from shouldRenderBottomOverlay (and remove it altogether), leaving only shouldRenderTopOverlay: boolean;: I think users would use either only bottom or bottom + top subtitles. But I don't see why someone would use only top subtitles

Sorry forgot to reply to that as well. There are two reasons why I think we should leave it as is:

- It's not obvious to me how we would work that into the current settings UI. For example you would have to prevent users from switching all tracks to the top. It would seem then the track specific settings would have to be aware of each other which might introduce complexity.
- There are users who use top only. For example to add second language subtitles to websites that already have their own subtitles.

artjomsR:
I meant to keep the functionality as is, simply skip the `if shouldRenderBottomOverlay` check and always invoke all functionality related to the bottom overlay. Once again, based on the principle that _most_ people would use the bottom container. 

But as you say, if considerable number of users use top only, then we can leave everything as is :)

scripts/anki-connect-proxy/anki-connect-proxy.exe
killergerbah, line 1: I think this file should be part of a different PR.

killergerbah (commented):
It looks like you removed the `subtitleAlignment` setting. If you're not using that setting, how are you intending for users to be able to pick the alignment for individual tracks?

artjomsR:
> It looks like you removed the `subtitleAlignment` setting. If you're not using that setting, how are you intending for users to be able to pick the alignment for individual tracks?

in common/settings/settings.ts, I've moved it from SubtitleSettings (which I understood as "global" settings) to TextSubtitleSettings (which I understood as settings which apply to individual tracks), which is effectively all this PR is about 

Uploading Replay_2024-10-12_10-26-57.mp4…

killergerbah:
Understood!

extension/src/controllers/subtitle-controller.ts
```diff
@@ -160,6 +172,27 @@ export default class SubtitleController {
             this.cacheHtml();
         }
 
+        const newAlignments = allTextSubtitleSettings(newSubtitleSettings).map((s) => s.subtitleAlignment);
```
killergerbah, line 175: I found a bug that I believe has to do with this line. See the video below. You'll notice that there are two tracks, but when switching all tracks, or track 0, to the 'top' alignment, one of the tracks disappears.

https://github.com/user-attachments/assets/045baef8-39f8-4d54-a932-c5d93a17dac1

Despite the name,`allTextSubtitleSettings` doesn't return the `TextSubtitleSettings` for a specific number of tracks, it only returns settings for the ones that are actually configured.

This means that we need to account for `TextSubtitleSettings` missing for some tracks. You'll notice that `SubtitleController._subtitleStyles` does this by assuming any missing track-specific settings should just inherit the settings for track 0. Unfortunately this means that calculating `shouldRenderBottomOverlay` and `shouldRenderTopOverlay` will always depend on the `_subtitles` field and we would need to re-calculate those fields whenever `_subtitles` changes.

So for now I suggest the following:
1. Remove `shouldRenderBottomOverlay` and `shouldRenderTopOverlay` fields. Always cache HTML for both overlays.
2. (Optional) Refactor `CachingElementOverlay` so that `topSubtitlesElementOverlay` and `bottomSubtitlesElementOverlay` can share the same DOM cache. This will cancel out the overhead of always having two overlays.

extension/src/controllers/subtitle-controller.ts
```diff
@@ -175,21 +208,8 @@ export default class SubtitleController {
...
-        this._applyElementOverlayParams(this.notificationElementOverlay, notificationOverlayParams);
-        this.subtitlesElementOverlay.hide();
-        this.notificationElementOverlay.hide();
-        this._subtitleAlignment = value;
-    }
-
-    get subtitleAlignment() {
-        return this._subtitleAlignment;
+    private _getSubtitleTrackAlignment(trackIndex: number) {
+        return this.subtitleTrackAlignments[trackIndex] || 'bottom';
```
killergerbah: As mentioned above the default value should really be the setting for the 0th track:

```
this.subtitleTrackAlignments[trackIndex] || this.subtitleTrackAlignments[0]
```

killergerbah (changes requested):

artjomsR:
I've added your suggested change of `|| this.subtitleTrackAlignments[0]` which seems to have fixed the bug for me, I've tested several times. So I'm not sure if the caching changes are still required now

killergerbah:
Sorry, yeah you're right. Thanks again for your patience.

Issue #465, opened 2024-07-17
Emby - IF Language is unknown, fails to import.


bpwhelan:
I will likely fix this since I added Emby support.

![image](https://github.com/user-attachments/assets/5280f241-0c24-4d6d-b91f-4f7809983675)

PR #514, opened 2024-09-11
Emby - Add Retry and fix unknown language

bpwhelan:
Updated to use the same retry technique as [this PR](https://github.com/killergerbah/asbplayer/pull/505), fixes where sometimes the session can't be found and the subtitle popup just spins (Never made an issue).

Also closes https://github.com/killergerbah/asbplayer/issues/465


killergerbah (approved):
Thanks!

PR #520, opened 2024-09-21
Combine Emby and Jellyfin

bpwhelan:
Having two different pages is problematic without ripping apart the page detection logic, since they both default to :8096. I should have done this when I added Emby in the first place, but I was a bit lazy and didn't feel like standing up a JF instance to test.

These two use basically the same exact stuff. The only difference is line 43, which is nicely taken care of with ??. Also apikey is not needed to retrieve the subtitle stream it seems, which makes sense really.

Evidence of both working locally using the same page code. Both of these are running on :8096 on different servers.

image_1

image


bpwhelan:
Looping @shekhirin in since he added the Jellyfin code.

TL;DR, Having them as two seperate pages was breaking things. Emby and Jellyfin only differ on like, one line, so just merged the two.

killergerbah:
Awesome thanks

Issue #531, opened 2024-10-06
Plex Support


ShanaryS:
**Is your feature request related to a problem? Please describe.**
When using Plex, we need to manually use a subtitle file from disk, instead using the ones with the Plex media.

**Describe the solution you'd like**
Automatic detection of subtitles when using Plex, similar to Emby and Jellyfin.

**Describe alternatives you've considered**
None.

**Additional context**
I'm not familiar with extension development but I was looking into how this can be done. Unfortunately, it seems plex doesn't have the ApiCilent like Emby/Jellyfin which makes it harder.

I did see the the request plex uses for the subtitles by monitoring network when switching the subtitles. The request with the minimally required params:
`https://127.0.0.1:32400/video/:/transcode/universal/subtitles?path=%2Flibrary%2Fmetadata%2F51494&protocol=dash&session=abc&X-Plex-Platform=Chrome&X-Plex-Token=abc`
Removing `protocol=dash` will load the entire subtitle, albeit slowly. Plex seems to only provide the subtitles as far as the video is buffered with dash enabled. `/library/metadata/123` is just the ratingKey for the media, which is Plex's unique id for content at least at a per library level. Since this requires session, it will only get the subtitle currently selected. I don't believe there is a way to get all but this isn't a big deal.

The only problem here getting a token which I believe there are two choices. Ask the user for a plex token that asbplayer stores, this is how apps usually interact with plex. This would work for non admin users since anyone can monitor their request traffic to get it. The alternative is asbplayer monitoring the requests on the page which I think is possible but I'm not sure how that will affect the extensions permissions.

---

I've done some more research into this. All external subtitles can be downloaded directly at any time using `https://127.0.0.1:32400/library/streams/123&X-Plex-Token=abc` where `123` is the subtitle stream id. We only need the `token` and `ratingKey` for external subtitles. All metadata is stored for each subtitle as well.

However internal subtitles can only be downloaded when it's in an active session and not being burned in. Luckily the selected boolean is stored with the subtitle metadata and the user can disable burn-in in their plex settings (for non-image formats). Internal subtitles requires `token`, `ratingKey`, and `session`.

In short, this is definitely possible and fairly straight forward, but the concern is how to get the `token`, `ratingKey`, and `session`. I'm not familiar with extensions so I don't know how to dynamically get this data, though the `token` can probably just be a setting.

---

[...]

PR #533, opened 2024-10-08
Support external subtitles for plex

ShanaryS:
Fixes: #531 

Tested on Chrome and Firefox with admin and non-admin accounts.

Supports external subtitles as well as subtitles from Plex's subtitle search feature. However internal subtitles don't work as we need the `session` id. I cannot find this anywhere as I did with the other values. The only way might be to monitor requests as plenty contain it, but that would require the `webRequest` permission on the extension. Even if internal subtitles are supported, they would work only if currently selected and not being burned in at the time the user loads the subtitles.

---

It seems downloading the subtitles from the session is finnicky. So even if we were to have all the information, the endpoint wasn't really designed for full downloads so would probably need some extra care. So I don't think it's worth trying to get internal subtitles to work at least for now. The way external subtitles are being downloaded is using an endpoint designed for downloading.

### DIFF
--- a/extension/src/manifest.json
+++ b/extension/src/manifest.json
@@ -111,6 +111,7 @@
                 "pages/osnplus-page.js",
                 "pages/bilibili-page.js",
                 "pages/nrk-tv-page.js",
+                "pages/plex-page.js",
                 "anki-ui.js",
                 "mp3-encoder-worker.js",
                 "pgs-parser-worker.js",
--- a/extension/src/pages.json
+++ b/extension/src/pages.json
@@ -125,6 +125,14 @@
             "autoSync": {
                 "enabled": true
             }
+        },
+        {
+            "host": "plex*|:32400*",
+            "script": "plex-page.js",
+            "path": ".*",
+            "autoSync": {
+                "enabled": true
+            }
         }
     ]
 }
--- a/extension/src/pages/plex-page.ts
+++ b/extension/src/pages/plex-page.ts
@@ -1,0 +1,146 @@
+import { VideoDataSubtitleTrack } from '@project/common';
+import { VideoData } from '@project/common';
+import { trackFromDef } from './util';
+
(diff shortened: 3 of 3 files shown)


killergerbah (changes requested):
Thank you for your contribution!
I'm not familiar with Plex but I tested this out using a local Plex server and wasn't able to get it to work immediately. This line:

```
const imgTags = document.querySelectorAll('img[src*="plex.direct"]');
```

excludes image tags with source URLs that look like `http://127.0.0.1/...`, which is what they looked like when I was testing.

My suggestion is to change the overall approach to one that depends less on the specific UI elements on the page and instead inspect the network traffic directly. For example, on my computer Plex is already automatically downloading the metadata in JSON. If you filter by "metadata" in your browser's network traffic inspector then you will see it.

Therefore, I believe Plex will be compatible with the JSON hijacking approach that is used in other asbplayer scripts. The server URL can be derived by hijacking the `fetch` call that fetches metadata. See `nrk-tv-page.ts` for an example of how `JSON` and `fetch` hijacking are combined to get subtitle data.

ShanaryS:
Hijacking is definitely much better, this is how I originally wanted to do it. External subtitles should hopefully work for you now.

It now allows getting the subtitle stream but it still is unusable currently. Trying to download the entire subtitle at once doesn't seems brittle, so the best way forward would be to support streaming subtitles in the extension and let it update as the video plays. If the video was resumed midway through, then the subtitles before would be missing. Also would probably need to handle seeking around. It wouldn't be worth going for internal subtitles without these. If it is to be implemented in the future, the commented code should suffice with the streaming being handled elsewhere.

killergerbah:
Thanks, subtitle detection is working for me now.
It is not likely that subtitle streaming is implemented any time soon but thanks for the note on internal subtitles.

PR #570, opened 2024-11-23
fix(subtitle-reader): Handle a file with an extension .subrip as with .srt

miroshQa:
I downloaded a video from a torrent, and opened it in jelly, for some reason, when I wanted fetch  subtitle track, I  get the error "Unsupported subtitle format", I opened debug and saw that the file format was ".subrip" for some reason, so I added handling for this case as well
image


killergerbah:
Thanks!

PR #571, opened 2024-11-24
fix: Building extension script in directories with spaces

NalinGHub:
The current build script doesn't account for directories with spaces due to how `pwd` doesn't escape space characters returned and how the `pushd` command accepts arguments. Simple fix is to quote the arg.

Also just added a random lint fix that I ran into when peeking around the codebase lol.


killergerbah (approved):

killergerbah:
Thanks!

Issue #521, opened 2024-09-21
Display Youtube Auto-translate Subtitles


aadi-333:
Hi! I've seen in a couple of other similar programs, language reactor and mouse tooltip translator, that loaded both the subtitles and youtube auto translate subtitles. This would be a highly useful feature for me if implemented. Thank you for your consideration.

Issue #460, opened 2024-07-08
Load youtube's secret translated subtitles


cspotcode:
**Is your feature request related to a problem? Please describe.**

Youtube generates "auto-translate" subtitles, but I cannot choose them in ASBPlayer's dropdown.

I would like to select these auto-translated subtitles for track 2 / dual subs in ASBPlayer.

**Describe the solution you'd like**

Expose these auto-translated subtitle languages as an option in ASBPlayer's "Load subtitles" dropdown.

**Describe alternatives you've considered**

Somehow downloading these subs manually and drag-and-dropping them into ASBPlayer.

**Additional context**

I've been using this browser extension which automatically loads and displays dual subtitles on YouTube. 
 https://chromewebstore.google.com/detail/mouse-tooltip-translator/hmigninkgibhdckiaphhmbgcghochdjc

It does this even for videos which only list one subtitle language because -- I assume -- it downloads YouTube's "auto-generated" subtitles, too.

This is great, except it doesn't have any of ASBPlayer's other features.

It would be great if ASBPlayer had the option of loading these auto-translated subtitles as dual subs / track 2.

I believe this is the code where it detects user subs in the desired language first, then falls back to auto-generated: https://github.com/ttop32/MouseTooltipTranslator/blob/c888f621d0286fa887d911f0f878cad58ddb3fe2/src/subtitle/youtube.js#L133-L143

I don't need ASBPlayer to do anything magical, but it would be nice if these auto-generated subs were *somehow* available in ASBPlayer's dropdown.

PR #609, opened 2025-01-04
Add Machine Translated Subtitles to YouTube

IanDesuyo:
Implementation of #460 and #521

By adding `&tlang=LANG_CODE` after `track.baseUrl`, we can get machine translated subtitles for the videos.

This PR is not yet complete. I will continue to work on it.

TODO:

- [x] Get machine translated subtitles
- [ ] Select target language
  - [ ] Language selection in settings UI (or depending on settings/YouTube language?)
  - [ ] Pass language into `youtube-page.ts`

Preview:
image
image


JeffAlyanak:
Definitely a cool feature! This may already be on your radar, but I'd suggest making the drop-down option more clearly labeled as auto-translated.

The quality of auto-translation can be really low for some languages and I think it's important that users are not unaware that they're selecting an auto sub track.

Issue #620, opened 2025-01-15
Audio not playing when pressing the 'Update Last Card' button


NirDafnai:
**What is the issue and what do you do to make it happen**
Usually when I press update last card, it previews the audio so I could know if it recorded it properly (timing wise). For some reason now it won't preview the audio. Could it be an issue on my end?

**What browser?**
Chrome 131.0.6778.265 64bit
**Is it happening with local video or streaming video?**
Local Video (H264 AC3)

nikkovc:
I'm having the same issues. It was working with the same files and browser a few days ago.

nikkovc:
Oh I can just click on the audio box. I didn't realize this, thanks! Having the option for the automatic option could still be nice though. Thanks again!

NirDafnai:
Hey, I created a branch that makes it configurable in the settings, however I don't have permissions to push that branch to the remote. Could you add permissions for me and then I'll be able to create a PR? 
@killergerbah

killergerbah:
@NirDafnai Create a [Pull Request](https://docs.github.com/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests) and I'll review the change.

NirDafnai:
Fixed in https://github.com/killergerbah/asbplayer/pull/625

PR #625, opened 2025-01-20
Add 'record with playback' to settings menu

NirDafnai:
Fixes issue https://github.com/killergerbah/asbplayer/issues/620

It's now configurable in settings menu:
image


common/locales/en.json
```diff
@@ -310,6 +310,7 @@
         "preCacheSubtitleDom": "Pre-cache Subtitle DOM",
         "preCacheSubtitleDomHelperText": "If enabled, asbplayer will pre-render subtitle text elements in an offscreen element, and re-use those elements for subtitle display. This allows external code to persistently modify subtitle text before it is displayed.",
         "recordingBind": "Recording",
+        "recordWithAudioPlayback": "Record with audio playback",
```
killergerbah: This copy should specify that it only applies to local video.
Something like: "Play audio while recording audio from local video"

killergerbah (changes requested):
Please update the `AudioClip.fromCard` call in `handleCopy` in `App.tsx` as well. That will ensure that this setting behaves consistently in all cases.

common/locales/en.json
NirDafnai: Updated

NirDafnai:
> Please update the `AudioClip.fromCard` call in `handleCopy` in `App.tsx` as well. That will ensure that this setting behaves consistently in all cases.
Fixed

common/components/AnkiDialog.tsx
```diff
@@ -352,7 +352,7 @@ const AnkiDialog = ({
     ]);
 
     const initialAudioClip = useMemo(
-        () => AudioClip.fromCard(card, settings.audioPaddingStart, settings.audioPaddingEnd, false),
+        () => AudioClip.fromCard(card, settings.audioPaddingStart, settings.audioPaddingEnd, settings.recordWithAudioPlayback),
```
killergerbah: Missed this on the first pass. You need to update the dependency array being passed to useMemo to ensure the audio clip changes whenever the setting changes.

common/components/AnkiDialog.tsx
```diff
@@ -602,8 +602,8 @@ const AnkiDialog = ({
     const handleTrack3TextChange = useCallback((text: string) => {
         setTrack3(text);
         setLastAppliedTimestampIntervalToText(undefined);
-    }, []);
-
+    }, []);   
+
```
killergerbah: The extra space will fail the Prettier formatting check on this repo. Easiest way to fix this is run Prettier on the file whenever you make a change.

common/app/components/App.tsx
```diff
@@ -426,7 +426,7 @@ function App({ origin, logoUrl, settings, extension, fetcher, onSettingsChanged,
                         newCard,
                         settingsRef.current.audioPaddingStart,
                         settingsRef.current.audioPaddingEnd,
-                        true
+                        settings.recordWithAudioPlayback
```
killergerbah: Here, use settingsRef to access the setting.

killergerbah (changes requested):

common/app/components/App.tsx
NirDafnai: Fixed

common/components/AnkiDialog.tsx
NirDafnai: Whoops, fixed!

common/components/AnkiDialog.tsx
NirDafnai: I'm not sure I understand, I added now the setting to the array below this line if that's what you meant. Correct me if i'm wrong

killergerbah: Your change is the right one. If you're interested in learning why you can look up react hooks and dependency arrays.

NirDafnai:
Forgot to add de localization, try running workflow now

NirDafnai:
Okay fixed formatting, now it should work.
sorry for the back and forth, this is my first open-source contribution :O

killergerbah:
Thanks!

PR #647, opened 2025-02-08
App.tsx: support .m4v media files

Viterkim:
m4v is the same as an mp4 basically. 
Renaming the file to mp4 and loading it in works, but is annoying.

(To avoid this)
image


killergerbah:
This should be out now

killergerbah (approved):

PR #661, opened 2025-02-24
Fix Jellyfin Subtitles

MarvNC:
For some subtitle types like `.mov_text` Jellyfin will return an error bad request. Luckily we can just request `.srt` and Jellyfin will produce one, even if the subtitle wasn't originally `.srt`. Not sure if it works with all types of subs though, might need some testing.


killergerbah:
It seems like asking Jellyfin to convert to `srt` has more predictable behavior than just passing the original "codec" through. I don't have time to test this thoroughly myself,  but theory this should always work according to this: https://github.com/jellyfin/jellyfin/blob/efb901c36976c006f2b9ad8420a06181819c9016/MediaBrowser.MediaEncoding/Subtitles/SubtitleEncoder.cs#L154

Thanks!

PR #664, opened 2025-02-27
Websocket - Sync timestamp command

mcgrizzz:
This adds a new WebSocket command to sync to a specific timestamp in the video. I’ve tested it locally in Chrome and followed the `load-subtitles` implementation to keep it consistent with existing code.

I’m building a vocabulary mining program that scans subtitles to find words and phrases you haven’t mined yet. This feature makes it convenient—users can click a timestamp in the program and jump straight to that spot in the video to mine it with asbplayer. I’m sure other projects could find this useful too.

### DIFF
--- a/common/web-socket-client/web-socket-client.ts
+++ b/common/web-socket-client/web-socket-client.ts
@@ -32,6 +32,14 @@ export interface LoadSubtitlesCommand {
     };
 }
 
+export interface SyncTimestampCommand {
+    command: 'sync-timestamp';
+    messageId: string;
+    body: {
+        timestamp: number;
+    };
+}
+
 export class WebSocketClient {
     private _socket?: WebSocket;
     private _pingInterval?: NodeJS.Timeout;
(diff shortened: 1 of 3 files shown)


common/web-socket-client/web-socket-client.ts
```diff
@@ -32,6 +32,14 @@ export interface LoadSubtitlesCommand {
     };
 }
 
+export interface SyncTimestampCommand {
+    command: 'sync-timestamp';
```
killergerbah: Can you change this to 'seek' instead? It is a little clearer IMO.

killergerbah (changes requested):

common/web-socket-client/web-socket-client.ts
mcgrizzz: Just pushed a commit to refactor sync into seek. I've also added the corresponding readme documentation.

killergerbah:
Just run prettier on your changes and should be good to merge

killergerbah:
Thanks!

Issue #675, opened 2025-03-19
Websocket will not reconnect on error or dropped connection


mcgrizzz:
**Before submitting an issue check out the common issues section of the README:**
https://github.com/killergerbah/asbplayer?tab=readme-ov-file#common-issues

**What is the issue and what do you do to make it happen**
If a websocket connection drops, you have to restart asbplayer by reopening the side-panel or the app before it will reconnect to the websocket. The PING PONG method doesn't work for dropped connections or errors.  

**What browser?**
Chrome (Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36)

**Is it happening with local video or streaming video?**
I've tested this with streaming video. 

**If the issue is happening with streaming video, what version of the extension are you using?**
The latest build from the source. 

**If the issue is happening with streaming video, on what website(s) is the issue happening?**
Any site

I'm working on a PR to fix this issue.

PR #676, opened 2025-03-19
Adds Websocket Auto-reconnect functionality

mcgrizzz:
This PR updates the websocket client with an auto-reconnect feature. If the connection drops, it will now attempt to reconnect automatically. You can set how many times it tries and how long it waits between attempts.  

Fixes issue #675.


mcgrizzz:
I'm just wondering what the process is for generating the new locales. 

Do I just run merge-loc-from-en.js then commit all those files, including the files in the "untranslated" folder?

killergerbah:
@mcgrizzz Thanks for working on this.

About localization, you would run `node mergee-loc-from-en.js` and just commit those changes. No need to deal with "untranslated" strings.

Can you explain why these new settings will add value to the user? I think with reasonable defaults the original problem will be 99% solved, which is that the client reconnects in a reasonable amount of time. I can't think of really good reasons why most users would want anything other than infinite retries on a reasonable interval like 10 seconds or so.

mcgrizzz:
I can't come up with a strong case either. My initial thought was that some users might not want constant reconnect attempts, but on reflection, that’s not an issue—if the WebSocket functionality is off, it won’t try to reconnect anyway.

I’ll tweak it to remove the new settings and just stick with infinite retries every 10 seconds. That should handle the original problem cleanly.

killergerbah:
Is there a specific case that auto-reconnect is catching, which isn't caught by the ping-pong loop + line 66? https://github.com/killergerbah/asbplayer/pull/676/files#diff-0e41db57a39b4bc7c624490b63c88026e5daae8d9cf27c2f65d48021590791c0L66

I tried reproducing the dropped connection with the ping/pong loop and I couldn't get it to break. I added the above line recently so I want to confirm that you were reproducing the problem even with that line.

mcgrizzz:
Thanks for testing this. I tried to reproduce the issue on the latest build but couldn’t replicate it. 

It seems I may have encountered this behavior earlier, before I synced my fork and began working on this PR, and I didn’t realize the change had been introduced in the meantime. 

Apologies for the oversight!

killergerbah:
Oh no problem thanks for taking the time to work on this and sorry about the unlucky timing. Looking forward to your future contributions.

PR #678, opened 2025-03-19
Fix Local File Websocket Seek

### LINKED ISSUE #677: Websocket seek does not work in local player

**What is the issue and what do you do to make it happen**
Open a local video in the app and try to use the websocket to seek to a timestamp. It will not seek. 

**What browser?**
Chrome (Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36)
Local video

N/A
N/A

I am using the latest build from source. The issue is the early return in [Player.tsx](https://github.com/killergerbah/asbplayer/blob/c05636df640232dc4ab9eefda38222de35d2218c/common/app/components/Player.tsx#L965) 

The problem is even though extension.supportsWebSocketClient is true, if we are playing a local video file, the websocket is not handled elsewhere. 

I've fixed it on my branch and will make a PR shortly.

mcgrizzz:
Fixes #677

### DIFF
--- a/common/app/components/Player.tsx
+++ b/common/app/components/Player.tsx
@@ -962,16 +962,22 @@ const Player = React.memo(function Player({
     }, [clock, rewindSubtitle?.start, mediaAdapter, seek]);
 
     useEffect(() => {
-        if (!webSocketClient || extension.supportsWebSocketClient) {
-            // Do not handle mining commands here if the extension supports the web socket client.
-            // The extension will handle the commands for us.
+        // Return early if:
+        // 1. No WebSocket client, OR
+        // 2. Extension supports WebSocket clients AND this is NOT a local video
+        const isLocalVideoFile = Boolean(videoFileUrl);
+        console.log("Video file: " + videoFile);
+        if (!webSocketClient || (extension.supportsWebSocketClient && !isLocalVideoFile)) {
             return;
         }
-
+        
+        // Handle WebSocket commands ourselves when:
+        // - Extension doesn't support WebSocket clients, OR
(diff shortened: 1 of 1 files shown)


common/app/components/Player.tsx
```diff
@@ -962,16 +962,18 @@ const Player = React.memo(function Player({
...
 
     useEffect(() => {
-        if (!webSocketClient || extension.supportsWebSocketClient) {
-            // Do not handle mining commands here if the extension supports the web socket client.
-            // The extension will handle the commands for us.
+        // Return early if:
+        // 1. No WebSocket client, OR
+        // 2. Extension supports WebSocket clients AND this is NOT a local video (otherwise the extension will handle it)
+        const isLocalVideoFile = Boolean(videoFileUrl);
+        if (!webSocketClient || (extension.supportsWebSocketClient && !isLocalVideoFile)) {
```
killergerbah: On reviewing the existing implementation and behavior the actual correct change here is to remove the `supportsWebSocketClient` check entirely. The extension is not forwarding "load-subtitle" nor "seek" commands to the asbplayer app, because there's no really good reason for it to, unlike "mine-subtitle."

So the check should be 
```
if (!webSocketClient) {
  return;
}
```

killergerbah (changes requested):

mcgrizzz:
Fixed.

killergerbah (approved):

killergerbah:
Thanks for noticing this and patching it up!

PR #686, opened 2025-03-29
Add setting to control how HTML in subtitles is handled

kowasaur:
Currently, whether HTML tags are removed from subtitles depends on the file type. Also, for subtitles where HTML tags are not removed, the HTML is not rendered properly in many places.

This pull request:
- Adds a setting to allow the user to decide whether HTML found in subtitles is either removed or rendered. The file type no longer decides this.
- Renders HTML tags properly in _all_ places that subtitles are rendered (I'm pretty sure, but I may have forgotten some places).

This is what the setting looks like
setting

The following examples use this SRT file
```
1
00:00:01,000 --> 00:00:02,000
この<ruby>漢字<rt>かんじ</rt></ruby>はちょっと... Here's some <span style="color: yellow">colour</span>.
```

This is the video player with "remove" selected
remove video player

This is the video player with "render" selected
render video player

Sorry that I didn't add anything to `loc/untranslated/*.json`. I tried running `scripts/loc/seed-untranslated.js` but it added a lot more than the things I added so I thought that's probably not what I'm supposed to do.


killergerbah:
@kowasaur Thank you for spending the time to do this. There is just one thing I want to check, which is that it's fine to filter out the HTML for any file format. I don't remember exactly why, but my guess is that I was filtering depending on file extensions since it only makes sense for HTML to appear for particular subtitle formats and not for others. However I can see the argument for why this might not matter and we should just filter or not-filter for any format.

killergerbah:
OK I looked into this briefly and all subtitle formats have at least partial support for HTML except for ASS and BluRay which is image-based. But it's very unlikely that anyone will care if XML starts disappearing from their ASS subtitles. So will just go ahead and merge this change given the value it provides. Thank you!

kowasaur:
Thanks for taking the time to look into this!

PR #723, opened 2025-05-26
Paint subtitle stroke behind text fill

RicBent:
Currently subtitle outlines cause the text itself to be overdrawn. This limits the outline thickness one can select before the text becomes unreadable. It is possible to increase the font thickness to remedy this issue a bit, but the result can look odd and requires weird balancing between the two options.

Screenshot 2025-05-26 at 19 51 48

This PR forces the outline to be drawn behind the text itself, making all text outline and text thickness combinations work well without any issues.

Screenshot 2025-05-26 at 19 51 30

The `paint-order` css property went baseline last year so using it should be fine.


killergerbah:
Great improvement! Thanks!

PR #729, opened 2025-06-06
Improved Naming scheme for video in Emby/Jellyfin

bpwhelan:
Unify and improve the the naming scheme for the "Video Name" in both Emby and Jellyfin. This will result in more consistent naming between the two and provide a lot more useful information for jellyfin folk.

Before and After JF:
chrome_Otxp2BrRrs

Before and After Emby:
chrome_X5RX2HU1Xm

### DIFF
--- a/extension/src/entrypoints/emby-jellyfin-page.ts
+++ b/extension/src/entrypoints/emby-jellyfin-page.ts
@@ -41,7 +41,7 @@ export default defineUnlistedScript(() => {
             const mediaID = session.PlayState.MediaSourceId;
             const nowPlayingItem = session.NowPlayingItem;
             const path = nowPlayingItem.Path;
-            response.basename = path ? path.split(/[\\/]/).pop() : (nowPlayingItem.FileName ?? nowPlayingItem.Name);
+            response.basename = nowPlayingItem.FileName ?? (path ? path.split(/[\\/]/).pop() : nowPlayingItem.Name);
 
             const subtitles: VideoDataSubtitleTrack[] = [];
             nowPlayingItem.MediaStreams.filter(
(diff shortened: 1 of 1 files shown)


bpwhelan:
I could see potentially removing the item id for cleanliness, but it could be potentially useful information for scripting, like if you wanted to find that episode in the future programmatically you can parse that and construct a url to take you straight there.

killergerbah:
@bpwhelan Can you add reasonable fallback behavior in case some of the `NowPlayingItem` fields are not defined? It looks like `FileName` is the only field guaranteed to be defined.

image

bpwhelan:
@killergerbah I think `FileName` is not available for Jellyfin IIRC, but I believe `Path` is valid for both (当たり前でしょ), I will look at getting the filename from that field to bring it more in line with the asbplayer app naming scheme and scrap the `NowPlayingItem` fields

bpwhelan:
Changed to use Path, and fall back to previous behavior if Path doesn't exist (should never happen, just safety) 

Emby behavior unchanged from main branch.

## Emby
image
## JF
image

killergerbah:
@bpwhelan Thanks! This looks good. Final suggestion is to use your original naming format if it's possible. P2 would be to use the parts that are available in `NowPlayingItem` if only some parts are missing.

killergerbah:
Merging this but feel free to implement your original naming scheme if you'd like. Thanks for the contribution!

bpwhelan:
Sorry meant to reply, if I were to do anything it would have to be all or nothing (either all of the nowplayingitem fields or just use filename) 

But I think filename makes the most sense as it's consistent with how the app works.

Thanks!

Issue #237, opened 2023-05-03
Allow selection of useful regex filter presets


killergerbah:

killergerbah:
Design:
Add menu above  "subtitle regex filter" setting called "subtitle regex preset". The first option in the menu should be "No preset." The rest should include the options in the issue description: "Sound effects," "furigana" etc.
If a user selects one of the  preset options, it should auto-populate the regex filter setting with the corresponding regex. If the user starts typing in the regex filter text box, the preset menu should automatically select "No preset."

5byuri:
Will work on this!

5byuri:
@killergerbah 

<img width="1134" alt="Image" src="https://github.com/user-attachments/assets/7b7fdc43-4a34-4302-97ac-0aed7bb370c7" />

wouldn't a combobox with autocomplete be a better design choice?
firstly it would remove an additional textfield, if users write a regex it already gets suggestions for other regex and can just continue type their own regex if they want a custom regex

killergerbah:
@5byuri Sorry I completely forgot to reply to this. One benefit of the preset feature would be to help users who don't know how to write regexes in the first place. These users would have no way of knowing what to type into the auto-complete box to get what they need. A simple menu would work around that.

I could see the autocomplete box working if it could be opened up and display all the "natural language representations" (no parentheses, music symbols, etc) to the user, instead of the plain regex representations.

PR #731, opened 2025-06-12
Add regex presets menu

SayadSindid:
In regard to #237

Localization should probably be added for each preset name don't know how to do.

Support could be added for the (.*)\n+(?!-)(.*) indicated in the guide in order to change the "regex filter text replacement field" when this regex is selected.

Additional features:
Maybe support for multiples custom preset made by the user?

**Note:** First pull request
If anything is wrong feel free to explain

### DIFF
--- a/common/components/SettingsForm.tsx
+++ b/common/components/SettingsForm.tsx
@@ -79,7 +79,6 @@ import AnkiConnectTutorialBubble from './AnkiConnectTutorialBubble';
 import DeckFieldTutorialBubble from './DeckFieldTutorialBubble';
 import NoteTypeTutorialBubble from './NoteTypeTutorialBubble';
 
-
 const defaultDeckName = 'Sentences';
 
 const defaultNoteType = {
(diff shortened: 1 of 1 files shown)


killergerbah (changes requested):
Thanks for contributing!
This is close, but a new setting is not necessary to store the selected preset. Please use `useState` instead.

common/components/SettingsForm.tsx
```diff
@@ -79,6 +79,14 @@ import AnkiConnectTutorialBubble from './AnkiConnectTutorialBubble';
 import DeckFieldTutorialBubble from './DeckFieldTutorialBubble';
 import NoteTypeTutorialBubble from './NoteTypeTutorialBubble';
 
+const regexPresetsList = [
+    { name: 'No preset', regex: '' },
```
killergerbah: These strings need to be localizable.

SayadSindid:
Added the possibility to localize and changed the logic to use state instead of having a new setting.

common/components/SettingsForm.tsx
```diff
@@ -2635,14 +2647,36 @@ export default function SettingsForm({
...
                                 fullWidth
                                 value={subtitleRegexFilter}
                                 color="primary"
                                 error={!validRegex}
                                 helperText={validRegex ? undefined : 'Invalid regular expression'}
-                                onChange={(event) => handleSettingChanged('subtitleRegexFilter', event.target.value)}
+                                onChange={(event) => {
+                                    handleSettingChanged('subtitleRegexFilter', event.target.value);
+                                    if (selectedRegexPresetName !== regexPresetsList[0].name) {
+                                        setSelectedRegexPresetName(regexPresetsList[0].name);
```
killergerbah, line 2677: It's risky to update state from inside the render function. You should consider using `useEffect` instead.
Also, ideally if the user types in one of the presets manually the correct one gets selected. This code is selecting the first preset no matter what.

PR #733, opened 2025-06-14
Resolve subtitles through base_url

kekulta:
This PR is inspired by https://github.com/killergerbah/asbplayer/pull/609.
The original PR appears outdated and unlikely to be completed, so here's a new implementation.

With huge help from the YouTube.js maintainer, I figured out how to retrieve poToken and resolve subtitles directly.
I hardcoded `&tlang=en` rn just to check that auto-translation works. If this approach is ok, I will clean it up and make second PR with support to auto-translated subs.

This is a draft (very messy!). If we agree this approach works, I'll clean it up.

I'm not a web developer, so this is essentially my first interaction with TypeScript.

I hope I didn't reinvent the wheel here :smile:

image

### DIFF
--- a/common/base64/index.ts
+++ b/common/base64/index.ts
@@ -1,4 +1,27 @@
-export function bufferToBase64(buffer: ArrayBufferLike): string {
+export function encodeBase64(str: string) {
+  // Browser environment
+  if (typeof TextEncoder !== 'undefined') {
+    const encoder = new TextEncoder();
+    const bytes = encoder.encode(str);
+    const binString = String.fromCharCode(...bytes);
+    return btoa(binString);
+  }
+  
+  // Fallback for legacy browsers
+  return btoa(unescape(encodeURIComponent(str)));
+}
+
+export function base64ToBuffer(base64: string): Uint8Array {
+  const str = atob(base64);
+  const len = str.length;
+  const bytes = new Uint8Array(len);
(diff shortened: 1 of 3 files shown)


killergerbah:
@kekulta Thank you, this is incredible work. If I'm understanding the change correctly, the POT is cached in session storage and can be decoded. I think this approach will work great.

I only have a few things to add before you clean things up:

- As you already commented in your change, some parts of the POT implementation like the session storage keys might be unstable. If possible it would be good to keep your integration as resistant to change as possible.
- If for some reason the POT extraction fails, it would good if we could still fallback to the existing InnerTube solution.

Some other minor comments which I think you're already aware of:

- base64 functions can be reused, instead of rewriting them in a few places.
- The SRT conversion can be done using the `SrtParser` already being used with the InnerTube transcription.
- There's no need to try to support legacy browsers - asbplayer will never run in a browser that isn't modern-ish.

killergerbah:
One other thing - the original idea to support YT's machine-translated tracks will need the user to specify their target language somewhere. Naturally, I think this would be another setting. But let me know if you had other ideas.

killergerbah:
Thanks again for this contribution. I used your PR as a starting point to integrate POT decoding as a fallback method for fetching YT subs. In particular, the POT seems to provide the only reliable way to fetch subs for members-only videos.

PR #743, opened 2025-07-09
Add seek duration

artjomsR:
Implements https://github.com/killergerbah/asbplayer/issues/287

I've tested firefox and chrome extensions and the local player. I didn't change the android version

Also, when I was generating translations, I couldn't run `node ./scripts/loc/merge-loc-from-en.js` as I was getting 
```
node:internal/modules/package_json_reader:256
throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
``` so I had to do `yarn add flat`, not sure why

### CODE REVIEW

File: common/locales/de.json [hunk 1]
```diff
@@ -314,6 +314,7 @@
         "autoPauseAtSubtitleStart": "Am Anfang des Untertitels",
         "autoPausePreference": "Auto-Pause",
         "autoPausePreferenceHelperText": "Aktiviert keine Auto-Pause",
+        "seekDuration": "Amount of seconds to skip fowards / backwards when seeking with respective shortcuts",
```
killergerbah: This label is pretty long - can we shorten it?

"Seconds to skip forwards/backwards when seeking with respective shortcuts"

killergerbah: That's the shortest I could think of for now

killergerbah (changes requested):

common/locales/de.json
artjomsR: `Seconds to skip when seeking forward/backward` or
`Seek duration (sec)` ?

Now what you say it, I'm thinking it'd be better to move these shortcut related settings from `MISC` into `KEYBOARD SHORTCUTS`. This way they're gonna be more discoverable by the users and won't need as much helper text because both shortcut and the setting will support each other with context. I'm not sure if there would be any technical challenges with migrating the settings like this?

killergerbah: Yeah I think that would be the ideal solution. Another input field directly under the keyboard shortcuts for seeking left/right.
Similarly, it would make sense for "Always play after invoking 'Seek to beginning of current/previous subtitle'" setting to go underneath the "seek to beginning of current/previous subtitle" shortcut setting.

Ideally there is a solution for both binds. The problem is that we are rendering the keyboard shortcut settings in a for loop over `keyBindProperties`. First idea that comes to mind is to add a JSX component field (`additionalControl` or similar) to `KeyBindProperties` that defaults to `null`. Then we can just modify `keyBindProperties` for the two keyboard shortcuts we actually care about, without having to modify the for loop itself with special case logic.

artjomsR: Ok I've pushed the latest commit https://github.com/killergerbah/asbplayer/pull/743/commits/feaafc7cb8057cbf83c57ab95a0142cf90368871 adding this. I've had to move `const {...} = settings` init because `keyBindProperties` now depends on `settings.seekDuration`

Let me know if you're happy with this approach and I'll tidy up the rest of the PR with the other settings

image

killergerbah: @artjomsR Sweet, yeah the code looks good. UI-wise I think I would want the new text field have the same width and be aligned with the keyboard binds.

artjomsR:
All done now :) I did a few force pushes to amend my work in progress but didn't change any old commits that you've already seen.

- I did some minor corrections to strings for consistency etc
- I've changed the UI as you've requested although I thought it was nice that these settings stood out since they're not actually keybinds in the Keybinds section

image
image
image
image

killergerbah:
@artjomsR Thanks! Really great work and I appreciate you entertaining my suggestions.

Yes I agree that with the alignment being all the same it is difficult to tell the keybinds apart from the additional settings. Maybe there is a solution to this while keeping the alignment - in my mind the alignment needs to be the same.

common/locales/zh_CN.json
```diff
@@ -313,9 +313,9 @@
         "autoPauseAtSubtitleEnd": "在字幕末尾",
         "autoPauseAtSubtitleStart": "在字幕开头",
         "autoPausePreference": "自动暂停首选项",
-        "autoPausePreferenceHelperText": "不启用自动暂停。设置启用自动暂停时暂停的首选项。",
```
killergerbah, line 316: Unfortunately we need to keep this key because currently all versions of the extension share the same loc file.

killergerbah (changes requested):

killergerbah:
Hey just one more change and I'll merge ^^
Thanks again so much!

artjomsR:
> Yes I agree that with the alignment being all the same it is difficult to tell the keybinds apart from the additional settings. Maybe there is a solution to this while keeping the alignment - in my mind the alignment needs to be the same.

It'll be easy to change later since it's contained in KeyBindRelatedSetting. And most of these settings stand out anyway since they have a different input type

> Unfortunately we need to keep this key because currently all versions of the extension share the same loc file.

Just revert the strings in the .JSON files, right? (Even though they won't be used in code). I'm away from my pc today and will try to do it tomorrow, but feel free to revert then if you wish to merge it quicker :)

artjomsR:
Ok done

killergerbah:
Sweeeet!

PR #748, opened 2025-07-26
Replace deprecated InputProps

artjomsR:
I've noticed the deprecation warnings while working on the previous PR, so figured I'd make a separate one for this. Bar 1 exception, the replacements are:

- InputProps -> slotProps.input
- inputProps -> slotProps.htmlInput

I've given it a quick test, and it seemed fine, feel free to do a more thorough check. And feel free to use this PR partially if you feel that'll be more useful :)

### DIFF
--- a/common/components/SettingsForm.tsx
+++ b/common/components/SettingsForm.tsx
@@ -296,29 +296,31 @@ const SelectableSetting = React.forwardRef<HTMLDivElement, SelectableSettingProp
                 error={error}
                 helperText={error ? t('settings.missingFieldError', { field: value }) : ''}
                 color="primary"
-                InputProps={{
-                    endAdornment: (
-                        <InputAdornment style={{ marginRight: -12 }} position="end">
-                            {(removable || onOrderChange) && (
+                slotProps={{
+                    input: {
+                        endAdornment: (
+                            <InputAdornment style={{ marginRight: -12 }} position="end">
+                                {(removable || onOrderChange) && (
+                                    <IconButton
+                                        onClick={(e) => {
+                                            e.stopPropagation();
+                                            setOptionsMenuAnchorEl(e.currentTarget);
+                                            setOptionsMenuOpen(true);
+                                        }}
+                                    >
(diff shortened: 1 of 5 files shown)


common/components/SettingsForm.tsx
```diff
@@ -296,29 +296,31 @@ const SelectableSetting = React.forwardRef<HTMLDivElement, SelectableSettingProp
                 error={error}
                 helperText={error ? t('settings.missingFieldError', { field: value }) : ''}
                 color="primary"
-                InputProps={{
-                    endAdornment: (
-                        <InputAdornment style={{ marginRight: -12 }} position="end">
-                            {(removable || onOrderChange) && (
+                slotProps={{
+                    htmlInput: {
+                        endAdornment: (
```
killergerbah, line 301: On my computer the TextField endAdornments only show up if they're inside `input: {...}`
Might be because I just updated MUI. Can you test after pulling from main?

artjomsR: Ah yep I see what you mean, I've missed that `ms` dissappeared for some fields facepalm 🤦 I can replace all the affected `htmlInput` with `input` to fix endAdornments, if that works for you?

killergerbah: For me it's not just `ms` but for example all the adornments that should be in the Anki settings:

image



This is what it should look like:
image

killergerbah: But yeah, if switching to `input` fixes that then please do :pray:

artjomsR:
Ok, I gave it another shot, I didn't create a new commit and amended it instead, since I changed the existing code a bit. All the items on the right-hand side (`ms`, refresh icons) were present for me on the settings and Anki export dialogue

The general approach seems to be something like
```
                                slotProps={{
                                    htmlInput: {
                                        min: 0,
                                        step: 1,
                                        ...,
                                    },
                                    input: {
                                        endAdornment: ...,
                                        ...,
                                    },
                                }}
```

killergerbah:
Awesome man, thanks as always

Issue #751, opened 2025-07-28
Add keyboard shortcuts to update subtitle position offset


mseh1128:
**Is your feature request related to a problem? Please describe.**
I think a fairly common use case is updating the subtitle position offset. If you're using multiple streaming sites or playing a video with embedded subtitles, you end up adjusting the position offset fairly frequently. It would be nice if you could do that with keyboard shortcuts as well. 

**Describe the solution you'd like**
Add keyboard shortcuts to increase or decrease the bottom (or top) subtitle position offsets by a set amount.

**Describe alternatives you've considered**
Alternative is updating them in settings

**Additional context**
I actually already implemented this in my local asbplayer repo. If other people find this useful, I think it could a be a "nice to have" feature.

<img width="773" height="284" alt="Image" src="https://github.com/user-attachments/assets/e7fd97db-8cd2-45cf-a687-4e0e4494dc26" />

PR #759, opened 2025-08-08
Add subtitle position offset keyboard shortcuts

mseh1128:
Implementation for #751 

Pretty standard keyboard shortcut implementation. The only thing that was kind of tricky was updating the settings (to sync w/ the popup window) for the extension in key-bindings.ts. I'm not too familiar with the codebase so hopefully I didn't miss anything 🤞.


common/locales/de.json
```diff
@@ -138,7 +138,11 @@
         "updateLastCard": "Zuletzt erstellte Anki-Karte mit aufgezeichnetem Untertitel",
         "exportCard": "Export card to Anki, bypassing dialog.",
         "toggleSidePanel": "Toggle side panel",
-        "toggleRepeat": "Toggle repeat mode"
+        "toggleRepeat": "Toggle repeat mode",
+        "increaseSubtitlePositionOffset": "Adjust subtitle position offset from bottom by +20px",
```
killergerbah: Can you change this to "Move bottom subtitles down," "Move bottom subtitles up," "Move top subtitles down," etc?
It doesn't match the setting exactly but it's easier to read and also future-proofs us against changing the 20px value.

extension/src/services/key-bindings.ts
```diff
@@ -262,6 +264,37 @@ export default class KeyBindings {
...
+        this._unbindAdjustSubtitlePositionOffset = this._keyBinder.bindAdjustSubtitlePositionOffset(
+            (event, increase) => {
+                event.preventDefault();
+                event.stopImmediatePropagation();
+
+                const currentOffset = context.subtitleController.bottomSubtitlePositionOffset;
+                const newOffset = currentOffset + (increase ? 20 : -20);
+
+                context.subtitleController.bottomSubtitlePositionOffset = newOffset;
+                context.settings.set({ subtitlePositionOffset: newOffset });
```
killergerbah: Edited this comment since I realized that you should publish this settings update to the rest of the app.

```
context.settings.set(...)
    .then(() => {
               const settingsUpdatedCommand: VideoToExtensionCommand<SettingsUpdatedMessage> = {
                    sender: 'asbplayer-video',
                    message: {
                        command: 'settings-updated',
                    },
                    src: context.video.src,
                };
                browser.runtime.sendMessage(settingsUpdatedCommand);
    })
    .catch(console.error)
 ```
 
 Then, you can remove this line:
 
 ```
                 context.subtitleController.bottomSubtitlePositionOffset = newOffset;
```

since the `subtitleController` will receive the settings update anyway.


Same comment for the other bind.

common/settings/settings-provider.ts
```diff
@@ -110,6 +110,10 @@ export const defaultSettings: AsbplayerSettings = {
         increasePlaybackRate: { keys: isMacOs ? '⇧+⌃+]' : 'ctrl+shift+]' },
         toggleSidePanel: { keys: '`' },
         toggleRepeat: { keys: isMacOs ? '⇧+R' : 'shift+R' },
+        increaseSubtitlePositionOffset: { keys: isMacOs ? '⇧+V' : 'shift+V' },
```
killergerbah: I feel like this is a pretty niche use-case so my initial preference is to make these shortcuts unbound by default (empty string). People who want to use the shortcut will find it. However, I'm open to changing my mind if you feel strongly about this.

extension/src/services/key-bindings.ts
mseh1128: Sounds good, will do. I was debating whether to do that, but wasn't sure. The commands/messages are still a little confusing to me. I'll probably ask you about them on Discord.

common/settings/settings-provider.ts
mseh1128: I agree. If someone was trying to use this, they'd likely look for it in "Keyboard shortcuts" anyway. I was also struggling to find suitable keybinds for this.

common/settings/settings-provider.ts
```diff
@@ -110,6 +110,10 @@ export const defaultSettings: AsbplayerSettings = {
         increasePlaybackRate: { keys: isMacOs ? '⇧+⌃+]' : 'ctrl+shift+]' },
         toggleSidePanel: { keys: '`' },
         toggleRepeat: { keys: isMacOs ? '⇧+R' : 'shift+R' },
+        moveBottomSubtitlesUp: { keys: isMacOs ? '' : '' },
```
killergerbah: Can just be `''`, no need for ternary expression

killergerbah (changes requested):
Thanks for your patience - was away for from my computer for a bit. Just one last nitpick.

common/settings/settings-provider.ts
mseh1128: Fixed. Thanks for the review.

killergerbah:
Amazing stuff, thank you.

Issue #716, opened 2025-05-15
Stremio Subtitle Detection


SpazzTL:
**Before submitting an issue check out the common issues section of the README:**
https://github.com/killergerbah/asbplayer?tab=readme-ov-file#common-issues

**What browser?**
Chrome & Firefox<br>
**Is it happening with local video or streaming video?**
Streaming Video via (https://web.stremio.com/)<br>
**If the issue is happening with streaming video, what version of the extension are you using?**
1.10.0<br>
**If the issue is happening with streaming video, on what website(s) is the issue happening?**
(https://web.stremio.com/)<br>
**Please provide logs from your developer console. At the very least, copy and paste the error from the "Copy Last Error" button (top-right) if the error is occuring in the asbplayer webapp.**
No errors logged.
<br>
**Notes**:
Subtitle detection may not be possible with this particular source. Although if it could it would be quite convenient, allowing me to not download subtitles & video before mining.

![Image](https://github.com/user-attachments/assets/3d43ff09-e7bc-475f-88fc-fb2b02e2654f)

![Image](https://github.com/user-attachments/assets/f97c6454-6452-407f-a94f-f87888af27fe)

SpazzTL:
May help with implementation:
``<div style="position: absolute; right: 0px; bottom: 5%; left: 0px; z-index: 1; text-align: center; opacity: 1;"><div style="display: inline-block; padding: 0.2em; white-space: pre-wrap; font-size: 4vmin; color: rgb(255, 255, 255); background-color: rgba(0, 0, 0, 0); text-shadow: rgb(0, 0, 0) 1px 1px 0.1em;">終わってしまったな</div><br></div>``
<img width="956" height="114" alt="Image" src="https://github.com/user-attachments/assets/d7bd0ab9-e200-4693-822a-1609b9402e9c" />


Further more; I found subtitles are loaded in only after enabled with a request such as this:
``http://127.0.0.1:11470/subtitles.vtt?from=SUBTITLESOURCE``
The results looks like This


````1
00:01:59.912 --> 00:02:01.080
ãƒ•ãƒªãƒ¼ãƒ¬ãƒ³

2
00:02:08.254 --> 00:02:10.464
çŽ‹éƒ½ãŒè¦‹ãˆã¦ããŸã­

3
00:02:10.548 --> 00:02:13.384
ç§ãŸã¡å‹‡è€…ä¸€è¡Œã®
[...]

PR #763, opened 2025-08-16
Stremio Support {Rough} 

SpazzTL:
A rough, initial start for Stremio Support. 

Subtitles MUST be enabled first on the site, before ASB will recognize them. (This is unavoidable AFAIK)
Subtitles are named ``Subtitle 1 -> Subtitle 2`` and so on, as Stremio subs have same/no names... 

Known Errors:
- Certain Subtitle Sources Fail 
- Subtitles duplicate sometimes

Solves issue #716

### DIFF
--- a/extension/src/entrypoints/stremio-page.ts
+++ b/extension/src/entrypoints/stremio-page.ts
@@ -1,0 +1,72 @@
+import {
+    VideoData,
+    VideoDataSubtitleTrack
+} from '@project/common';
+import {
+    trackFromDef
+} from '@/pages/util';
+
+export default defineUnlistedScript(() => {
+    const discoveredSubtitles = new Map < string, VideoDataSubtitleTrack > ();
+    let trackIndex = 1;
+
+    const dispatchSubtitles = () => {
+        const response = {
+            error: '',
+            basename: document.title,
+            subtitles: Array.from(discoveredSubtitles.values()),
+        };
+
+        document.dispatchEvent(
+            new CustomEvent('asbplayer-synced-data', {
+                detail: response,
+            })
+        );
+    };
+
+    const originalFetch = window.fetch;
+    window.fetch = async (...args) => {
+        const url = typeof args[0] === 'string' ?
+            args[0] :
+            args[0] instanceof Request ?
+            args[0].url :
+            null;
+
+        if (url) {
+            const stremioSubtitleMatch = url.match(/https:\/\/subs\d+\.strem\.io\/.+\/file\/\d+/);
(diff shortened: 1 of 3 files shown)


extension/src/entrypoints/stremio-page.ts
```diff
@@ -0,0 +1,72 @@
...
+                extension = 'srt';
+            } else if (opensubtitlesMatch) {
+                extension = 'vtt';
+            }
+            
+            if (extension) {
+                if (!discoveredSubtitles.has(url)) {
+                    const track = trackFromDef({
+                        label: `Stremio ${trackIndex++}`,
+                        language: '',
```
killergerbah, line 51: There's no way to extract some kind of language information? An empty string will probably break the "Remember these tracks on this site" feature.

SpazzTL: I will try and tinker tomorrow if I have time, but I can't think of any non-scuffed methods at the moment.  

Currently thinking of finding out what buttons have been clicked/interacted with or just checking the subtitle text lol.

killergerbah: I see, then maybe we can just disable the "Remember these tracks" feature for this site. I can work on that.

killergerbah: Another idea: are the tracks usually in a specific order? Like English first then original?

SpazzTL: > Another idea: are the tracks usually in a specific order? Like English first then original?

Not really, subtitles can also change (be removed / added). I think its best to disable the "Remember these tracks".
Screenshots below if they help, showing the web.stremio subtitle options, and the asb player loaded subs.
image
image
image
image
> The subtitles are ordered/numbered here in order loaded... not ideal, but for my use case, works very well.... `

killergerbah:
Thanks @SpazzTL 
To be honest I haven't tested this but I trust you. I'll share this on the Discord server before it's released as well.
BTW I do see language info in the screenshots you posted. Have you already tried intercepting HTTP response payloads for this info? There are examples of other `page-*` scripts doing this.

SpazzTL:
> Thanks @SpazzTL To be honest I haven't tested this but I trust you. I'll share this on the Discord server before it's released as well. BTW I do see language info in the screenshots you posted. Have you already tried intercepting HTTP response payloads for this info? There are examples of other `page-*` scripts doing this.

I'll start refining this hopefully shortly. I haven't tried intercepting HTTP response payloads yet, I think?. I'll look into other scripts and see if I can come up with a more stable experience. I don't think remembering tracks should be enabled for this script, just due to how stremio handles subtitles. I haven't used this or asbplayer in general much recently, but I plan to work on this script while watching some media with it soon.

PR #770, opened 2025-09-05
Implement bulk export

agloo:
New functionality:
This adds a button to the side panel overlay between `Download Subtitles as SRT` and `Mining History` that starts exporting one card after another from the current subtitle to the end of the video. While it does this, it blocks out the side panel with a cancellable progress bar to avoid weird state if you try to seek around/kick off an export.

Architecture:
This driven by a queue, with events being consumed by logic in SidePanel. SidePanel will kick off the export of the next card every time it gets a card-exported message back.
I found that this put the extension into some weird states, most notably that I was getting a lot of pausing-stopped-video and empty-card-created errors. I downgraded those from exceptions to a no-op and empty cards, respectively.

I tested this almost exclusively on one YouTube video--happy to look at whatever else you'd like to see covered.

### DIFF
--- a/common/anki/anki.ts
+++ b/common/anki/anki.ts
@@ -106,7 +106,7 @@ export interface ExportParams {
     ankiConnectUrl?: string;
 }
 
-export async function exportCard(card: CardModel, ankiSettings: AnkiSettings, exportMode: AnkiExportMode) {
+export async function exportCard(card: CardModel, ankiSettings: AnkiSettings, exportMode: AnkiExportMode = 'default'): Promise<string> {
     const anki = new Anki(ankiSettings);
     const source = sourceString(card.subtitleFileName, card.mediaTimestamp);
     let audioClip =
@@ -375,6 +375,18 @@ export class Anki {
                 }
 
                 throw new Error('Could not update last card because the card info could not be fetched');
+            case 'bulk':
+                const res = await this._executeAction('addNote', params, ankiConnectUrl).catch((error) => {
+                    if (error instanceof Error && error.message.includes('duplicate')) {
+                        return { result: undefined, duplicate: true } as any;
+                    }
+                    throw error;
+                });
+                if ((res as any)?.duplicate) {
+                    // Return an empty string to indicate a duplicate
+                    return '' as any;
+                }
+                return (res as any).result;
             case 'default':
                 return (await this._executeAction('addNote', params, ankiConnectUrl)).result;
             default:
--- a/common/app/components/BulkExportModal.tsx
+++ b/common/app/components/BulkExportModal.tsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import Button from '@mui/material/Button';
+import Typography from '@mui/material/Typography';
+import Box from '@mui/material/Box';
+import LinearProgress from '@mui/material/LinearProgress';
+import CircularProgress from '@mui/material/CircularProgress';
+import Dialog from '@mui/material/Dialog';
+import DialogTitle from '@mui/material/DialogTitle';
+import DialogContent from '@mui/material/DialogContent';
+import DialogActions from '@mui/material/DialogActions';
+
+interface BulkExportModalProps {
+    open: boolean;
+    currentIndex: number; // completed count
+    totalItems: number;   // total count
+    onCancel: () => void;
+}
+
+export default function BulkExportModal({ open, currentIndex, totalItems, onCancel }: BulkExportModalProps) {
+    const { t } = useTranslation();
+    const progress = totalItems > 0 ? (currentIndex / totalItems) * 100 : 0;
+    return (
+        <Dialog
+            open={open}
+            onClose={(event, reason) => {
+                if (reason === 'backdropClick' || reason === 'escapeKeyDown') return;
+                onCancel();
+            }}
+            fullWidth
+            maxWidth="sm"
+            aria-labelledby="bulk-export-title"
+            aria-describedby="bulk-export-description"
+            disableEscapeKeyDown
+            sx={{
+                '& .MuiDialog-paper': {
+                    bgcolor: 'background.default',
+                },
+            }}
+            BackdropProps={{ sx: { backgroundColor: 'rgba(0, 0, 0, 1)' } }}
+        >
+            <DialogTitle id="bulk-export-title">
+                {t('info.bulkExport.title', 'Bulk Export in Progress')}
+            </DialogTitle>
+            <DialogContent dividers>
+                <Box display="flex" alignItems="center" justifyContent="center" mb={3}>
+                    <CircularProgress size={64} color="primary" />
+                </Box>
+                <Typography id="bulk-export-description" variant="body1" align="center" gutterBottom>
+                    {t('info.bulkExport.description', 'Exporting subtitles to Anki. Please wait...')}
+                </Typography>
+                <Box width="100%" mt={3}>
+                    <Typography variant="h6" align="center" gutterBottom>
+                        {t('info.bulkExport.progressCompleted', 'Exported {{current}} of {{total}}', { current: currentIndex, total: totalItems })}
+                    </Typography>
+                    <LinearProgress
+                        variant="determinate"
+                        value={progress}
+                        sx={{ height: 10, borderRadius: 5 }}
+                    />
+                </Box>
+            </DialogContent>
+            <DialogActions sx={{ justifyContent: 'center' }}>
+                <Button onClick={onCancel} variant="outlined" color="primary" data-testid="bulk-export-cancel-button">
+                    {t('action.cancel', 'Cancel')}
+                </Button>
+            </DialogActions>
+        </Dialog>
+    );
+}
--- a/common/app/components/Player.tsx
+++ b/common/app/components/Player.tsx
@@ -1075,6 +1075,7 @@ const Player = React.memo(function Player({
                         settings={settings}
                         keyBinder={keyBinder}
                         webSocketClient={webSocketClient}
+                        exporting={miningContext.bulkExporting}
                     />
                 </Grid>
             </Grid>
--- a/common/app/services/mining-context.ts
+++ b/common/app/services/mining-context.ts
@@ -1,15 +1,70 @@
-type MiningEvent = 'started-mining' | 'stopped-mining';
+type MiningEvent =
+    | 'started-mining'
+    | 'stopped-mining'
+    | 'bulk-export-started'
+    | 'bulk-export-completed'
+    | 'bulk-export-item-completed'
+    | 'bulk-export-cancelled';
 type Callback = () => void;
+
+export interface BulkExportItem {
+    subtitle: import('../../src/model').SubtitleModel;
+    index: number;
+}
 
 export class MiningContext {
     private readonly _callbacks: { [event in MiningEvent]: Callback[] } = {
         'started-mining': [],
         'stopped-mining': [],
+        'bulk-export-started': [],
+        'bulk-export-completed': [],
+        'bulk-export-item-completed': [],
+        'bulk-export-cancelled': [],
     };
     private _mining = false;
+    private _bulkExporting = false;
+    private _bulkExportCancelled = false;
+    private _bulkExportQueue: BulkExportItem[] = [];
+    private _currentBulkExportIndex = 0;
 
     get mining() {
         return this._mining;
+    }
+
+    get bulkExporting() {
+        return this._bulkExporting;
+    }
+
+    get bulkExportCancelled() {
+        return this._bulkExportCancelled;
+    }
+
+    get bulkExportQueue() {
+        return this._bulkExportQueue;
+    }
+
+    get currentBulkExportIndex() {
+        return this._currentBulkExportIndex;
+    }
+
+    isBulkExportCancelled() {
+        return this._bulkExportCancelled;
+    }
+
+    get bulkExportProgress() {
+        if (!this._bulkExporting || this._bulkExportQueue.length === 0) {
+            return { current: 0, total: 0, remaining: 0 };
+        }
+        
+        const totalRemaining = this._bulkExportQueue.length;
+        const currentProgress = this._currentBulkExportIndex;
+        const remaining = Math.max(0, totalRemaining - currentProgress);
+        
+        return {
+            current: currentProgress,
+            total: totalRemaining,
+            remaining: remaining
+        };
     }
 
     onEvent(event: MiningEvent, callback: Callback) {
@@ -26,6 +81,52 @@ export class MiningContext {
         this._mining = false;
         this._callbacks['stopped-mining'].forEach((c) => c());
     }
+
+    startBulkExport(items: BulkExportItem[], startIndex: number = 0) {
+        this._bulkExporting = true;
+        this._bulkExportCancelled = false;
+        this._bulkExportQueue = startIndex > 0 ? items.slice(startIndex) : items;
+        this._currentBulkExportIndex = 0;
+        
+        this._callbacks['bulk-export-started'].forEach((c) => c());
+    }
+
+    completeBulkExportItem() {
+        this._currentBulkExportIndex++;
+        
+        this._callbacks['bulk-export-item-completed'].forEach((c) => c());
+        
+        if (this._currentBulkExportIndex >= this._bulkExportQueue.length) {
+            this.completeBulkExport();
+        }
+    }
+
+    completeBulkExport() {
+        this._bulkExporting = false;
+        this._bulkExportCancelled = false;
+        this._bulkExportQueue = [];
+        this._currentBulkExportIndex = 0;
+        this._callbacks['bulk-export-completed'].forEach((c) => c());
+    }
+
+    cancelBulkExport() {
+        if (this._bulkExportCancelled) {
+            return;
+        }
+        
+        this._bulkExportCancelled = true;
+        this._bulkExporting = false;
+        this._bulkExportQueue = [];
+        this._currentBulkExportIndex = 0;
+        
+        this._callbacks['bulk-export-cancelled'].forEach((c) => c());
+        
+        // Force stop any ongoing mining operations
+        this._mining = false;
+        this._callbacks['stopped-mining'].forEach((c) => c());
+    }
+
+    // Removed unused processBulkExportQueue to reduce API surface
 
     private _remove(callback: Function, callbacks: Callback[]) {
         for (let i = callbacks.length - 1; i >= 0; --i) {
--- a/common/src/message.ts
+++ b/common/src/message.ts
@@ -112,6 +112,7 @@ export interface RecordMediaAndForwardSubtitleMessage extends Message, CardTextF
     readonly imageDelay: number;
     readonly playbackRate: number;
     readonly mediaTimestamp: number;
+    readonly exportMode?: AnkiExportMode;
 }
 
 export interface StartRecordingMediaMessage extends Message, ImageCaptureParams {
@@ -160,6 +161,7 @@ export interface CopySubtitleMessage extends Message, CardTextFieldValues {
     readonly postMineAction: PostMineAction;
     readonly subtitle?: SubtitleModel;
     readonly surroundingSubtitles?: SubtitleModel[];
+    readonly exportMode?: AnkiExportMode;
 }
 
 export interface CopySubtitleWithAdditionalFieldsMessage extends Message, CardTextFieldValues {
@@ -190,6 +192,9 @@ export interface CardUpdatedMessage extends Message, CardModel {
 export interface CardExportedMessage extends Message, CardModel {
     readonly command: 'card-exported';
     readonly cardName: string;
+    readonly isBulkExport?: boolean;
+    readonly skippedDuplicate?: boolean;
+    readonly exportError?: string;
 }
 
 export interface CardSavedMessage extends Message, CardModel {
@@ -367,6 +372,10 @@ export interface SubtitlesToVideoMessage extends Message {
 
 export interface RequestSubtitlesMessage extends Message {
     readonly command: 'request-subtitles';
+}
+
+export interface RequestCurrentSubtitleMessage extends Message {
+    readonly command: 'request-current-subtitle';
 }
 
 export interface RequestSubtitlesFromAppMessage extends MessageWithId {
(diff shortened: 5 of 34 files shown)


killergerbah:
@agloo This is amazing work, I'm really impressed by how deeply you were able to understand the codebase to implement a big feature like this.

I have one overall comment that would require (hopefully) a small refactor: it looks like this feature depends on the side panel being open. As a user I think I would expect the bulk export to continue regardless of whether the side panel is open or not.

My suggestion is:
- Move the bulk export logic from `SidePanel` into a "controller" class like `bulk-export-controller.ts` under `extension/src/controllers`. It should be easy to start a bulk export from an event inside `binding.ts`. You'll have easy access to a lot of state which might simplify the logic.
- If you'd like to keep the nice side panel UI and progress bar, you can publish progress via a message.
- There should be some indication on the video element that the bulk export is happening. Maybe that can be a `subtitle-controller` notification.

common/anki/anki.ts
```diff
@@ -375,6 +375,18 @@ export class Anki {
                 }
 
                 throw new Error('Could not update last card because the card info could not be fetched');
+            case 'bulk':
+                const res = await this._executeAction('addNote', params, ankiConnectUrl).catch((error) => {
```
killergerbah: There's a mixture of async/await and `Promise.catch` in here. I think you can rewrite this to use `try`/`catch` instead which would be slightly easier to read IMO.

```
try {
   const res = await ...
   return (res as any).result;
} catch (error) {
    if (error instanceof Error && error.message.includes('duplicate')) {
        return '';
    }
    throw error;
}
```

common/app/components/BulkExportModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+        >
+            <DialogTitle id="bulk-export-title">
+                {t('info.bulkExport.title', 'Bulk Export in Progress')}
+            </DialogTitle>
+            <DialogContent dividers>
+                <Box display="flex" alignItems="center" justifyContent="center" mb={3}>
+                    <CircularProgress size={64} color="primary" />
+                </Box>
+                <Typography id="bulk-export-description" variant="body1" align="center" gutterBottom>
+                    {t('info.bulkExport.description', 'Exporting subtitles to Anki. Please wait...')}
```
killergerbah, line 48: Can be removed later so that loc falls back to `en.json`.

common/app/services/mining-context.ts
```diff
@@ -1,17 +1,72 @@
-type MiningEvent = 'started-mining' | 'stopped-mining';
+type MiningEvent =
+    | 'started-mining'
+    | 'stopped-mining'
+    | 'bulk-export-started'
```
killergerbah: If you move the bulk logic out of side panel I think these events and queue management will move outside of this file too.

extension/src/handlers/video/record-media-handler.ts
```diff
@@ -112,10 +112,12 @@ export default class RecordMediaHandler {
 
             try {
                 const audioBase64 = await audioPromise;
-                audioModel = {
-                    ...baseAudioModel,
-                    base64: audioBase64,
-                };
+                if (audioBase64 !== '') {
```
killergerbah: If we allow this, the `audioModel` can be  silently `undefined`. In what situation is the return value turning up empty?

Edit: I read farther below and I see that you are returning empty string intentionally. I think an exception should be thrown instead.

extension/src/handlers/video/rerecord-media-handler.ts
```diff
@@ -47,15 +47,22 @@ export default class RerecordMediaHandler {
...
-                ),
-            };
+            const audioBase64 = await this._audioRecorder.startWithTimeout(
+                rerecordCommand.message.duration / rerecordCommand.message.playbackRate +
+                    rerecordCommand.message.audioPaddingEnd,
+                false,
+                { src: rerecordCommand.src, tabId: sender.tab?.id! }
+            );
+            audio =
+                audioBase64 === ''
```
killergerbah: The ternary expression is unnecessary:
```
{
    ...baseAudioModel,
    base64: audioBase64,
}
````

File: extension/src/services/audio-recorder-service.ts [hunk 6]
```diff
@@ -175,9 +177,10 @@ export default class AudioRecorderService {
 
     private _prepareForAudioDataResponse(requestId: string): Promise<string> {
         if (this.audioBase64Promise !== undefined) {
-            this.audioBase64Reject!(
-                new Error('New audio recording request received before the current one could finish')
-            );
+            // During bulk export, a new audio request can arrive while a previous one
```
killergerbah: Rather than resolving with an empty string, which would make "empty string" a possible return value that should be handled by all callers, can we keep the `reject` with an exception that can be optionally handled at the judgment of callers? The exception can be a custom one distinguishable from others.

common/src/model.ts
```diff
@@ -47,6 +47,7 @@ export interface CardModel extends CardTextFieldValues {
     readonly audio?: AudioModel;
     readonly file?: FileModel;
     readonly mediaTimestamp: number;
+    readonly exportMode?: AnkiExportMode;
```
killergerbah: I understand this might be convenient for passing `exportMode` around with the card, but would it be possible to instead add `exportMode` directly to the message interfaces instead?

common/anki/anki.ts
agloo: Yup, apologies for the churn on this--I'm more used to working in a language where exceptions are more for exceptional cases. Just updated with two new exception types for duplicate cards and preeempted audio.

common/src/model.ts
```diff
@@ -88,7 +88,7 @@ export interface AudioModel {
     readonly error?: AudioErrorCode;
 }
 
-export type AnkiExportMode = 'gui' | 'updateLast' | 'default';
+export type AnkiExportMode = 'gui' | 'updateLast' | 'default' | 'bulk';
```
killergerbah: Sorry for realizing this so late, but it doesn't seem like a distinct `AnkiExportMode` for `bulk` is necessary. I see that you're using `bulk` to decide whether a `DuplicateNoteError` should be thrown, and to trigger certain branches in `CardPublisher._exportCard`.
- As for `DuplicateNoteError`: in this case an error will be thrown anyway regardless of the export mode, so it always makes sense to throw the more specific error instead.
- As for `CardPublisher._exportCard`: there are enough branches in this method that it would make sense to have a new `publish` method specifically for bulk export. You could decide which method to call from `record-media-handler.ts`

extension/src/controllers/bulk-export-controller.ts
```diff
@@ -0,0 +1,253 @@
...
+        const startedMessage = {
+            sender: 'asbplayerv2',
+            message: {
+                command: 'bulk-export-started',
+                total: this._queue.length,
+            },
+            src: this._context.video.src,
+        };
+        try {
+            await browser.runtime.sendMessage(startedMessage);
```
killergerbah: I think in this case fire-and-forget is fine. Instead of `awaiting` you can just write:
```
browser.runtime.sendMessage(startedMessage)
```
or
```
browser.runtime.sendMessage(startedMessage).catch(console.error)
```

extension/src/controllers/bulk-export-controller.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    private _notifyProgress() {
+        if (!this._exporting) {
+            return;
+        }
+        const total = this._queue.length;
+        const current = Math.min(this._currentIndex, total);
+        // Show status overlay; use a different message when a duplicate was skipped last
+        if (this._lastSkippedDuplicate) {
```
killergerbah: Also: if `CardExportedMessage.exportError` is defined then take this branch

extension/src/handlers/video/record-media-handler.ts
```diff
@@ -118,7 +118,12 @@ export default class RecordMediaHandler {
                 };
             } catch (e) {
                 if (!(e instanceof DrmProtectedStreamError)) {
-                    throw e;
+                    if (e instanceof AudioRequestSupersededError) {
+                        // Treat as no-audio
+                        // Intentionally leave audioModel undefined
```
killergerbah: In this case, would we be exporting a card without any audio?

Other note: if we get here then we would also reach the lines below:
```
                audioModel = {
                    ...baseAudioModel,
                    error: AudioErrorCode.drmProtected,
                };
```
which I'm guessing is different from your intention.

extension/src/handlers/video/record-media-handler.ts
```diff
@@ -147,16 +152,18 @@ export default class RecordMediaHandler {
             }
         }
 
+        const { exportMode, ...messageWithoutExportMode } = recordMediaCommand.message;
         const card: CardModel = {
             image: imageModel,
             audio: audioModel,
-            ...recordMediaCommand.message,
+            ...(messageWithoutExportMode as Omit<typeof recordMediaCommand.message, 'exportMode'>),
```
killergerbah: Is there a reason you're writing this instead of just:
```
...messageWithoutExportMode,
```

extension/src/handlers/video/rerecord-media-handler.ts
```diff
@@ -47,24 +47,28 @@ export default class RerecordMediaHandler {
...
-                    rerecordCommand.message.duration / rerecordCommand.message.playbackRate +
-                        rerecordCommand.message.audioPaddingEnd,
-                    false,
-                    { src: rerecordCommand.src, tabId: sender.tab?.id! }
-                ),
+                base64: audioBase64,
             };
         } catch (e) {
-            if (!(e instanceof DrmProtectedStreamError)) {
+            if (e instanceof AudioRequestSupersededError) {
```
killergerbah: Was this change necessary? From what I can tell, re-record isn't being used by the bulk feature. I don't think there's a need to check specifically for `AudioRequestSupersededError` here, and with the new logic, execution could continue and publish a card without audio.

extension/src/handlers/asbplayerv2/bulk-export-cancellation-handler.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default class BulkExportCancellationHandler {
+    readonly sender = 'asbplayerv2';
+    readonly command = 'bulk-export-cancelled';
+    
+    constructor(private readonly _cardPublisher: CardPublisher) {}
+    
+    handle(request: any, sender: any, sendResponse: any): boolean {
+        // Check if we already processed this cancellation to prevent duplicate processing
+        if (this._cardPublisher.bulkExportCancelled) {
```
killergerbah: Nitpick: it's probably more efficient and readable to skip this check and just call this:
```
        this._cardPublisher.setBulkExportCancelled(true);
```

extension/src/services/card-publisher.ts
```diff
@@ -62,21 +81,76 @@ export class CardPublisher {
         });
     }
 
-    private async _exportCard(card: CardModel, src: string | undefined, tabId: number) {
+    private async _exportCard(card: CardModel, src: string | undefined, tabId: number, mode?: AnkiExportMode) {
```
killergerbah: I would recommend writing a new public method specifically for bulk export. That would allow you to have less `if` `else` branches in this code.

I wrote in a comment on `model.ts` why I think we should not use a new `AnkiExportMode` to branch for the bulk vs not bulk.

extension/src/ui/components/SidePanel.tsx
```diff
@@ -266,6 +270,66 @@ export default function SidePanel({ settings, extension }: Props) {
...
+        browser.runtime.sendMessage(cancelCommand);
+    }, [syncedVideoTab]);
+
+    // Local bulk export UI state
+    const [bulkOpen, setBulkOpen] = useState<boolean>(false);
+    const [bulkCurrent, setBulkCurrent] = useState<number>(0);
+    const [bulkTotal, setBulkTotal] = useState<number>(0);
+
+    // Force SidePanel to re-render when mining starts/stops so UI can refresh
+    const [, setBulkUiTick] = useState<number>(0);
```
killergerbah: Is this actually being used? From what I can tell, it should not be possible for the effect below to run, as the only usages of `MiningContext.started` and `MiningContext.stopped` are from the `App.tsx`.

common/src/message.ts
```diff
@@ -190,6 +192,10 @@ export interface CardUpdatedMessage extends Message, CardModel {
 export interface CardExportedMessage extends Message, CardModel {
     readonly command: 'card-exported';
     readonly cardName: string;
+    readonly isBulkExport?: boolean;
+    readonly skippedDuplicate?: boolean;
+    readonly exportError?: string;
+    readonly exportMode?: AnkiExportMode;
```
killergerbah: `exportMode` appears to be unused, so I think it can be removed.

common/src/message.ts
```diff
@@ -160,6 +161,7 @@ export interface CopySubtitleMessage extends Message, CardTextFieldValues {
     readonly postMineAction: PostMineAction;
     readonly subtitle?: SubtitleModel;
     readonly surroundingSubtitles?: SubtitleModel[];
+    readonly exportMode?: AnkiExportMode;
```
killergerbah: As mentioned in my comment on `model.ts`, instead a specific `AnkiExportMode` here you can use a `boolean` instead and call different `CardPublisher` methods from `record-media-handler.ts` based on that.

extension/src/handlers/video/record-media-handler.ts
```diff
@@ -147,16 +152,18 @@ export default class RecordMediaHandler {
...
         };
         this._cardPublisher.publish(
```
killergerbah: As mentioned elsewhere, I think you should branch here and call different publish methods instead of using a distinct `AnkiExportMode` to branch inside of `CardPublisher.publish`.

common/anki/anki.ts
```diff
@@ -375,6 +382,16 @@ export class Anki {
                 }
 
                 throw new Error('Could not update last card because the card info could not be fetched');
+            case 'bulk':
+                try {
+                    const res = await this._executeAction('addNote', params, ankiConnectUrl);
+                    return (res as any).result;
+                } catch (error) {
+                    if (error instanceof Error && error.message.includes('duplicate')) {
```
killergerbah: This `if` -> `throw` block can be moved into `_executeAction`

killergerbah (commented):
Sorry for being so late to review this - lately I am short on time during the weekdays. And thanks for making all the requested changes, I know it probably took a big chunk of time.  I think this is really close, hopefully the remaining comments do not take as long to address.

extension/src/controllers/bulk-export-controller.ts
```diff
@@ -0,0 +1,253 @@
...
+        try {
+            await browser.runtime.sendMessage(startedMessage);
+        } catch {}
+
+        // Kick off first item
+        this._sendNext();
+    }
+
+    async cancel() {
+        if (!this._exporting || this._cancelled) {
```
killergerbah, line 143: I was somehow able to trigger an edge-case where we get into this block even though `SidePanel` is still displaying "in progress." So maybe as a defensive measure you can send the `bulk-export-cancelled` from here as well.

common/anki/anki.ts
killergerbah: NP and I don't want to leave you with the impression that my opinions are conventional in the language. I actually love languages with exceptions in them, and I love it when they represent exceptional cases so that it's easier to write code for the non-exceptional cases.

This is my reasoning for when deciding to use an exception vs not:
- Exceptions should be used when a method is unable to fulfill its contract.
- The method's contract should be expressed by the signature as much as possible.

In the case of `AudioRecorder`, I have decided that the `record` methods should always return non-empty audio if they succeed, and therefore any case where that is not possible should be expressed via an exception. My argument for this is that this contract for `record` overall serves the needs of the codebase better, but it's entirely possible that a different contract would be simpler to deal with in another codebase, or that I am wrong.

extension/src/controllers/bulk-export-controller.ts
agloo: This branch shows a duplicate-specific notification, but if you mean we shouldn't just drop other kinds of errors (e.g. Anki is down, network issues) on the floor I agree with you.
To avoid a parameter bomb, I'm going to just put the error handling in bind() above and make notifyProgress() just log progress without errors.

extension/src/handlers/asbplayerv2/bulk-export-cancellation-handler.ts
agloo: Yup, that setter has the same pattern and without side effects from these functions I agree that we should just set the bool. I had this up thinking about the potential for data races (e.g. you have a cancel in flight as you start another bulk export), but these seem benign enough that I'll just make it a public bool and set it directly here for simplicity.

extension/src/handlers/video/record-media-handler.ts
agloo: The idea was that since this error is only possible in bulk export we'd like to plow through and not hold up an export. Thinking more, I don't think the current architecture can realistically even result in a superseded request, since:
- Bulk export disables the sidebar export button
- Recording disables the bulk export button

I'm going to remove the special error type here in favor of an error with a string, since this is more an exceptional thing that I'd like to know about if it happens. Sound good?

extension/src/handlers/video/record-media-handler.ts
agloo: Nope, this was overly verbose. Removing it in the export-mode-to-flag change...

extension/src/handlers/video/rerecord-media-handler.ts
agloo: Addressed by removing the error (see above)

extension/src/services/card-publisher.ts
agloo: Makes sense

extension/src/ui/components/SidePanel.tsx
agloo: Yup, it's dead code and the afaik react will just rerender when the listener below updates the state. Removing it...

common/anki/anki.ts
agloo: Will do.
In doing this, I realized that it's redundant with the `error.message.includes('duplicate')` in card-publisher.ts' _exportCardBulk. This seems like it could conflate errors from other sources, so I'll remove that and keep the throw in anki.ts.

agloo (commented):
Thanks for your patience!

common/app/components/Player.tsx
```diff
@@ -1075,6 +1075,7 @@ const Player = React.memo(function Player({
                         settings={settings}
                         keyBinder={keyBinder}
                         webSocketClient={webSocketClient}
+                        exporting={miningContext.bulkExporting}
```
killergerbah: Looks like this can be removed now

extension/src/services/card-publisher.ts
```diff
@@ -79,6 +99,63 @@ export class CardPublisher {
...
+                    message: {
+                        ...card,
+                        command: 'card-exported',
+                        cardName: '',
+                        isBulkExport: true,
+                        skippedDuplicate: true,
+                    },
+                    src,
+                };
+                try { await browser.tabs.sendMessage(tabId, cardExportedCommand); } catch {}
```
killergerbah: This method of calling `sendMessage` will swallow errors. Even if it's extremely unlikely to fail, I would still recommend calling `sendMessage(...)` without any silent catches, especially since any failure of `sendMessage` would almost surely cause noticeably bad behavior.

In this case, there's also no need to await either since we're not reading any response. Same comment for other instances of this pattern.

extension/src/services/card-publisher.ts
```diff
@@ -79,6 +99,63 @@ export class CardPublisher {
...
+                    command: 'card-exported',
+                    cardName: '',
+                    isBulkExport: true,
+                    skippedDuplicate: false,
+                    exportError: e instanceof Error ? e.message : 'Unknown error',
+                },
+                src,
+            };
+            try { await browser.tabs.sendMessage(tabId, cardExportedCommand); } catch {}
+            try { await browser.runtime.sendMessage(cardExportedCommand); } catch {}
```
killergerbah: `browser.runtime.sendMessage` can be removed. Same for below.

extension/src/ui/components/SidePanel.tsx
```diff
@@ -266,6 +270,54 @@ export default function SidePanel({ settings, extension }: Props) {
...
+    // Local bulk export UI state
+    const [bulkOpen, setBulkOpen] = useState<boolean>(false);
+    const [bulkCurrent, setBulkCurrent] = useState<number>(0);
+    const [bulkTotal, setBulkTotal] = useState<number>(0);
+
+    // Listen for bulk export lifecycle messages from background
+    useEffect(() => {
+        const listener = (message: any) => {
+            if (message?.sender === 'asbplayerv2' && message?.message?.command === 'bulk-export-started') {
+                const total = (message.message as Message & { total?: number }).total ?? 0;
```
killergerbah: You can cast to `BulkExportStartedPayload` instead. If you use the existing type, you will get some additional safety because it'll be more likely that the message sent and the message received are the same. For example, if you change the message type you would be forced to also change the payload at the sending and receiving sites.

extension/src/handlers/video/record-media-handler.ts
```diff
@@ -117,14 +117,14 @@ export default class RecordMediaHandler {
                     base64: audioBase64,
                 };
             } catch (e) {
-                if (!(e instanceof DrmProtectedStreamError)) {
+                if (e instanceof DrmProtectedStreamError) {
```
killergerbah: This is a nitpick - the logic remains unchanged from before, but the code is now harder to read since the `throw` has been moved down.

extension/src/handlers/video/rerecord-media-handler.ts
```diff
@@ -47,24 +47,25 @@ export default class RerecordMediaHandler {
...
-                    rerecordCommand.message.duration / rerecordCommand.message.playbackRate +
-                        rerecordCommand.message.audioPaddingEnd,
-                    false,
-                    { src: rerecordCommand.src, tabId: sender.tab?.id! }
-                ),
+                base64: audioBase64,
             };
         } catch (e) {
-            if (!(e instanceof DrmProtectedStreamError)) {
+            if (e instanceof DrmProtectedStreamError) {
```
killergerbah: Same comment as above. I think it's easier to read if the `audio` assignment is closer to where the code will continue.

extension/src/services/binding.ts
```diff
@@ -635,6 +639,24 @@ export default class Binding {
                             subtitleFileNames: this.subtitleController.subtitleFileNames ?? [],
                         });
                         break;
+                    // This is useful because when we kick off bulk export the side panel needs to know
+                    // what subtitle to start from.
+                    case 'request-current-subtitle':
+                        const [currentSubtitle] = this.subtitleController.currentSubtitle();
```
killergerbah, line 645: `SubtitleController.currentSubtitle()` can be changed to return `SubtitleModelWithIndex` so you can avoid using `findIndex`. That will guarantee you get the right index.

restorebooks1-prog:
Thanks for your work; I really appreciate it! I'm also a programmer, and I’ve been one for a year. I have to say, I’m not able to do such excellent work.

Issue #735, opened 2025-07-04
Support Multiple Playback Modes at the Same Time


nzarbayezid:
It would be a great improvement if we could enable multiple Playback Modes at once.

For example, I want to use Condensed Mode to play only the dialogue parts of a movie. At the same time, I’d like to enable Auto-Pause so the video automatically stops after each line. This way, I can repeat the dialogue (a method known as Shadowing) and practice speaking. Then, I can press the spacebar to continue to the next line. It's a powerful combination for language learning.

Why is this useful?
Shadowing—repeating what you hear as you watch—is a well-known and effective technique for learning languages. This feature would make that process much easier and more natural.

Also, Playback Mode is one of the best parts of asbplayer, it’s incredibly useful for both shadowing and sentence mining.

Thanks to the dev, amazing tool!

mseh1128:
Hi. I'd like to work on this feature. I'd also find it useful to enable certain Playback modes at the same time. 

To me, it seems like "Condensed" and "Fast Forward" are the only mutually exclusive Playback modes. I think the rest (Auto-Pause, Repeat) make sense to be enabled at the same time. Please let me know if I'm missing something.

killergerbah:
Yeah that sounds right. It might get a little complicated handling the interaction between play-modes enabled at the same time but good luck.

PR #791, opened 2025-09-29
Allow multiple enabled playback modes

mseh1128:
Implementation for #735 

Took forever, but was able to get multiple playback modes working at the same time. In particular, handling the interactions for "Repeat+Autopause" and "Autopause+Condensed" was tricky and required some significant (and hopefully not too hacky) changes to Player.tsx/binding.ts.


killergerbah:
Hey thanks for doing this. I've been pretty busy but will try to read this on the weekend.

mseh1128:
Btw I was testing this some more today, and realized I forgot about Auto Pause at start of subtitles. I'm working on making this work w/ Repeat mode.

killergerbah:
@mseh1128 On the first pass my only comment is that there seems to be a lot of duplicated code related to modifying the play mode set. Maybe this should be its own collection class? "Resolve conflicts" is the only part that is different depending on the context, but it should still be possible to extract that out from all the duplicated lines.

mseh1128:
Thanks for the review. I refactored most of the playMode related code into play-mode-manager.ts. Also, I was able to get "auto pause at start" working with Repeat and Condensed modes surprisingly easily.

extension/src/services/binding.ts
```diff
@@ -1242,6 +1311,11 @@ export default class Binding {
     }
 
     async play() {
+        if (this._playModes.has(PlayMode.repeat) && this._pendingRepeatTime > 0) {
```
killergerbah: I think you'll need to intercept the video's `play` event instead. The user might play the video through its interface instead of using an asbplayer keyboard shortcut.

common/app/services/video-channel.ts
```diff
@@ -441,10 +442,10 @@ export default class VideoChannel {
...
     }
 
-    playMode(playMode: PlayMode) {
-        const message: PlayModeMessage = {
-            command: 'playMode',
-            playMode: playMode,
+    playModes(playModes: Set<PlayMode>) {
+        const message: PlayModesMessage = {
+            command: 'playModes',
+            playModes: Array.from(playModes),
```
killergerbah: From what I've seen it's idiomatic to use `[...playModes]` to make an array copy

common/app/services/play-mode-manager.ts
```diff
@@ -0,0 +1,131 @@
...
+        for (const mode of oldModes) {
+            if (!newModes.has(mode)) {
+                removed.add(mode);
+            }
+        }
+
+        return { added, removed };
+    }
+
+    static normalizePlayModes(modes: Set<PlayMode>, oldModes: Set<PlayMode>): Set<PlayMode> {
```
killergerbah: I think you can remove this method and rely solely on the `toggle` method, as this method seems to essentially be duplicating the conflict detecting logic of `toggle`. There is only one place where you're using the `Binding.playModes` setter. Instead of using the setter you can use `togglePlayMode` instead.

common/app/components/Player.tsx
```diff
@@ -895,22 +974,37 @@ const Player = React.memo(function Player({
...
-                updatePlaybackRate(1, true);
-            }
+            setPlayModes((prevModes) => {
+                const manager = new PlayModeManager(prevModes);
+
+                if (targetMode === PlayMode.normal && prevModes.has(PlayMode.fastForward)) {
+                    updatePlaybackRate(1, true);
+                }
+
+                if (targetMode === PlayMode.fastForward && prevModes.has(PlayMode.fastForward)) {
```
killergerbah: Is this intentional?

common/app/components/Player.tsx
```diff
@@ -720,6 +789,7 @@ const Player = React.memo(function Player({
                 clock.stop();
             }
 
+            pendingTimeRef.current = 0;
```
killergerbah: Can you make the name of this variable more clear? Maybe `pendingAutoRepeatTargetTimestamp`?
Also, is it possible to centralize the resetting of this timestamp? It looks like you're resetting it any time the user intentionally seeks somewhere.

common/app/components/Player.tsx
```diff
@@ -233,31 +241,74 @@ const Player = React.memo(function Player({
...
+                    seek(subtitle.start, clock, true);
+                }
+                return;
+            }
+
+            if (isCondensedEnabled) {
+                const slice = subtitleCollection.subtitlesAt(subtitle.end + 1);
+
+                if (slice.nextToShow && slice.nextToShow.length > 0) {
+                    const nextSubtitle = slice.nextToShow[0];
```
killergerbah, line 301: Is this handling the condensed playback behavior as well? If so then you should remove the old handling in the effect farther down this file.

killergerbah (commented):
Hey thanks for doing that refactor. Did a second pass.

common/app/components/Player.tsx
mseh1128: Yep. If fastForward was previously enabled and we are toggling it off now, we update the playback rate to 1.

There was similar logic in the previous togglePlayMode() that I was trying to replicate:

```
const newPlayMode = playMode === togglePlayMode ? PlayMode.normal : togglePlayMode;
setPlayMode(newPlayMode);
onPlayModeChangedViaBind(playMode, newPlayMode);
channel?.playMode(newPlayMode);

if (playMode === PlayMode.fastForward) {
    updatePlaybackRate(1, true);
}
```

killergerbah: I see, could this be simplified by checking if `newModes` doesn't contain `PlayMode.fastForward`?

common/app/components/Player.tsx
mseh1128: This handles condensed playback only when auto-pause is enabled. If auto-pause is not enabled, then the old handling w/ the interval happens. I guess we could remove the old handling and have handleOnWillStopShowingSubtitle() always handle condensed mode. I'm not sure how reliable this would be compared to the interval approach though.

common/app/components/Player.tsx
mseh1128: Updated `pendingTimeRef` to `pendingAutoRepeatTargetTimestamp`, and updated `seek()` to handle resetting of the timestamp

common/app/components/Player.tsx
mseh1128: Yep. I updated play mode manager to handle fastForward being removed.

common/app/services/play-mode-manager.ts
mseh1128: Sounds good. I replaced the `Binding.playModes` setter calls w/ `togglePlayMode` instead.

common/app/services/video-channel.ts
mseh1128: Agreed. Made the change.

mseh1128:
Sorry for the delay. I fixed the issues from the previous code review. Thanks 🙂.

common/app/components/Player.tsx
```diff
@@ -220,39 +235,82 @@ const Player = React.memo(function Player({
...
+                        await seek(nextSubtitle.start, clock, true);
+                        lastSeekDurationRef.current = Date.now() - t0;
+
+                        if (wasPlaying) clock.start();
+                    }
+                }
+            }
         },
-        [playMode, clock, mediaAdapter, videoFileUrl, settings.autoPausePreference, seek]
+        [playModes, clock, mediaAdapter, videoFileUrl, settings.autoPausePreference, seek, subtitleCollection, resetPendingAutoRepeatTargetTimestamp]
```
killergerbah: Might be due to the merge but this line breaks for me because it's above where `subtitleCollection` is being initialized

extension/src/services/binding.ts
```diff
@@ -125,9 +126,15 @@ export default class Binding {
     private recordingMediaStartedTimestamp?: number;
     private recordingMediaWithScreenshot: boolean;
     private pausedDueToHover = false;
-    private _playMode: PlayMode = PlayMode.normal;
+    private _playModes: Set<PlayMode> = new Set([PlayMode.normal]);
     private _seekDuration = 3;
     private _speedChangeStep = 0.1;
+    private _pendingAutoRepeatTargetTimestamp = 0;
+    private _lastKnownTime = 0;
```
killergerbah: The value of this field doesn't seem to be used anywhere. Did you mean to store only `_pendingAutoRepeatTargetTimestamp` instead? I think `_pendingAutoRepeatTargetTimestamp` is not properly being reset sometimes. For example, it's pretty easy to trigger a bug where auto-repeat is enabled, but seeking to the next subtitle still retains the repeat timestamp for the previous one.

mseh1128: Looks like I removed the resetting of the timestamp in seekedListener() by accident. I added it back and tested so should work now.

common/app/components/Player.tsx
mseh1128: Yep, nice catch. `subtitleCollection` was initialized further down in the merge. I moved it back up so should work now.

killergerbah:
@mseh1128 I did some more testing and found this issue where condensed mode stops working when its enabled with other modes at the same time. At the beginning, condensed mode is demonstrated working. When navigating 5 seconds back into a no-subtitle-region, the player automatically snaps back to the beginning of the ～♪ subtitle. But when auto repeat or auto pause is enabled with condensed mode at the same time, seeking back 5 seconds no longer snaps you back.

https://github.com/user-attachments/assets/0a1a1f13-e8a7-4ca0-a236-de8b996a3380

It's probably too late but I wonder if this complexity can be dealt with by implementing multiple play modes as pure, reusable and testable logic rather than having to implement it twice in two different locations close to the UI.

mseh1128:
> @mseh1128 I did some more testing and found this issue where condensed mode stops working when its enabled with other modes at the same time. At the beginning, condensed mode is demonstrated working. When navigating 5 seconds back into a no-subtitle-region, the player automatically snaps back to the beginning of the ～♪ subtitle. But when auto repeat or auto pause is enabled with condensed mode at the same time, seeking back 5 seconds no longer snaps you back.
> Screen.Recording.2026-01-18.at.9.52.51.AM.mov
> 
> It's probably too late but I wonder if this complexity can be dealt with by implementing multiple play modes as pure, reusable and testable logic rather than having to implement it twice in two different locations close to the UI.

I had previously disabled the ```useEffect()``` interval for Condensed mode when auto-pause/repeat was enabled and instead handled it in ```handleOnWillStopShowingSubtitle()```. That's why the snap back wasn't happening. I had done this b/c the ```useEffect()``` interval was interfering with the Auto-Pause functionality (i.e. the video was skipping to the next subtitle rather than auto-pausing). I'm trying a different approach this time by checking whether ```pendingAutoRepeatTargetTimestamp > 0``` within the interval (indicating that an Auto-pause just happened) rather than just disabling the ```useEffect()``` like I had done before. This should fix the issue you were having.

Also, while testing I realized that seeking via keyboard shortcuts or the progress bar was broken due to some code I had removed before. I tried fixing this by using ```forwardToMedia``` to determine whether the action was user-initiated or not, and that seems like it's working. But I'm not sure if it is always the case that ```forwardToMedia``` should cause the ```pendingAutoRepeatTargetTimestamp``` to reset.

mseh1128:
I agree that implementing multiple play modes as reusable, testable logic would've been better. Especially for testing purposes. That being said, there are a few significant differences in how multiple play modes are handled in the extension vs the web app. I haven't looked much at the testing framework here, but it would be nice to have testable logic for some of the different Play Mode interactions.

common/app/components/Player.tsx
```diff
@@ -715,7 +715,11 @@ const Player = React.memo(function Player({
                     clock.stop();
                 }
 
-                await seek(currentTime * 1000, clock, forwardToMedia);
+                // When forwardToMedia is false, the message came from the video element's seeked event,
```
mseh1128, line 820: This is the logic I had added to deal with seeks via the progress bar or keyboard shortcuts. This seems like the entrypoint for those seeks in Player.tsx, which is why I added it here. I'm not sure if this ```forwardToMedia``` check is always applicable though.

mseh1128: Sorry. part of the code snippet got cut off. You can see it in my latest commit.

killergerbah:
Sorry for delay @mseh1128 
Seems like it's working. But to merge this I think the playback modes UI needs to be revised. E.g. using checkmarks instead of normal list, so that it's obvious that multiple play modes can be selected. I can attempt this if you'd like, might just take a while for me to find the time to do it.

mseh1128:
No worries. I should have some time soon to try making the checkmarks change.

mseh1128:
I updated the playback modes UI to use checkmarks. See screenshots below.

Also, I added a change to not close the dropdown when a play mode is selected, since the user may want to select multiple playback modes at the same time.  


Web App:
image

Extension:
image

killergerbah:
Amazing work @mseh1128 
Thank you for your patience and diligence pushing this through

PR #792, opened 2025-10-01
Update Indonesian Translation for Version 1.12

yaaacha:
I found some minor mistranslation context in keyboard shortcut and translating some untranslated part for new version update


killergerbah:
Awesome, thank you

PR #802, opened 2025-10-12
Add .webm support to file input extensions

LuqueDaniel:
* Added `.webm` to accepted file extensions.
* Moved file input `accept` value to a constant.


killergerbah:
Thanks!

PR #803, opened 2025-10-12
feat(firefox): enable ws connection for one-click mining

mwojick:
Seems to work on Firefox now. Not sure what changed, maybe an issue with MV3?


killergerbah:
@mwojick Wow great find.. yeah something must have changed. I typed "mixed content extension" into the Mozilla bug tracker and nothing that looks related came up. It's possible that I tested incorrectly too but that seems hard to screw up.

PR #809, opened 2025-10-23
enhancement(plex): support internal subtitles

### LINKED ISSUE #580: Plex support?
i see asbplayer currently supports both Emby and Jellyfin, can we have Plex support too? 😢

ShanaryS:
fixes #580

This is a follow to my PR last year #533 which added external subtitle support to Plex. Plex finally added documentation to their api [last month](https://www.plex.tv/blog/plex-pro-week-25-api-unlocked/). This made it much easier to tackle this problem though I probably gave up too early last time.

The roadblock from before was that while I had the right endpoint, I needed the right arguments to have Plex deliver the entire subtitle. Now with the docs and a lot of testing, I was able to get this working along with a few other improvements.
* Can now transcode the entirety of a media's internal subtitles, works with no changes to the app. The big downside is it takes a long time, anywhere from 15s to a few minutes depending on the media. This is mitigated by setting Plex to the lowest video quality before choosing to load the sub in asbplayer.
* Will check for image subtitles or if Plex is burning the subtitles and raise an appropriate error with instructions.
* Updated the `basename` to `Show SXXEYY · Episode Title` instead of just `Episode Title`, movies are now `Title · Year` instead of `Title`.
* Subtitle labels now contain more relevant information when available, e.g `Chinese · Mandarin / Traditional (SRT External)` instead of `Chinese (SRT External)`.
* Handle multiple versions of a media existing (relevant for internal subs) for proper drop down selection.
* Use the 3 letter language code for `language` instead of name (2 letter codes also available).

This generally works well but it can fail if there is too much going on. Plex has to transcode the media for this which I assume has some internal limits. This generally should not be a problem unless the user keeps interrupting the transcodes.

One thing I would like to see that's kinda out of scope for this PR is some indication that the subtitle is loading, specifically being fetched. Since it can take a few minutes for Plex to extract the subtitles, it leaves for a poor UX since the dialog closes immediately. I think a spinner replacing the `OK` button and keeping the dialog up until it's fully loaded would go a long way. This is especially painful if the user has auto load on for the the subtitles.

### DIFF
--- a/extension/src/entrypoints/plex-page.ts
+++ b/extension/src/entrypoints/plex-page.ts
@@ -2,11 +2,45 @@ import { VideoDataSubtitleTrack } from '@project/common';
 import { VideoData } from '@project/common';
 import { trackFromDef } from '@/pages/util';
 
+const SUBTITLE_IMAGE_CODECS = ['pgs', 'vobsub']; // Plex will only burn in these subtitles
+
+function generateId(length: number): string {
+    const chars = 'abcdefhijklmnopqrstuvwxyz0123456789';
+    let id = '';
+    for (let i = 0; i < length; i++) {
+        id += chars.charAt(Math.floor(Math.random() * chars.length));
+    }
+    return id;
+}
+
+function buildPlexLabel(stream: Element, codec: string | null, isExternal: boolean): string {
+    let label = '';
+    const title = stream.getAttribute('title');
+    if (title) {
+        const language = stream.getAttribute('language');
+        if (language) {
+            if (title.includes(language)) {
+                label = title.replace(language, `${language} ·`).replace('(', '').replace(')', ''); // e.g. title=Spanish (Latin America)
+            } else {
+                label = `${language} · ${title}`; // e.g. title=Canadian, language=French -> French · Canadian
+            }
+        } else {
+            label = (stream.getAttribute('displayTitle') ?? title ?? 'Unknown').replace('SDH', ' · SDH');
+        }
+    } else {
+        label = (stream.getAttribute('displayTitle') ?? 'Unknown').replace('SDH', ' · SDH');
+    }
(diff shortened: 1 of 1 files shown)


killergerbah:
Hey @ShanaryS thanks for this. It looks like really meticulous work. I tested it out and looks like it's working.
Just some general comments:
- As you've probably already thought about, it would be nice if we could get the internal subtitle track list without having to select it in Plex first. When I selected it in Plex I had to wait a minute or so, then when I selected it in asbplayer I had to wait another minute.
- asbplayer actually supports `pgs` subtitles. Not sure if it's possible to just download the `pgs` file in that case.

killergerbah (approved):

ShanaryS:
@killergerbah Thanks for the quick merge!

> As you've probably already thought about, it would be nice if we could get the internal subtitle track list without having to select it in Plex first.

There is no way around this unfortunately. Plex doesn't allow us to choose what internal subtitle gets transcoded, we just tell it which playback session to transcode for so the user has to select it. The information for all the subtitles and everything we could need is in the responses or even more endpoints. But even if we were to display all the internal subtitles, it's an illusion of choice as they all would transcode the one selected in the Plex player so I opted not to show them.

> When I selected it in Plex I had to wait a minute or so, then when I selected it in asbplayer I had to wait another minute.

You can select it before hitting play. It also can default to a specific language from the settings in your account.

> asbplayer actually supports pgs subtitles. Not sure if it's possible to just download the pgs file in that case.

Plex always burns in image formats which means this command would fail (it actually just hangs indefinitely). The only way to get the internal subtitles is for Plex to transcode them for which they have a lot of baked in logic for.

If I find anything else I'll be sure to add it, but these are the limitations we have the work with for now.

killergerbah:
Got it, thanks for the explanations.

PR #810, opened 2025-10-25
fix(videoName): don't duplicate subtitle label

ShanaryS:
I was seeing things like `Movie - Japanese SDH (SRT External) - Japanese SDH (SRT External)` the app view as well as the source title in my Anki cards.


killergerbah:
Sweet thanks

PR #812, opened 2025-10-26
Fix/youtube csp

mwojick:
Along with the other recent changes this also seems to be needed to fix CSP issues on Youtube, similar to what is done in youtube.ts. I don't fully understand the CSP stuff and there could be more issues but this seems to fixed the issue with mp3 encoding at least

### DIFF
--- a/common/app/components/App.tsx
+++ b/common/app/components/App.tsx
@@ -242,7 +242,7 @@ function App({
             regexFilter: settings.subtitleRegexFilter,
             regexFilterTextReplacement: settings.subtitleRegexFilterTextReplacement,
             subtitleHtml: settings.subtitleHtml,
-            pgsParserWorkerFactory: async () => new pgsParserWorkerFactory(),
+            pgsParserWorkerFactory: () => new pgsParserWorkerFactory(),
         });
     }, [settings.subtitleRegexFilter, settings.subtitleRegexFilterTextReplacement, settings.subtitleHtml]);
     const webSocketClient = useAppWebSocketClient({ settings });


killergerbah:
@mwojick I'm surprised that there is still an issue with the mp3 encoder worker since the intention of 9e6b14d7b9a9ca7322102b9f8ab5f48f3a014181 + 41df16b5cfeed40f33cdaff12106ff4cc86f31a1 was to prevent the worker from being spawned inside the page. Were you observing that the CSP was being applied even though the worker was being spawned from the background script, or that the worker was still being spawned in the YouTube page?

mwojick:
@killergerbah I could've sworn I was getting a CSP error related to the Worker creation even with those changes but now I can't reproduce it anymore. Maybe a caching issue or I just didn't test it right (or maybe some a/b testing going on with yt). I'll close for now though, and thnx for clarifying.

killergerbah:
@mwojick Okay but out of curiosity how were you testing? Make card from YT video on Firefox?

killergerbah:
Also to be fair I think the PGS worker might benefit from this change, it might not be practical or not possible to put it in the background because of all the data that needs to get transferred.

mwojick:
Yeah I usually just test by updating last card on yt + firefox. fwiw this is the error I was seeing before:

```
TypeError: Worker constructor: Sink type mismatch violation 
blocked by CSP video.js:18474:15
    startAudioRecordingErrorResponse moz-extension://2248e4df-f2
97-4df0-a44d-e28d3852f318/content-scripts/video.js:18474
    listener moz-extension://2248e4df-f297-4df0-a44d-e28d3852f31
8/content-scripts/video.js:19090
    (Async: promise callback)
    listener moz-extension://2248e4df-f297-4df0-a44d-e28d3852f31
8/content-scripts/video.js:19089
Content-Security-Policy: The page’s settings blocked assigning 
to an injection sink because it violates the following 
directive: “require-trusted-types-for ’script’“
Source: Worker 
constructor|moz-extension
```

PR #814, opened 2025-10-30
fix(plex): use unique session ids for decision query

ShanaryS:
If no subtitles are selected, calling `/decision` seems to end the `Plex Transcoder` responsible for the player. This is addressed by using unique ids like for the subtitles.


killergerbah:
Thanks as always @ShanaryS

PR #820, opened 2025-11-08
Add title to keybinds' textfields in settings

iamllama:
Hello there! I sporadically mine from netflix and always forget what shortcuts to use after not having used it for awhile 😅. However, long shortcuts appear cut-off in the popup's settings, and i need to open up the fullscreen settings to see what they are

This pr proposes to show a tooltip by way of the `title` attr (tested on edge/firefox)

image

Admittedly a very minor annoyance, but would be appreciated nonetheless, thanks!


killergerbah:
Thanks!

PR #823, opened 2025-11-12
Pad out Anki Source field for sortability

agloo:
With bulk export here I'm doing reviews of entire episodes in order, and I've realized that individual flashcards don't naturally sort in their Anki ordering. This makes it very hard to look through them in chronological order.
More precisely, if we have two source fields that are:
1m00s
1h00m00s
Then the h will alphanumerically sort before the m and we'll get the later card first.
The fix for this is simple--we just need to properly pad out our numerals in the source field. This would give those two timestamps these strings:
00h01m00.0s
01h00m00.0s
We could also pad out the millisecond count on the audio, but not all flashcards will have audio.

This will give us an ordering that works until we get to a 101-hour-long video, or one that has two subtitles .1 seconds apart--both of these limits seem reasonable.

This is changing `sourceString()` which will affect a few more code sites. As luck would have it those sites correspond to all the cards we create, whether that be in the UI or bulk export or whatever. I figure we'd want the fields consistent across export methods, so I'm not adding any extra args there.

### DIFF
--- a/common/util/util.ts
+++ b/common/util/util.ts
@@ -18,7 +18,7 @@ export function humanReadableTime(timestamp: number, nearestTenth = false, fully
     if (fullyPadded) {
         let secondsStr: string;
         if (nearestTenth) {
-            // For decimal seconds, pad integer part to 2 digits (e.g., 8.2 -> 08.2, 1 -> 01)
+            // For decimal seconds, pad integer part to 2 digits (e.g., 8.2 -> 08.2, 1 -> 01.0)
             const secondsParts = String(seconds).split('.');
             const integerPart = secondsParts[0];
             const decimalPart = secondsParts.length > 1 ? '.' + secondsParts[1] : '.0';


common/util/util.ts
```diff
@@ -15,11 +15,25 @@ export function humanReadableTime(timestamp: number, nearestTenth = false): stri
     const minutes = Math.floor(totalSeconds / 60) % 60;
     const hours = Math.floor(totalSeconds / 3600);
 
-    if (hours > 0) {
-        return hours + 'h' + String(minutes).padStart(2, '0') + 'm' + String(seconds).padStart(2, '0') + 's';
-    }
+    if (fullyPadded) {
+        let secondsStr: string;
+        if (nearestTenth) {
+            // For decimal seconds, pad integer part to 2 digits (e.g., 8.2 -> 08.2, 1 -> 01)
```
killergerbah: I know the comment is referring to the integer part but for clarity it might make sense to write `1 -> 01.0` which represents the actual transformation that will occur.

killergerbah:
Sweet thanks

Issue #818, opened 2025-11-06
Unsupported websites: allow skipping asb's initial dialog window


artjomsR:
# Motivation

Reduce no. of key presses / clicks to start asb player on unsupported websites  

---	 
   
In contrast, when you use asb player on a supported website (e.g. YTube), you've got `Remember these track choices for this site` to minimise the number of clicks to launch asb. It would be good to have some similar functionality for *unsupported* websites. The principle would be the same - automate some of the asb player launch process and reduce friction.  
	    
What could we actually automate on unsupported websites, i.e., what is the workflow? To launch asb player on an unsupported website, the user would normally:
- Open the page with the video
- Invoke some asb shortcut to launch the sub loading initial dialog window
- Either
1) or click `Open files` under each `Subtitle track x`, and then go!
2) Press `Open files`, select the subtitles and go!
- 1 seems too much of a manual process, given you might want to load different individual subs for each video. So I don't think it's possible to automate it. But we could automate the 2nd workflow
	 
# Proposition
If the user selects `Remember these track choices for this site` and clicks `Open files`, we will remember their preference. The next time they visit this website and launch the asb player using a shortcut, the initial asb dialogue will be skipped. Instead, the OS's file selection dialogue (the one which would've appeared after clicking `Open files`) will appear immediately. We then proceed as normal

## UI changes
- We can repurpose the `Remember these track choices for this site` checkbox, so it works for both supported and unsupported websites: "Remember these track ~~choices~~ options for this site"
- I guess we'd need a way for the user to reintroduce this dialog after it got hidden by default. I think the easiest way is for the user to invoke `Select subtitle tracks to load` shortcut with asb player already open. Such consecutive presses should _always_ open the initial asb sub load dialogue. 
- To avoid confusing the user, we can display the standard asb notification when user loads the subtitles on a website for the first time with `Remember these track options for this site` checked. E.g. `asb player will now always prompt to load subtitles by default on this website. To revert this behaviour, use the "Select subtitle tracks to load" shortcut again, once asb player is loaded.`

killergerbah:
@artjomsR This could be a nicer UX, however I don't think the browser can open the file picker automatically. We could still smooth the process by opening the "subtitle select" dialog and highlighting open files?

artjomsR:
[...]

PR #824, opened 2025-11-12
Focus on Open files for unsupported websites

artjomsR:
My attempt at fixing https://github.com/killergerbah/asbplayer/issues/818 

If the website is unsupported, it should now focus on `Open files`, and will focus on OK for all the supported websites, just as before. I'm not sure if there's anything else that can be done for that issue, or would this PR close the original issue?

Feel free to modify this PR directly if you feel it's easier than explaining :)

### DIFF
--- a/common/src/model.ts
+++ b/common/src/model.ts
@@ -198,6 +198,7 @@ export interface VideoDataUiModel {
     defaultCheckboxState?: boolean;
     settings: VideoDataUiSettings;
     hasSeenFtue: boolean;
+    isSupportedPage: boolean;
     hideRememberTrackPreferenceToggle: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


artjomsR:
Also, I think it'd be beneficial to add a `isSupportedPage` on https://github.com/killergerbah/asbplayer/blob/4558b4aeba7259d26b59a6525eb77657e023b1f4/extension/src/ui/components/VideoDataSyncDialog.tsx#L308 - I'm not sure if there's any benefit to displaying the `Remember these track choices for this site` checkbox for unsupported websites?

killergerbah:
>  I'm not sure if there's any benefit to displaying the Remember these track choices for this site checkbox for unsupported websites?

It only matters if you have auto-loading enabled. Then you need to check that box to automatically load an empty track.

killergerbah:
@artjomsR Only one worry I have about this change. I think there are a lot of users who use asbplayer with empty subtitle track - maybe to just take screenshots/record audio or for the video controls. So I think this change would make those users less efficient, even if it makes the ones who want to load subtitle files more efficient. One idea would be to just remember the last thing the user did?

artjomsR:
Fair enough, I might try it sometime later, but it would be a new attempt so might as well close this PR

PR #813, opened 2025-10-29
feat(subtitles): color subtitle words based on Anki intervals

ShanaryS:
fixes #193
fixes #789

This uses the Yomitan API discussed in the issues. It creates a new tab in the UI called `Dictionary` where all these settings are stored. They all can be moved to the existing sections but that would seem to be confusing.

---

## Design:
0. Call AnkiConnect/Yomitan in a way that avoid CORS issues for extension
1. Tokenize text with Yomitan
2. Treat symbols, punctuation, and numbers as known (doesn't contain a unicode letter class character)
3. Lookup tokens in Anki using word fields
	* Use lemmatized token from Yomitan only if inflected version is uncollected
4. Lookup tokens in Anki using sentence fields only if word fields are uncollected
	* Tokenize the sentence from Anki and match by tokens to reduce false positives
	* Use lemmatized token from Yomitan only if inflected version is uncollected
5. Color words based on Anki stability/intervals
    * Stability will be used if the user has FSRS enabled otherwise we fallback to the intervals
    * Tokens are marked with the highest status if there are multiple conflicting cards
	* Suspended cards can always be treated normally or as a specific status
	* If only some cards are suspended for a given token, only use choose the status from the unsuspended ones
7. Cache all of these steps at the per token and per subtitle event level
	* Poll Anki on interval and when asbplayer creates/updates a card to trigger recheck of uncollected tokens
	* Never update colors based on changed review or suspended status, unlikely and wasteful
	* Strikethrough tokens with red if Anki or Yomitan connection fails, automatically colors tokens when connection resumes
	* App and SidePanel listens for an event for when the colors are updated
8. Subtitle events are tokenized on the fly for the `showingSubtitles` and a buffer of future events
	* Prioritize building for `showingSubtitles` by cancelling previous for responsive and relevant work (e.g. seeking)
9. Support coloring subtitles from the App without extension being installed
	* `SubtitleController: Extension` | `SubtitlePlayer: App`
    * `SubtitlePlayer` always handles coloring the website or SidePanel, listens for extension requests if in use

## UI/Options
- [x] Using naming that is conscious of future expansion
- [x] Ability to customize per track (similar to subtitle appearance)
- [x] Ability to customize the color and line thickness
- [x] Add options on how to treat suspended cards
- [x] Ability to control how inflections and lemmas are handled
- [x] Allow searching multiple word/sentence fields for known status
  - [x] Read fields from Anki and present as autofill multi selection dropdown
- [x] Ability to change card interval mature cutoff
[...]

PR #829, opened 2025-11-18
feat: add support for converting Netflix-style ruby text to HTML

mcgrizzz:
This adds optional parsing of netflix style ruby text: https://partnerhelp.netflixstudios.com/hc/en-us/articles/215767517-Japanese-Timed-Text-Style-Guide

I encounter this a lot, I'm sure others do too. But the netflix style ruby text can look really messy when we don't parse it. Here's a visual example of the new behavior:

## Convert Ruby: OFF (original behavior)
image
image

## Convert Ruby: ON + Render HTML: Render
image
image

## Convert Ruby: ON + Render HTML: Remove (This effectively removes the ruby)
image
image

### DIFF
--- a/common/subtitle-reader/subtitle-reader.ts
+++ b/common/subtitle-reader/subtitle-reader.ts
@@ -3,10 +3,10 @@ import SrtParser from '@qgustavor/srt-parser';
 import { WebVTT } from 'vtt.js';
 import { XMLParser } from 'fast-xml-parser';
 import { SubtitleHtml, SubtitleTextImage } from '@project/common';
-import { convertNetflixRubyToHtml } from '../util';
 
 const vttClassRegex = /<(\/)?c(\.[^>]*)?>/g;
 const assNewLineRegex = RegExp(/\\[nN]/, 'ig');
+const netflixRubyRegex = /([\p{sc=Hira}\p{sc=Kana}\p{sc=Han}々〆〤ヶ]+)\((?=[^)]*[\p{sc=Hira}\p{sc=Kana}])([^)]+)\)/gu;
 const helperElement = document.createElement('div');
 
 interface SubtitleNode {
@@ -501,6 +501,12 @@ export default class SubtitleReader {
         return helperElement.textContent ?? helperElement.innerText;
     }
 
+    private _convertNetflixRubyToHtml(text: string, enabled: boolean): string {
+        if (!enabled) return text;
+
+        return text.replace(netflixRubyRegex, (_match, base, ruby) => `<ruby><rb>${base}</rb><rt>${ruby}</rt></ruby>`);
+    }
+
     private _xmlParser() {
         if (this.xmlParser === undefined) {
             this.xmlParser = new XMLParser({
@@ -518,7 +524,7 @@ export default class SubtitleReader {
                 ? text
                 : text.replace(this._textFilter.regex, this._textFilter.replacement).trim();
 
-        text = convertNetflixRubyToHtml(text, this._convertNetflixRuby);
+        text = this._convertNetflixRubyToHtml(text, this._convertNetflixRuby);
 
         if (this._removeXml) {
             text = this._decodeHTML(text);
--- a/common/util/util.ts
+++ b/common/util/util.ts
@@ -393,22 +393,3 @@ export function seekWithNudge(media: HTMLMediaElement, timestampSeconds: number)
 
     return media.currentTime;
 }
-
-const netflixRubyPattern =
-    /([\p{sc=Hira}\p{sc=Kana}\p{sc=Han}々〆〤ヶ]+)\((?=[^)]*[\p{sc=Hira}\p{sc=Kana}])([^)]+)\)/gu;
-
-/**
- * Parses and converts ruby text notation to HTML ruby elements.
- * For now it supports:
- *  Netflix format(https://partnerhelp.netflixstudios.com/hc/en-us/articles/215767517-Japanese-Timed-Text-Style-Guide)
- *      漢字(かんじ) -> <ruby><rb>漢字</rb><rt>かんじ</rt></ruby>
- *
- * @param text The subtitle text to parse
- * @param enabled Whether ruby text display is enabled
- * @returns The text with ruby markup (if enabled) or original text (if disabled)
- */
-export function convertNetflixRubyToHtml(text: string, enabled: boolean): string {
-    if (!enabled) return text;
-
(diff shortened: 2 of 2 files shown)


mcgrizzz:
Currently unable to push a fix due to github git operations being down... Will push when back up.

common/subtitle-reader/subtitle-reader.ts
```diff
@@ -3,6 +3,7 @@ import SrtParser from '@qgustavor/srt-parser';
 import { WebVTT } from 'vtt.js';
 import { XMLParser } from 'fast-xml-parser';
 import { SubtitleHtml, SubtitleTextImage } from '@project/common';
+import { parseRubyText } from '../util';
```
killergerbah: Looks like this file didn't need to be changed

extension/src/controllers/subtitle-controller.ts
```diff
@@ -455,9 +460,24 @@ export default class SubtitleController {
     }
 
     private _buildTextHtml(text: string, track?: number) {
+        let processedText = parseRubyText(text, this.convertRubyText);
+
+        if (this.subtitleHtml === SubtitleHtml.remove) {
```
killergerbah: I'd like to avoid having to mix settings like this to achieve very specific behaviors. We already have a regex filter that would do the same thing as `convert ruby text + remove html`.
Instead, I think `convert ruby text` should just be a Netflix-specific setting that displays the ruby when enabled and does nothing when disabled. Let me know what you think.

killergerbah (commented):
Thanks for doing this, this is super useful and a lot of people have requested this feature.
My main comment is that I think the setting and how it maps to behavior can be simplified:
- Netflix-only because no other streaming service supplies word readings this way. See `PageSettings`.
- If you turn it on, Netflix should render readings as ruby. If you turn it off, then it should do nothing.
- I think the name of the setting could be clearer. Maybe something like "detect and display ruby." "Convert ruby text" sounds like you already have the ruby, and you need to convert it.

mcgrizzz:
Thanks for the feedback. 

> * Netflix-only because no other streaming service supplies word readings this way. See `PageSettings`.

That is true. Netflix is the only platform that actually *provides* readings in this format. The reason I didn’t want to scope the setting strictly under "Netflix" is that a lot of people (myself included) download subtitles from sites like jimaku that are in Netflix format, then watch them elsewhere (Plex, Crunchyroll, local files, etc.). In that workflow the playback source isn’t Netflix, but the subtitles still contain Netflix-style ruby.

Because of that, I’m a bit worried that putting this under Netflix-specific settings would make it harder to discover for those cases.

> * I think the name of the setting could be clearer. Maybe something like "detect and display ruby." "Convert ruby text" sounds like you already have the ruby, and you need to convert it.

Yup, totally agree.

> I'd like to avoid having to mix settings like this to achieve very specific behaviors. We already have a regex filter that would do the same thing as convert ruby text + remove html.

Somehow I completely missed there was already a regex filter. That is a much cleaner solution.

killergerbah:
Ah yeah you're right. I agree it shouldn't be a Netflix specific setting then. Can just keep it at the top level.

mcgrizzz:
Just pushed the following changes:

1. Remove specific handling in __buildTextHtml_
2. Remove unused import in subtitle-reader.ts
3. Renamed _convertRubyText_ boolean to _convertNetflixRuby_ internally
4. Renamed _parseRubyText_ method to _convertNetflixRubyToHtml_ to make it more clear what it is doing
5. Setting Message changed from "Convert Ruby Text" to "Detect and Display Ruby"

common/subtitle-reader/subtitle-reader.ts
```diff
@@ -58,18 +58,21 @@ const sortVttCues = (list: VTTCue[]) => {
 export default class SubtitleReader {
```
killergerbah, line 59: Can you move the ruby parsing into this class instead of handling it in `SubtitlePlayer.tsx` and `subtitle-controller.ts`? I'm worried that if we put that type of parsing code too far outside this class then it'll have undesireable interactions with other features that do stuff to the subtitle text, like the [coloring feature ](https://github.com/killergerbah/asbplayer/pull/813) Shanary is working on.

It looks like you sort of started on that approach already in this file.

common/app/components/SubtitlePlayer.tsx
```diff
@@ -239,6 +259,26 @@ const SubtitleRow = React.memo(function SubtitleRow({
     const disabledClassName = disabled ? classes.disabledSubtitle : '';
     const { t } = useTranslation();
 
+    const processedText = useMemo(() => {
+        let text = subtitle.text;
+
+        if (subtitleHtml === SubtitleHtml.remove && convertNetflixRuby) {
```
killergerbah: I think for this feature you can just totally ignore the `subtitleHtml` setting. This setting can operate independently of that.

mcgrizzz:
I've made those changes

The only issue is there's some unexpected behavior when you change the HTML remove/render settings with subtitles already loaded. The correct behavior occurs when you reload the subtitle though. 

I guess this is because once subtitle-reader.ts processes the subtitle, it doesn't reprocess when the user changes the Render HTML or Display Ruby setting.

Examples:

[`Render HTML`: `RENDER`] + [`Detect and Display Ruby`: `ON`]
1. Load subtitles
2. Correct ruby text:（<ruby><rb>炭治郎</rb><rt>たんじろう</rt></ruby>）ハア ハア… 
3. Set [`Render HTML`: `REMOVE`]
4. You now see something like: `（<ruby><rb>炭治郎</rb><rt>たんじろう</rt></ruby>）ハア ハア… `
5. Reload the subtitle with the same settings and you get the expected （炭治郎）ハア ハア…

[`Render HTML`: `REMOVE`] + [`Detect and Display Ruby`: `ON`]

1. Load Subtitles
2. You see the correct （炭治郎）ハア ハア…
3. Set [`Render HTML`: `RENDER`]
4. You now incorrectly still see （炭治郎）ハア ハア…

Obviously what I did with the useMemo is a bit hacky, but it did fix this weird toggle behavior. I'm happy to do whatever you think is best though.

common/subtitle-reader/subtitle-reader.ts
```diff
@@ -3,6 +3,7 @@ import SrtParser from '@qgustavor/srt-parser';
 import { WebVTT } from 'vtt.js';
 import { XMLParser } from 'fast-xml-parser';
 import { SubtitleHtml, SubtitleTextImage } from '@project/common';
+import { convertNetflixRubyToHtml } from '../util';
```
killergerbah: I think this function can just be a private method in `SubtitleReader` as shouldn't be used anywhere else for now.

PR #830, opened 2025-11-19
Allow duplicate Anki exports

agloo:
This causes missing cards on bulk export, since when you're creating a deck from an 800+ line piece of media you're bound to have some repeated words like short salutations and interjections.

Anki is okay with duplicates being exported and there are use cases, some outside of bulk export, e.g.:
* What if you have matching text-only card and want to practice audio?
* What if you're going through a deck created from a piece of media in chronological order and a short sentence is important for the plot?

So I'm flipping the flag to allow this.

### DIFF
--- a/extension/src/services/card-publisher.ts
+++ b/extension/src/services/card-publisher.ts
@@ -107,6 +107,9 @@ export class CardPublisher {
             cardName = await exportCard(card, ankiSettings, 'default');
         } catch (e) {
             if (e instanceof DuplicateNoteError) {
+                // This shouldn't be reachable right now, since we've allowed duplicate note exports.
+                // It's left in because without it we'd halt a bulk export on the first duplicate we
+                // find, and duplicates are very easy to encounter on shorter sentences.
                 const cardExportedCommand: ExtensionToVideoCommand<CardExportedMessage> = {
                     sender: 'asbplayer-extension-to-video',
                     message: {


extension/src/services/card-publisher.ts
```diff
@@ -107,6 +107,9 @@ export class CardPublisher {
             cardName = await exportCard(card, ankiSettings, 'default');
         } catch (e) {
             if (e instanceof DuplicateNoteError) {
+                // This shouldn't be reachable right now, since we've allowed duplicate note exports.
+                // It's left in because without it we'd halt a bulk export on the first duplicate we
```
killergerbah, line 111: So this is defensive? I was confused by the comment because the second line seems to suggest we definitely could get here.

killergerbah (approved):

extension/src/services/card-publisher.ts
agloo: Yup, it's defensive.
By the second line I meant to say "If, hypothetically, we undid this PR's flag change started throwing an error on duplicate again then we could definitely get here, and for that reason I'm leaving this block here to avoid breaking bulk export on the off chance someone wants to flip that flag back, but with the code as it is it should be impossible to hit."

killergerbah: Got it

PR #833, opened 2025-11-30
Properly fetch subtitles from Jellyfin with a base URL configured

### LINKED ISSUE #586: Jellyfin internal subtitles are not detected
**What is the issue**
Jellyfin internal subtitles are not detected

**What do you do to make it happen**
On both Jellyfin 10.9.11 and 10.10.3, load a video with internal subtitles (automatically loaded from the .srt file which is in the same directory as the video file). Confirm that the expected subtitle is being used from the CC menu, and if not, select it. I have "Auto-load detected subtitles", so the internal subtitles should be loaded automatically. From there:

- asbplayer's subtitle view won't start up.
- the side panel shows "Load subtitles onto a video to start using asbplayer."
- Clicking "Load Subtitles" from the asbplayer right-click menu opens the "Select Subtitles" modal with all three "Subtitle Track" dropdowns showing "Empty" and containing no options.
- Jellyfin plays with regular .srt subtitles showing, not asbplayer's.

Loading external subtitles with drag-and-drop does still work.

**What browser?**
- Chrome 131.0.6778.140
- Brave 1.71.121 
_All other extensions disabled, as well as Brave Shields_

Streaming

v1.6.1

Jellyfin

Hawkheart:
Instead of assuming that the Jellyfin server is running on `/`, use the API path from the API client we're hooking as a base.

fixes #586

### DIFF
--- a/extension/src/entrypoints/emby-jellyfin-page.ts
+++ b/extension/src/entrypoints/emby-jellyfin-page.ts
@@ -17,7 +17,7 @@ export default defineUnlistedScript(() => {
                 );
             }
 
-            const deviceID = ApiClient?._deviceId;
+            const deviceID = ApiClient.deviceId();
 
             let session;
             for (let attempt = 0; attempt < 5; attempt++) {
@@ -49,7 +49,7 @@ export default defineUnlistedScript(() => {
             ).forEach((sub: { Codec: string; DisplayTitle: any; Language: any; Index: number; Path: string }) => {
                 const extension = 'srt';
                 var url =
-                    ApiClient._serverAddress +
+                    ApiClient.serverAddress() +
                     '/Videos/' +
                     nowPlayingItem.Id +
                     '/' +


extension/src/entrypoints/emby-jellyfin-page.ts
```diff
@@ -49,7 +49,7 @@ export default defineUnlistedScript(() => {
             ).forEach((sub: { Codec: string; DisplayTitle: any; Language: any; Index: number; Path: string }) => {
                 const extension = 'srt';
                 var url =
-                    window.location.origin +
+                    ApiClient._serverAddress +
```
killergerbah: Should this still fallback to the `window.location.origin` ? I'm worried that `ApiClient._serverAddress` could change as the field is prefixed by an underscore which usually indicates its a "private" field?

Hawkheart: I was following the lead of the code a bit above which fetches the device ID in a similar way.

It seems like current versions of both Jellyfin and Emby expose both through "public" getter functions (`ApiClient.deviceId()` and `ApiClient.serverAddress()`) that are probably cleaner to use with less risk of changes.

killergerbah (approved):

killergerbah:
Thanks!

PR #835, opened 2025-12-01
feat(subtitles): support token reading annotations

ShanaryS:
- Use Yomitan to add ruby annotations to the tokens (e.g. furigana).
  - Options are: Always, Learning or below, Unknown or below, Never
  - Tied to colorization being enabled for now, will be easier to enable individually once we move to `IndexedDB`
- Option to display richText only on hover similar to unblur to avoid "cheating"
  - Applies to both coloring and annotations, not much value in separating if the goal is avoiding "cheating"
  - Applies to the entire subtitle event not per token as I feel it's actually a better experience
  - Works by rendering both html but setting richText to `display: none` until hovered where they are swapped.


killergerbah (approved):
This is great, thanks
We may want to move the settings around / change some wording a bit later

PR #836, opened 2025-12-03
Minor updates to dictionary settings wording and arrangement

killergerbah:
Updates the English source text for the new dictionary-related features.

- Colorization -> Annotation seems better as an umbrella term for "enhancing subtitle appearance and content with learning aids"
- Moved the new "readings" setting up to the top since "on hover only" applies to both colorization and readings
- Minor English wording updates for clarity/consistency

image


killergerbah:
Yeah dictionary feels generic enough and not user facing

PR #838, opened 2025-12-04
enhancement(yomitan): support optimizations for tokenization

### LINKED ISSUE #837: Toggling the status of a suspended card also unintentionally affects uncollected cards
@ShanaryS I'm currently using suspended cards as a way to pseudo-import my known words, so I noticed this bug rather quickly as I kept hearing uncollected words and noticed that they weren't being colored correctly. Rather strangely, not every uncollected word is affected, but the majority are.  

With 'Treat Suspended' set to 'Treat Normal', uncollected words are colored normally; however, if suspended cards are assigned a different status (e.g., status 5), uncollected cards are also affected. 

I have uncollected cards set to this 

Image

Treat Normal:

Image

Treat as Status 5:

Image

Intended function:

Image

ShanaryS:
This feature is guarded by a version check so it won't try to use it, it's currently always set to `false` until the Yomitan PR yomidevs/yomitan#2251 is resolved. In the event there are any changes there, the implementation can be easily tweaked before enabling the version check.

I'm making a PR now as the `IndexedDB` code was written on top of it. I refactored the Yomitan class to be per track rather than global as there is too much state to manage now per track. Also fixes #837.

### DIFF
--- a/common/components/DictionarySettingsTab.tsx
+++ b/common/components/DictionarySettingsTab.tsx
@@ -75,8 +75,8 @@ const DictionarySettingsTab: React.FC<Props> = ({ settings, onSettingChanged, an
     const [dictionaryYomitanUrlError, setDictionaryYomitanUrlError] = useState<string>();
     const dictionaryRequestYomitan = useCallback(async () => {
         try {
-            const yomitan = new Yomitan();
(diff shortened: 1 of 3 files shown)


killergerbah:
I don't have any comments to add, just want to test with your suped-up Yomitan when I get a chance this weekend.

PR #844, opened 2025-12-15
feat(dictionary): cache tokens in IndexedDB

ShanaryS:
This eliminates the runtime dependency of Anki (though if it's available, it will be used to automatically keep cache in sync). This gives a massive speedup since we no longer need to query Anki as well as Yomitan for the Anki cards (especially sentence cards). Support for local tokens is not exposed to the user in this PR.

* The IndexedDB is accessed through `DictionaryProvider` (works the same as `SettingsProvider`).
* There is 3 stores used:
	* `meta`: Use for tracking the current and previous builds to know if settings changed or build is in process
	* `tokens`: Stores each word with their lemmas status, etc
	* `ankiCard`: Anki tokens derive their status from this store as we need to track other things like suspended status
* Cache is only re-built when a card/note modification time changes which covers review, suspend, edit, or added changes.
	* During playback we still use the existing method to detect new cards as checking modified times for all cards frequently is not desirable, Anki does not allow querying `cards > modifiedAt`. This means we will only trigger a cache build once for a given card during a playback session (though subsequent card triggers will update everything). This covers the main case of collecting a token but tracking more is too inefficient.
* Builds are guarded by a `buildId` so concurrent builds are prevented even in case of shutdowns with no cleanup. Build will expire when completed or if more than 5 minutes has passed since the last update (every few seconds).
* Builds can happen while it's being queried during playback, updated tokens will be sent (if build was not manually triggered) so that they can be recolored.
* On settings change for a track, it will be cleared next build before being re-populated. If a build is interrupted before completion, it will resume where it left off (and any changes since).
* Triggers were added so that when an Anki card is exported/updated from `SubtitlePlayer`, it triggers a cache build immediately. This already existed if it was initiated from the extension (such as keybind).
* The settings tab will display helper text if the user changes settings that will require a cache rebuild as well as build progress (build status messages are exposed to translations).
* Reading annotations can be enabled independent of coloring, `Always` and `Never` won't trigger cache builds for that track (unless coloring is enabled).
* Added ability to restrict tracks per deck in case multiple unrelated decks share the same note type.
* Disabling Anki is achieved by keeping the fields empty which will clear that track from the db on the next build. Disabled tracks are kept in the db indefinitely.
* Yomitan will require the version with the optimizations in place for tokenize, this significantly improves things and should not be a problem since this feature is unreleased and extensions should auto update.
* When settings are updated, `SubtitleColoring` will now only reset the cache on settings that affect it, rather than all. It will now also clear the richText that's currently being displayed if the user turns off the features as we rely on it being enabled to update it.
* We will now color more intelligently. The first request will only color 10 events, then subsequent requests can be up to 100. They will only be triggered once the user is about to need another batch, rather than single events on each new showing subtitle.
* Local tokens are not per track unlike Anki, this makes it easier for users to switch between tracks and have local tokens behave consistently. If per track local tokens are desired, it can be trivially added as an opt in later.
* Priority for tokens is `Local > Anki word > Anki sentence`
* Added the unlimitedStorage permission which should keep the DB persisted, preventing users from losing local tokens.
* Did not add offline resume, that is not requiring Anki to finish building if it was interrupted. This would only matter for first time builds.
* The limitations that still exist are inherent to this problem domain:
	* Homographs can't be dealt with: bat (animal) and bat (baseball) cannot be separated (obviously just a fundamental limitation with written language). Would need AI for context which is a non-starter.
	* Words that are false friends, same spelling but different meaning across languages. Not worth addressing but mitigated by filtering by decks.

### DIFF
--- a/common/components/DictionarySettingsTab.tsx
+++ b/common/components/DictionarySettingsTab.tsx
@@ -37,8 +37,109 @@ import { Yomitan } from '../yomitan/yomitan';
 import SwitchLabelWithHoverEffect from './SwitchLabelWithHoverEffect';
 import SettingsTextField from './SettingsTextField';
 import SettingsSection from './SettingsSection';
-import { DictionaryBuildAnkiCacheState } from '../src/message';
+import {
+    DictionaryBuildAnkiCacheProgress,
+    DictionaryBuildAnkiCacheState,
+    DictionaryBuildAnkiCacheStateError,
+    DictionaryBuildAnkiCacheStateErrorBuildExpirationData,
+    DictionaryBuildAnkiCacheStateErrorCode,
+    DictionaryBuildAnkiCacheStateErrorTrackNumberData,
+    DictionaryBuildAnkiCacheStateType,
+    DictionaryBuildAnkiCacheStats,
+} from '../src/message';
 import { DictionaryProvider } from '../dictionary-db';
+import { humanReadableTime } from '../util';
+
+const localizedDate = (timestamp: number) => {
+    return new Date(timestamp).toLocaleTimeString([], {
+        hour: '2-digit',
+        minute: '2-digit',
+        second: '2-digit',
+    });
+};
+
+const useBuildAnkiCacheState: () => {
+    severity: 'error' | 'info';
+    msg: string;
+    setBuildAnkiCacheState: (state: DictionaryBuildAnkiCacheState) => void;
+} = () => {
+    const { t } = useTranslation();
+    const [buildAnkiCacheState, setBuildAnkiCacheState] = useState<DictionaryBuildAnkiCacheState>();
+    let msg: string = '';
+
+    if (buildAnkiCacheState !== undefined) {
+        switch (buildAnkiCacheState.type) {
+            case DictionaryBuildAnkiCacheStateType.error:
+                const error = buildAnkiCacheState.body as DictionaryBuildAnkiCacheStateError;
+                switch (error.code) {
+                    case DictionaryBuildAnkiCacheStateErrorCode.concurrentBuild:
+                        msg = t('settings.dictionaryBuildInProgress', {
+                            time: localizedDate(
+                                (error.data as DictionaryBuildAnkiCacheStateErrorBuildExpirationData).expiration
+                            ),
+                        });
+                        break;
+                    case DictionaryBuildAnkiCacheStateErrorCode.noAnki:
+                        msg = t('settings.dictionaryBuildAnkiError');
+                        break;
+                    case DictionaryBuildAnkiCacheStateErrorCode.noYomitan:
+                        msg = t('settings.dictionaryBuildYomitanError', {
+                            trackNumber: (error.data as DictionaryBuildAnkiCacheStateErrorTrackNumberData).track + 1,
+                        });
+                        break;
+                    case DictionaryBuildAnkiCacheStateErrorCode.failedToSyncTrackStates:
+                    case DictionaryBuildAnkiCacheStateErrorCode.failedToBuild:
+                    default:
+                        msg = error.msg ? t('info.error', { message: error.msg }) : t('info.errorNoMessage');
+                        break;
+                }
+                break;
+            case DictionaryBuildAnkiCacheStateType.stats:
+                const stats = buildAnkiCacheState.body as DictionaryBuildAnkiCacheStats;
+                const parts: string[] = [];
+                if (stats.tracksToBuild !== undefined) {
+                    parts.push(
+                        t('settings.dictionaryBuildAnkiTracks', {
+                            tracks: stats.tracksToBuild.map((track) => `#${track + 1}`).join(', '),
+                        })
+                    );
+                }
+                if (stats.modifiedCards !== undefined) {
+                    parts.push(
+                        `${t('settings.dictionaryBuildModifiedCards', { numCards: stats.modifiedCards.toLocaleString('en-US') })}`
+                    );
+                }
+                if (stats.tracksToClear?.length && stats.orphanedCards !== undefined) {
+                    parts.push(
+                        t('settings.dictionaryBuildOrphanedCards', {
+                            numCards: stats.orphanedCards.toLocaleString('en-US'),
+                            tracks: stats.tracksToClear.map((track) => `#${track + 1}`).join(', '),
+                        })
+                    );
+                }
+                if (stats.buildTimestamp !== undefined) {
+                    const duration = Math.floor((Date.now() - stats.buildTimestamp) / 1000);
+                    if (duration > 0) {
+                        parts.push(`[${duration.toLocaleString('en-US')}s]`);
+                    }
+                }
+                msg = parts.join(' | ');
+                break;
+            case DictionaryBuildAnkiCacheStateType.progress:
+                const progress = buildAnkiCacheState.body as DictionaryBuildAnkiCacheProgress;
+                const rate = progress.current / (Date.now() - progress.buildTimestamp);
+                const eta = rate ? Math.ceil((progress.total - progress.current) / rate) : 0;
+                msg = `${progress.current.toLocaleString('en-US')} / ${t('settings.dictionaryBuildModifiedCards', { numCards: progress.total.toLocaleString('en-US') })} [ETA: ${localizedDate(Date.now() + eta)} (${humanReadableTime(eta)})]`;
+                break;
+        }
+    }
+
+    return {
+        severity: buildAnkiCacheState?.type === DictionaryBuildAnkiCacheStateType.error ? 'error' : 'info',
+        msg,
+        setBuildAnkiCacheState,
+    };
+};
 
 const Alert: React.FC<AlertProps> = ({ children, ...props }) => {
     return (
@@ -53,8 +154,6 @@ const Alert: React.FC<AlertProps> = ({ children, ...props }) => {
         </MuiAlert>
     );
 };
-
-const command = 'dictionary-build-anki-cache-state';
 
 interface Props {
     settings: AsbplayerSettings;
@@ -169,9 +268,7 @@ const DictionarySettingsTab: React.FC<Props> = ({
     };
 
     const [buildingAnkiCache, setBuildingAnkiCache] = useState<boolean>(false);
-    const [buildAnkiCacheResult, setBuildAnkiCacheResult] = useState<DictionaryBuildAnkiCacheState | undefined>(
-        undefined
-    );
+    const { severity: buildMessageSeverity, msg: buildMessage, setBuildAnkiCacheState } = useBuildAnkiCacheState();
 
     const handleBuildAnkiCache = useCallback(async () => {
         try {
@@ -179,18 +276,21 @@ const DictionarySettingsTab: React.FC<Props> = ({
             await dictionaryProvider.buildAnkiCache(activeProfile, settings);
         } catch (e) {
             console.error('Failed to send build Anki cache message', e);
-            const message = e instanceof Error ? e.message : String(e);
-            setBuildAnkiCacheResult({ command, msg: message, error: true, modifiedTokens: [] });
+            setBuildAnkiCacheState({
+                type: DictionaryBuildAnkiCacheStateType.error,
+                body: {
+                    code: DictionaryBuildAnkiCacheStateErrorCode.failedToBuild,
+                    msg: e instanceof Error ? e.message : String(e),
+                } as DictionaryBuildAnkiCacheStateError,
+            });
         } finally {
             setBuildingAnkiCache(false);
         }
-    }, [dictionaryProvider, settings, activeProfile]);
+    }, [dictionaryProvider, settings, activeProfile, setBuildAnkiCacheState]);
 
     useEffect(() => {
-        return dictionaryProvider.onBuildAnkiCacheStateChange((state) => {
-            if (state.msg.length) setBuildAnkiCacheResult(state);
-        });
-    }, [dictionaryProvider]);
+        return dictionaryProvider.onBuildAnkiCacheStateChange(setBuildAnkiCacheState);
+    }, [dictionaryProvider, setBuildAnkiCacheState]);
 
     return (
         <Stack spacing={1}>
@@ -218,11 +318,9 @@ const DictionarySettingsTab: React.FC<Props> = ({
                 <Typography variant="caption" color="textSecondary">
                     {t('settings.buildAnkiCacheHelperText')}
                 </Typography>
-                {buildAnkiCacheResult && (
+                {buildMessage && buildMessageSeverity && (
                     <div style={{ marginTop: 8 }}>
-                        <Alert severity={buildAnkiCacheResult.error ? 'error' : 'success'}>
-                            {buildAnkiCacheResult.msg}
-                        </Alert>
+                        <Alert severity={buildMessageSeverity}>{buildMessage}</Alert>
                     </div>
                 )}
             </div>
--- a/common/dictionary-db/dictionary-db.ts
+++ b/common/dictionary-db/dictionary-db.ts
@@ -1,5 +1,12 @@
 import { Anki, escapeAnkiDeckQuery, escapeAnkiQuery, NoteInfo } from '@project/common/anki';
-import { DictionaryBuildAnkiCacheState } from '@project/common';
+import {
+    DictionaryBuildAnkiCacheState,
+    DictionaryBuildAnkiCacheStateError as DictionaryBuildAnkiCacheError,
+    DictionaryBuildAnkiCacheStateErrorCode,
+    DictionaryBuildAnkiCacheStateType,
+    DictionaryBuildAnkiCacheStats,
+    DictionaryBuildAnkiCacheProgress,
+} from '@project/common';
 import {
     AsbplayerSettings,
     dictionaryStatusCollectionEnabled,
@@ -7,11 +14,10 @@ import {
     DictionaryTrack,
     TokenStatus,
 } from '@project/common/settings';
-import { HAS_LETTER_REGEX, humanReadableTime, inBatches, mapAsync } from '@project/common/util';
+import { HAS_LETTER_REGEX, inBatches, mapAsync } from '@project/common/util';
 import { Yomitan } from '@project/common/yomitan/yomitan';
 import Dexie from 'dexie';
 import { v4 as uuidv4 } from 'uuid';
-import i18n, { TFunction } from 'i18next';
 
 const BUILD_MIN_EXPIRATION_MS = 5 * 60 * 1000; // 5 minutes
 
@@ -611,12 +617,8 @@ export class DictionaryDB {
     async buildAnkiCache(
         inputProfile: string | undefined,
         settings: AsbplayerSettings,
-        statusUpdates: (state: DictionaryBuildAnkiCacheState) => void,
-        options?: { t?: TFunction<['translation', ...string[]], undefined> }
+        statusUpdates: (state: DictionaryBuildAnkiCacheState) => void
     ): Promise<void> {
-        const t = options?.t ?? i18n.t;
-        let msg = 'Unknown error during build';
-        let error = true;
         const modifiedTokens = new Set<string>();
 
         const anki = new Anki(settings);
@@ -624,9 +626,15 @@ export class DictionaryDB {
             const permission = (await anki.requestPermission()).permission;
             if (permission !== 'granted') throw new Error(`permission ${permission}`);
         } catch (e) {
-            msg = `${t('settings.dictionaryBuildAnkiError')}`;
-            console.error(`${msg}: ${e}`);
-            statusUpdates({ command, msg, error, modifiedTokens: Array.from(modifiedTokens) });
+            console.error(e);
+            statusUpdates({
+                type: DictionaryBuildAnkiCacheStateType.error,
+                body: {
+                    code: DictionaryBuildAnkiCacheStateErrorCode.noAnki,
+                    msg: e instanceof Error ? e.message : String(e),
+                    modifiedTokens: Array.from(modifiedTokens),
+                } as DictionaryBuildAnkiCacheError,
+            });
             return;
         }
 
@@ -642,28 +650,25 @@ export class DictionaryDB {
             for (const [track, dt] of settings.dictionaryTracks.entries()) {
                 const key: DictionaryMetaKey = [profile, track];
                 let prevSettings: string | null = null;
-                const canBuild = await this.db.transaction('rw', this.db.meta, async () => {
+                const existingBuild = await this.db.transaction('rw', this.db.meta, async () => {
                     if (await this._ensureBuildId(key, buildId, { buildTs })) {
                         prevSettings = (await this.db.meta.get(key))!.settings;
-                        return true;
+                        return undefined;
                     }
-                    const expiration = await this.db.meta
-                        .where('[profile+track]')
-                        .equals(key)
-                        .first()
-                        .then((meta) =>
-                            new Date(meta!.lastBuildExpiresAt).toLocaleTimeString([], {
-                                hour: '2-digit',
-                                minute: '2-digit',
-                                second: '2-digit',
-                            })
-                        );
-                    msg = `${t('settings.dictionaryBuildInProgress', { time: expiration })}`;
-                    console.error(msg);
-                    return false;
+                    return await this.db.meta.where('[profile+track]').equals(key).first();
                 });
-                if (!canBuild) {
-                    statusUpdates({ command, msg, error, modifiedTokens: Array.from(modifiedTokens) });
+                if (existingBuild !== undefined) {
+                    console.warn(`Build already in progress - expires at ${existingBuild.lastBuildExpiresAt}}`);
+                    statusUpdates({
+                        type: DictionaryBuildAnkiCacheStateType.error,
+                        body: {
+                            code: DictionaryBuildAnkiCacheStateErrorCode.concurrentBuild,
+                            modifiedTokens: Array.from(modifiedTokens),
+                            data: {
+                                expiration: existingBuild.lastBuildExpiresAt,
+                            },
+                        } as DictionaryBuildAnkiCacheError,
+                    });
                     return; // Since we set the buildId for all tracks regardless of enabled status, concurrent builds are prevented
                 }
                 activeTracks.push(key);
@@ -677,9 +682,16 @@ export class DictionaryDB {
                 try {
                     await yomitan.version();
                 } catch (e) {
-                    msg = `${t('settings.dictionaryBuildYomitanError', { trackNumber: track + 1 })}`;
-                    console.error(`${msg}: ${e}`);
-                    statusUpdates({ command, msg, error, modifiedTokens: Array.from(modifiedTokens) });
+                    console.error(e);
+                    statusUpdates({
+                        type: DictionaryBuildAnkiCacheStateType.error,
+                        body: {
+                            code: DictionaryBuildAnkiCacheStateErrorCode.noYomitan,
+                            msg: e instanceof Error ? e.message : String(e),
+                            data: { track },
+                            modifiedTokens: Array.from(modifiedTokens),
+                        } as DictionaryBuildAnkiCacheError,
+                    });
                     return;
                 }
                 trackStates.set(track, { dt, yomitan });
@@ -713,14 +725,23 @@ export class DictionaryDB {
                     if (settingsToUpdate.length) await this.db.meta.bulkUpdate(settingsToUpdate); // Set the new settings so we can resume the build if it fails midway
                 });
                 if (!trackStates.size) {
-                    msg = `${t('settings.dictionaryBuildAnkiTracks', { tracks: 'null' })} | ${t('settings.dictionaryBuildOrphanedCards', { numCards: numCardsFromOrphanedTracks.toLocaleString('en-US'), tracks: tracksToClear.map((track) => `#${track + 1}`).join(', ') })}`;
-                    error = false;
-                    statusUpdates({ command, msg, error, modifiedTokens: Array.from(modifiedTokens) });
+                    statusUpdates({
+                        type: DictionaryBuildAnkiCacheStateType.stats,
+                        body: {
+                            tracksToClear,
+                            orphanedCards: numCardsFromOrphanedTracks,
+                            modifiedTokens: Array.from(modifiedTokens),
+                        } as DictionaryBuildAnkiCacheStats,
+                    });
                     return;
                 }
             } else if (!trackStates.size) {
-                msg = `${t('settings.dictionaryBuildAnkiTracks', { tracks: 'null' })}`;
-                statusUpdates({ command, msg, error, modifiedTokens: Array.from(modifiedTokens) });
+                statusUpdates({
+                    type: DictionaryBuildAnkiCacheStateType.stats,
+                    body: {
+                        modifiedTokens: Array.from(modifiedTokens),
+                    } as DictionaryBuildAnkiCacheStats,
+                });
                 return;
             }
 
@@ -739,9 +760,15 @@ export class DictionaryDB {
                     await this._buildAnkiCardStatuses(track, ts, modifiedCards, anki);
                 }
             } catch (e) {
-                msg = `Could not sync track states with Anki: ${e}`;
-                console.error(msg);
-                statusUpdates({ command, msg, error, modifiedTokens: Array.from(modifiedTokens) });
+                console.error(e);
+                statusUpdates({
+                    type: DictionaryBuildAnkiCacheStateType.error,
+                    body: {
+                        msg: e instanceof Error ? e.message : String(e),
+                        code: DictionaryBuildAnkiCacheStateErrorCode.failedToSyncTrackStates,
+                        modifiedTokens: Array.from(modifiedTokens),
+                    } as DictionaryBuildAnkiCacheError,
+                });
                 return;
             }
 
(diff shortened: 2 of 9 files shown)


common/app/components/App.tsx
```diff
@@ -425,6 +433,20 @@ function App({
                 if (settings.lastSelectedAnkiExportMode !== params.mode) {
                     onSettingsChanged({ lastSelectedAnkiExportMode: params.mode });
                 }
+
+                if (params.mode === 'updateLast') {
```
killergerbah: The code might already be inconsistent on this point, but I suggest keeping extension communication inside `chrome-extension.ts`

common/app/components/Player.tsx
```diff
@@ -429,22 +436,37 @@ const Player = React.memo(function Player({
...
         subtitleCollectionRef.current = subtitleColoring;
 
+        const tokensUpdatedListener = (event: MessageEvent) => {
+            if (event.type !== 'message') return;
+            const data: ExtensionToAsbPlayerCommand<DictionaryBuildAnkiCacheState> = event.data;
+            if (data.sender !== 'asbplayer-extension-to-player') return;
+            if (data.message.command !== 'dictionary-build-anki-cache-state') return;
+            subtitleColoring.tokensWereModified(data.message.modifiedTokens);
+        };
+        window.addEventListener('message', tokensUpdatedListener);
```
killergerbah: Same here, subscribing via `ChromeExtension.subscribe` will reduce duplicated code dealing with `window` messages.

common/app/components/Player.tsx
```diff
@@ -478,6 +500,23 @@ const Player = React.memo(function Player({
         void refreshColors();
     }, [extension, tab, onSubtitles]);
 
+    useEffect(() => {
+        const windowListener = (event: MessageEvent) => {
```
killergerbah: Here as well, you can subscribe via `ChromeExtension`

common/components/DictionarySettingsTab.tsx
```diff
@@ -130,14 +193,87 @@ const DictionarySettingsTab: React.FC<Props> = ({ settings, onSettingChanged, an
...
+        };
+        const windowListener = (event: MessageEvent) => {
+            if (event.type === 'message') listener(event.data);
+        };
+        const browserListener = (message: any) => listener(message);
+
+        let extensionContext = true;
+        try {
+            // @ts-ignore
+            browser.runtime.onMessage.addListener(browserListener);
```
killergerbah: These type of checks can be obviated if the implementation is injected via props, similar to how we're dealing with settings storage. I see you're already taking advantage of this by adding additional methods to the `SettingsStorage` interface but I think it would make sense to create a separate abstraction for dictionary-related operations and events.

For example, something like `DictionaryService` which implements the methods that you added to `SettingsStorage`
```
interface DictionaryService {
    dictionaryGetBulk: (profile: string | undefined, track: number, tokens: string[]) => Promise<TokenResults>;
    
    dictionaryGetByLemmaBulk: (profile: string | undefined, track: number, lemmas: string[]) => Promise<LemmaResults>;
    
    dictionarySaveRecordLocalBulk: (
        profile: string | undefined,
        localTokenInputs: DictionaryLocalTokenInput[]
    ) => Promise<DictionaryTokenKey[]>;
    
    dictionaryDeleteRecordLocalBulk: (profile: string | undefined, tokens: string[]) => Promise<number>;
    
    buildAnkiCache: (
        profile: string | undefined,
        settings: AsbplayerSettings,
        options?: { useOriginTab?: boolean }
    ) => Promise<DictionaryBuildAnkiCacheState>;

    // ...
}
```

and additionally implements a way to subscribe to `dictionary-build-anki-cache-state` messages

```
    onAnkiCacheStateChange(...)
```

or similar. Maybe there's a better name than `DictionaryService`.

While using `SettingsStorage` for the above might be convenient, this (hopefully small) refactor would keep the responsibilities separate and also creates a natural location for subscription code like this. You could then inject `DictionaryService` into the settings form instead of `buildAnkiCache`.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1359 @@
...
+    dictionaryYomitanUrl: string;
+    dictionaryYomitanScanLength: number;
+    dictionaryAnkiWordFields: string[];
+    dictionaryAnkiSentenceFields: string[];
+    dictionaryAnkiMatureCutoff: number;
+}
+
+type DictionaryMetaKey = [string, number];
+interface DictionaryMetaRecord {
+    profile: string;
```
killergerbah, line 54: Curious why you chose to include `profile+track` in the primary key? I can see that it might make it easier to keep the database consistent with each profile, since it's easy to detect when profiles change or are removed.

However, I think you would gain some flexibility with the schema, more re-use of database entries, and a better experience for users with lots of similar profiles, if instead you keyed by the relevant settings (or hash of the relevant settings) instead. Then it's just a matter of garbage collecting entries that become orphaned due to settings changes. I think this is probably not a hard problem, since instead of detecting profile-specific changes, you instead detect changes to dependent settings on any profile.

Another problem that just occurred to me - anecdotally some users have language-specific profiles and are likely to target different decks per profile. But we don't currently have a way to target specific decks - just specific fields, which decks may share due to using the same note type.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1359 @@
...
+    token: string;
+    status: TokenStatus;
+    lemmas: string[];
+    states: DictionaryTokenState[];
+}
+
+type DictionaryAnkiCardKey = [number, number, string];
+interface DictionaryAnkiCardRecord {
+    profile: string;
+    track: number;
```
killergerbah, line 83: Same here, I guess `[profile+track]` part of the key is just for garbage collection purposes? With this schema it seems like you would get a lot of duplicated cards across similar profiles, unless we expect most users to have different target word/sentence fields across profiles.

common/app/components/App.tsx
ShanaryS: The goal of this (and the other messages) is to work when no extension is present as well. Does sending through `chrome-extension.ts` handle that case?

common/components/DictionarySettingsTab.tsx
ShanaryS: I considered doing these but just kept it simple to reduce boilerplate on the initial review. This will also be good as this gets a lot more complex for the UI to support some of the later features.

common/dictionary-db/dictionary-db.ts
ShanaryS: My expectation is there won't be a lot of profiles and if there are there won't be much overlap. The DB size for 20k Anki cards (20k tokens as well) is `10MB`. In the worse case of multiple profiles and multiple tracks it could be a few hundred MBs. We also would need the profiles for storing local tokens as I think it makes sense to separate that per profile.

I'll implement the deck selection and think about reducing duplication in the design. Keeping the profile and track makes it easier to ensure consistency and reduce the complexities of managing the db (note that local tokens are trackless) so I'll see if it's worth making the change.

ShanaryS: I think the settings hash works well for Anki and it can be profile agnostic but I think the "hash" should just be the Profile for local tokens. Local tokens don't depend on any setting, even if the parser changes due to the nature of these being entirely user confirmed. But I don't think they should be shared across profiles, I think it's fair to assume that a second profile is of a different language or for a different person.

I think the only concern right now is how to clean up these old hashes across profiles. Right now `SettingsProvider` abstracts away profiles but we would need to be able to access all of them regardless of the one current selected to check all their hashes.

Overall I think the hash design is better and more closely aligns with the intent. But we are trading potential data deduplication (which I don't think should be common or bad even in the worst case) for complexity in state management, which I supposed is always the tradeoff.

common/app/components/App.tsx
killergerbah: Oh I see, yeah `chrome-extension.ts` wouldn't handle that. If you end up separating dictionary code from `SettingsStorage` though you could hide these window message details with that abstraction.

common/dictionary-db/dictionary-db.ts
killergerbah: `SettingsProvider` provides a way to list profiles, so while it would be awkward you could still iterate over the existing dictionary settings:
```
const clonedProvider = new SettingsProvider(storage); // clone to avoid propagating profile changes
const allProfiles = await provider.profiles();
for (const profile of allProfiles) {
    await provider.setActiveProfile(profile);
    // query this profile
}
```

But I agree there would be additional complexity overall to deal with this. I don't feel super strongly about this and I guess even with current implementation, cache building will be invisible to the user anyway, so feel free to do what you think is right.

ShanaryS: I opted to keep `profile+track` over the settings hash as there is too much additional logic needed to make it work. We can move to it in the future if it becomes necessary with a migration, but otherwise I don't think it's worth doing unless we have to.

client/src/components/WebsiteApp.tsx
```diff
@@ -14,18 +16,20 @@ interface Props {
 
 const WebsiteApp = (props: Props) => {
     const extension = useChromeExtension({ component: 'application' });
+    const dictionaryStorage = useMemo(() => {
+        if (extension.supportsAppIntegration) return new AppExtensionDictionaryStorage(extension);
```
killergerbah: Should be `extension.supportsDictionary`

Not for this PR, but should also consider hiding the dictionary feature for people who have an extension installed, but the extension is too old.

common/app/components/Player.tsx
```diff
@@ -454,29 +465,41 @@ const Player = React.memo(function Player({
...
+                if (!prevSubtitles?.length) return prevSubtitles;
                 const allSubtitles = prevSubtitles.slice();
                 for (const s of updatedSubtitles) {
                     allSubtitles[s.index] = { ...allSubtitles[s.index], richText: s.richText };
                 }
                 return allSubtitles;
             });
+        })();
+
+        // Trigger immediate anki refresh if card was modified through the SidePanel
```
killergerbah: This doesn't look like it gets triggered specifically by the side panel, but by the extension Anki UI. The Anki UI could be triggered in other ways, like a keyboard shortcut.

common/components/DictionarySettingsTab.tsx
```diff
@@ -47,22 +50,68 @@ const Alert: React.FC<AlertProps> = ({ children, ...props }) => {
...
+    activeProfile,
+    anki,
+}) => {
     const { t } = useTranslation();
     const { ankiConnectUrl, dictionaryTracks } = settings;
+    const initialDictionaryTracksRef = useRef(dictionaryTracks);
     const [selectedDictionaryTrack, setSelectedDictionaryTrack] = useState<number>(0);
+    const selectedDictionary = dictionaryTracks[selectedDictionaryTrack];
+
+    const isDirty = useCallback(
```
killergerbah: It looks like you could use `areDictionaryTracksEqual` here

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1439 @@
...
+                    msg = `${t('settings.dictionaryBuildAnkiTracks', { tracks: 'null' })} | ${t('settings.dictionaryBuildOrphanedCards', { numCards: numCardsFromOrphanedTracks.toLocaleString('en-US'), tracks: tracksToClear.map((track) => `#${track + 1}`).join(', ') })}`;
+                    error = false;
+                    return { command, msg, error, modifiedTokens: Array.from(modifiedTokens) };
+                }
+            } else if (!trackStates.size) {
+                msg = `${t('settings.dictionaryBuildAnkiTracks', { tracks: 'null' })}`;
+                return { command, msg, error, modifiedTokens: Array.from(modifiedTokens) };
+            }
+
+            const anki = new Anki(settings);
```
killergerbah: Should the permission check be moved farther up? Since by this point card records have been deleted.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1439 @@
...
+        if (modifiedNotes.length) {
+            for (const cardInfo of await anki.cardsInfo(modifiedCardIds)) {
+                modifiedCardsDeck.set(cardInfo.cardId, cardInfo.deckName);
+            }
+            const dts = Array.from(trackStates.values()).map((ts) => ts.dt);
+            for (let i = modifiedNotes.length - 1; i >= 0; i--) {
+                if (
+                    !dts.some(
+                        (dt) =>
+                            (!dt.dictionaryAnkiDecks.length ||
```
killergerbah: It might be because I'm getting old but logic like this takes me a while to read haha

common/dictionary-db/dictionary-provider.ts
```diff
@@ -0,0 +1,85 @@
...
+        options?: { useOriginTab?: boolean }
+    ) => Promise<DictionaryBuildAnkiCacheState>;
+    addBuildAnkiCacheStateChangeCallback: (callback: (message: DictionaryBuildAnkiCacheState) => void) => void;
+    removeBuildAnkiCacheStateChangeCallback: (callback: (message: DictionaryBuildAnkiCacheState) => void) => void;
+    ankiCardWasModified: () => void;
+    addAnkiCardModifiedCallback: (callback: () => void) => void;
+    removeAnkiCardModifiedCallback: (callback: () => void) => void;
+}
+
+export class DictionaryProvider {
```
killergerbah, line 30: It doesn't seem like this class is needed (right now)

common/hooks/use-settings-profile-context.ts
```diff
@@ -29,15 +31,16 @@ export const useSettingsProfileContext = ({ settingsProvider, onProfileChanged }
     );
     const onRemoveProfile = useCallback(
         async (name: string) => {
+            await dictionaryProvider.deleteProfile(name);
```
killergerbah, line 34: I think this should be fire-and-forget instead. This is probably still fast, but it's probably still a lot slower than the async calls below. It's also a garbage-collecting operation, so we don't lose much even if this call fails while the ones below succeed.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -98,114 +141,141 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
-        return dictionaryTrackHoverOnly(this.trackStates[track].dt);
+    ankiCardWasModified() {
+        this.ankiRecentlyModifiedTrigger = true;
     }
 
-    private _tokenStatusValid(tokenStatus: TokenStatus | undefined | null) {
-        if (tokenStatus === undefined || tokenStatus === null) return false;
-        if (tokenStatus === TokenStatus.UNCOLLECTED) return !this.uncollectedNeedsRefresh;
-        return true;
+    hoverOnly(track: number) {
```
killergerbah, line 201: Seems unused

common/settings/settings.ts
```diff
@@ -97,12 +104,15 @@ export enum TokenReadingAnnotation {
...
 
 export function dictionaryTrackEnabled(dt: DictionaryTrack): boolean {
-    return dt.dictionaryColorizeSubtitles;
+    return dt.dictionaryColorizeSubtitles || dt.dictionaryTokenReadingAnnotation !== TokenReadingAnnotation.NEVER;
 }
 
-export function dictionaryTrackHoverOnly(dt: DictionaryTrack | undefined): boolean {
-    if (!dt) return false;
-    return dictionaryTrackEnabled(dt) && dt.dictionaryColorizeOnHoverOnly;
+export function dictionaryStatusEnabled(dt: DictionaryTrack): boolean {
```
killergerbah: Consider renaming to `dictionaryStatusCollectionEnabled`? It took me a little bit to understand the difference between this and `dictionaryTrackEnabled`.

extension/src/handlers/dictionary/dictionary-handler.ts
```diff
@@ -0,0 +1,111 @@
...
+                                    message: state,
+                                };
+
+                                try {
+                                    await browser.runtime.sendMessage(message); // Post updates throughout extension
+                                } catch {
+                                    // No one is currently listening
+                                }
+
+                                if (typeof originTabId !== 'number') return; // Only used when triggered from the App with extension installed
```
killergerbah: Also triggered any time `SubtitleColoring` requests `build-anki-cache` (possibly from any extension content script)?

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -98,114 +141,141 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
                 return;
             }
             this.ankiRecentlyModifiedCardIds = new Set(cardIds);
             if (this.ankiRecentlyModifiedFirstCheck) {
                 this.ankiRecentlyModifiedFirstCheck = false;
                 return;
             }
-            this.uncollectedNeedsRefresh = true;
-            await this._updateAnkiCache();
+            const state = await this.dictionaryProvider.buildAnkiCache(
```
killergerbah: Is calling `buildAnkiCache` twice in this method intentional?

common/dictionary-db/dictionary-provider.ts
```diff
@@ -0,0 +1,85 @@
...
+    buildAnkiCache: (
+        profile: string | undefined,
+        settings: AsbplayerSettings,
+        options?: { useOriginTab?: boolean }
+    ) => Promise<DictionaryBuildAnkiCacheState>;
+    addBuildAnkiCacheStateChangeCallback: (callback: (message: DictionaryBuildAnkiCacheState) => void) => void;
+    removeBuildAnkiCacheStateChangeCallback: (callback: (message: DictionaryBuildAnkiCacheState) => void) => void;
+    ankiCardWasModified: () => void;
+    addAnkiCardModifiedCallback: (callback: () => void) => void;
+    removeAnkiCardModifiedCallback: (callback: () => void) => void;
```
killergerbah: It's probably fine in your diff but this type of API design for event unsubscription scares me because it's easy to forget to call `removeAnkiCardModifiedCallback` or accidentally call `addAnkiCardModifiedCallback` many times in a row. My personal preference is to return an unsubscription callback from every subscription method.

client/src/components/WebsiteApp.tsx
ShanaryS: Yeah I plan to but I can't currently due to the version not being available.

common/app/components/Player.tsx
ShanaryS: It was more so that this code execution should only happen with `tab` is available, so when the subtitles are controlled by `SubtitleController`. In that case the export button doesn't show up in `SubtitlePlayer` when mirroring through the app. But yeah any thing that uses it under those conditions would trigger it.

common/components/DictionarySettingsTab.tsx
ShanaryS: I'm not sure if it would work without modification. This only triggers if certain fields were changed, ones that will cause an Anki cache rebuild such as the fields. These settings have a helper text below them when changed.

common/dictionary-db/dictionary-db.ts
ShanaryS: The only records that are deleted at this point are when the user has Anki disabled for this track (no fields). We clear the entries in this scenario and won't be using Anki for this track.

common/dictionary-db/dictionary-db.ts
ShanaryS: Yeah it's annoying and I tried a few different ways to write it but it was always confusing. So I just kept it as one check. We need `cardsInfo` as only cards are a part of decks in Anki but this query is about 10x slower `notesInfo`. It's written like this so we can avoid `cardsInfo` unless those cardIds were modified or a new track is enabled.

common/dictionary-db/dictionary-provider.ts
ShanaryS: It is used a lot. This is where everything interfaces with the db, much like `SettingsProvider`. It's used in settings, deleting profiles, Anki UI updates/exports, and `SubtitleColoring` for db queries and subscriptions. It will also be expanded for when local tokens are enabled.

common/hooks/use-settings-profile-context.ts
ShanaryS: If it fails then the entries will live in the db forever. I don't think we should allow the profile to be deleted if this fails (which should be impossible). It shouldn't take much time even with lots of entries.

extension/src/handlers/dictionary/dictionary-handler.ts
ShanaryS: Yes this comment is outdated but the principle is the same, we need to send it to the tab directly in some cases.

common/subtitle-coloring/subtitle-coloring.ts
ShanaryS: It's used in `SubtitleController` to check if we should have hover only for the html. `VideoPlayer` uses the boolean directly as we don't have access to the `SubtitleColoring` instance there.

common/subtitle-coloring/subtitle-coloring.ts
ShanaryS: We want to call it whenever Anki is first connectable during a mining session, this will keep the cache updated for users with no intervention. The second call is only when an Anki card was added/modified, we trigger a cache build. This will send what tokens were updated from this build through `addBuildAnkiCacheStateChangeCallback`.

We poll Anki every `10s` for the recently modified cards (in case the user adds without going through asbplayer) and all asbplayer exports/updates will trigger `addAnkiCardModifiedCallback`.

common/dictionary-db/dictionary-provider.ts
killergerbah: I see, `DictionaryProvider` didn't seem to add any behavior on top of `DictionaryStorage` but if you're planning to add something later then makes sense.

common/components/DictionarySettingsTab.tsx
```diff
@@ -130,15 +168,81 @@ const DictionarySettingsTab: React.FC<Props> = ({ settings, onSettingChanged, an
...
+            console.error('Failed to send build Anki cache message', e);
+            const message = e instanceof Error ? e.message : String(e);
+            setBuildAnkiCacheResult({ command, msg: message, error: true, modifiedTokens: [] });
+        } finally {
+            setBuildingAnkiCache(false);
+        }
+    }, [dictionaryProvider, settings, activeProfile]);
+
+    useEffect(() => {
+        const removeCB = dictionaryProvider.onBuildAnkiCacheStateChange((state) => {
```
killergerbah: Advantage of returning the unsubscription callback is that you can just return it in React `useEffect` functions like this
Not that it matters, just slightly more pleasant to read

killergerbah: IMO slightly more pleasant to read if you just return without the intermediate function variable and explicit invocation.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                t
+            );
+
+            const duration = Math.ceil((Date.now() - buildTs) / 1000);
+            msg = `${t('settings.dictionaryBuildAnkiTracks', {
+                tracks: Array.from(trackStates.keys())
+                    .map((track) => `#${track + 1}`)
+                    .join(', '),
+            })}`;
+            msg += ` | ${t('settings.dictionaryBuildModifiedCards', { numCards: numUpdatedCards.toLocaleString('en-US') })}`;
```
killergerbah: Can localization of the progress strings be confined to the dictionary settings form instead? This would require you to model the progress/errors via a data structure but I think this is probably not so difficult as there's a limited number of types of status updates.

While the refactor might be kind of annoying, keeping the front-end and logical concerns separate as a matter of principle will will give us more flexibility to render the progress in different ways on the settings form, without a dependency on this code.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+            for (const key of trackBuildIdsToClear) keysAndChanges.push({ key, changes: { lastBuildExpiresAt } });
+            await this.db.meta.bulkUpdate(keysAndChanges);
+        });
+        const time = new Date(Date.now() + eta).toLocaleTimeString([], {
+            hour: '2-digit',
+            minute: '2-digit',
+            second: '2-digit',
+        });
+        let etaSeconds = Math.ceil(eta / 1000);
+        const etaHours = Math.floor(etaSeconds / 3600);
```
killergerbah: `humanReadableTime` already does something like this

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                        .modify((record, ref) => {
+                            const remainingCardIds = record.cardIds.filter((id) => !cardIdsSet.has(id));
+                            if (!remainingCardIds.length) {
+                                modifiedTokens.add(record.token);
+                                for (const lemma of record.lemmas) modifiedTokens.add(lemma);
+                                delete (ref as any).value;
+                            } else if (remainingCardIds.length !== record.cardIds.length) {
+                                modifiedTokens.add(record.token);
+                                for (const lemma of record.lemmas) modifiedTokens.add(lemma);
+                                record.cardIds = remainingCardIds;
```
killergerbah, line 534: Minor nitpick - seems like you'd get more code re-use and readability if you only branch for the final modification to the record.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+        noteIds: number[]
+    ): Promise<Map<number, DictionaryAnkiCardRecord[]>> {
+        if (!noteIds.length) return new Map();
+        return this.db.ankiCards
+            .where('[profile+noteId]')
+            .anyOf(noteIds.map((noteId) => [profile, noteId]))
+            .toArray()
+            .then((ankiCards) => {
+                if (!ankiCards.length) return new Map();
+                const noteIdRecordMap = new Map<number, DictionaryAnkiCardRecord[]>();
```
killergerbah: Just a minor comment on variable naming - for me as a native English speaker it's easier to understand a variable name like `cardRecordsByNoteId` since it clearly specifies the value and the key in the name. It's also implied that the type is a map due to the use of the `by` preposition.

`noteIdRecordMap` map works too but it's not obvious that `noteId` is specifying the key since it's being used as an adjective which grammatically carries less information on its own.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                for (const ankiCard of ankiCards) {
+                    const val = noteIdRecordMap.get(ankiCard.noteId);
```
killergerbah: Another note on variable naming - in general if you can make the code easier to read by just choosing a different name then that's probably a good choice (IMO). For example I think `cardRecords` would be better, especially since `noteIdRecordMap` doesn't make it completely obvious what the value of the map is.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+            await this.db.meta.update(key, {
+                lastBuildStartedAt: buildTs,
+                lastBuildExpiresAt: initialExpiration,
+                buildId,
+            });
+            return true;
+        });
+    }
+
+    private async _buildIdHealthCheck(buildId: string, trackBuildIdsToClear: DictionaryMetaKey[]): Promise<void> {
```
killergerbah: Suggestion to use a different name than `trackBuildIdsToClear`. I see that build IDs do get cleared from the database in some places, but it makes less sense in the context of this method.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                        suspended: suspendedCards.has(cardId),
+                    });
+                }
+            }
+
+            for (const existingAnkiCards of existingAnkiNoteIdMap.values()) {
+                for (const ankiCard of existingAnkiCards) {
+                    if (!modifiedCardIdsSet.has(ankiCard.cardId)) continue;
+                    if (suspendedCards.has(ankiCard.cardId)) {
+                        if (!ankiCard.suspended) cardIdsToSuspend.push(ankiCard.cardId);
```
killergerbah: I guess this may be due to a quirk of Anki Connect's API but why does suspended state need to be dealt with separately? I see your comment above that `Reviews or suspension status are reflected in card mod time`, so why isn't suspension status already taken care of by `modifiedCards`?

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                            if (field) texts.push(field);
+                        }
+                    }
+                    await ts.yomitan.tokenizeBulk(texts);
+                }
+
+                const trackTokensMap = new Map<
+                    number,
+                    Map<DictionaryTokenSource, Map<string, { lemmas: string[]; cardIds: Set<number> }>>
+                >();
```
killergerbah, line 1203: Another naming nitpick - `partialTokenRecordsByTrack` feels like a nicer name. I think that would make it clearer that the map is being used to collect partial data that will be merged with the existing records.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                const ankiFieldsMap = new Map<number, Map<DictionaryTokenSource, string[]>>();
+                for (const [track, ts] of trackStates.entries()) {
+                    trackTokensMap.set(
+                        track,
+                        new Map([
+                            [DictionaryTokenSource.ANKI_WORD, new Map()],
+                            [DictionaryTokenSource.ANKI_SENTENCE, new Map()],
+                        ])
+                    );
+                    ankiFieldsMap.set(
```
killergerbah, line 1213: At the risk of introducing too much abstraction I think you can avoid intermediate data structures like this one by taking advantage of the fact that `TrackStatesForDB` is an internal type which can be implemented by a class. The class could provide short utility methods for things like this.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+        for (const key of trackBuildIdsToClear) {
+            try {
+                await this._clearBuildId(key, buildId);
+            } catch (e) {
+                console.error(`Error clearing Anki token build ID for track ${key[1] + 1}: ${e}`);
+            }
+        }
+    }
+
+    private async _syncTrackStatesWithAnki(
```
killergerbah, line 833: Suggestion to add a comment about what this method does. In particular, I think it's helpful to make it clear when arguments to a method are intended to be mutated, so that those arguments can be distinguished from the pure inputs. Same for other methods that mutate their arguments.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                    ) {
+                        return;
+                    }
+                    if (!trackStates.has(record.track)) return;
+                    if (trackTokensMap.get(record.track)!.get(record.source)!.has(record.token)) return; // We want tokens that were not updated but refers to updated cards (e.g. field value changed, different tokens)
+                    const validCardIds = record.cardIds.filter((id) => !modifiedCardsBatch.has(id));
+                    if (!validCardIds.length) {
+                        modifiedTokens.add(record.token);
+                        for (const lemma of record.lemmas) modifiedTokens.add(lemma);
+                        delete (ref as any).value;
```
killergerbah: I commented on a similar block of code somewhere else but if it's easy to do then I think you'll squeeze more readability out of the code by reducing duplication. For example, here the `validCardIds` checks could be used to branch on the modification to the record, without duplicating the updates to `modifiedTokens`.

common/dictionary-db/dictionary-db.ts
```diff
@@ -0,0 +1,1416 @@
...
+                    } else if (validCardIds.length !== record.cardIds.length) {
+                        modifiedTokens.add(record.token);
+                        for (const lemma of record.lemmas) modifiedTokens.add(lemma);
+                        record.cardIds = validCardIds;
+                    }
+                });
+        });
+    }
+
+    private async _gatherModifiedTokens(profile: string, modifiedTokens: Set<string>) {
```
killergerbah, line 1381: Suggestion to comment that `modifiedTokens` is being expanded by this method with existing entries in the database

common/dictionary-db/dictionary-db.ts
ShanaryS: That very flexibility was what I was trying to avoid. Some messages only makes sense to be displayed in certain states and constructing that seems like it would be complicated. Also some messages don't have translations if it's a random error. We also would need to handle translations in `SubtitleColoring` as it can log errors whenever tokens updates.

I still think it's worth doing though. I'm thinking of returning an object instead where the keys correspond to the translation key and values to the translation params. As well as a separate field for unspecified errors.

common/dictionary-db/dictionary-db.ts
ShanaryS: Under the current design it's no longer needed so I removed `_updateSuspendedCards()`. Before the change to allow partial builds it was necessary to update them separately but not anymore.

common/dictionary-db/dictionary-db.ts
ShanaryS: These utility methods would only be used in a single place with their structure determined from the context it's being used. I think it's clearer to only store data in `trackStates` and formatting it where it's needed. I think having the functions separated far from its use makes it harder to understand.

killergerbah: Okay

common/dictionary-db/dictionary-db.ts
killergerbah: If you provide both English string + the translation key, you could use the English string for logging in `SubtitleColoring`. The translation key could be used for display on the settings form.

killergerbah: But instead of translation key, maybe provide an an enum? That will give us some decoupling of the progress/error type from the  loc file.

ShanaryS: I can't get a design that I'm happy with so I think it's best if you make the changes.

killergerbah: Pushed, let me know if anything looks weird to you

ShanaryS: It looks good, the only issue I noticed is that the elapsed time changes when a build is done and the button is pressed. I think this lies in how the UI is calculating it but it's not a big deal.

I also re-review the rest of the PR as well. Earlier your comment about `dictionaryProvider` was accurate, we could use `DictionaryStorage` directly. I'm not sure if it's worth updating though. I could see it being useful for statistics.

killergerbah: > It looks good, the only issue I noticed is that the elapsed time changes when a build is done and the button is pressed. I think this lies in how the UI is calculating it but it's not a big deal.

I briefly tried this and couldn't repro something like what you're describing. All I did was finish a cache build, and then press the button again. But then it just shows you "Anki track(s): #1 | 0 modified card(s)
".

If there is issue that's easy to fix we can clean that up. I also want to note the following we should probably fix before release:
- Leaving the "annotations" tab and coming back will reset the button state, even if a build is in progress, until the next build update is received.
- The button is clickable even if card targeting is not completely set up (e.g. 0 word/sentence fields).
- We should probably add some more helper text to direct the user to setup stuff, if they enabled "colorize subtitles based on Anki maturity." Since some users might expect everything to work without further configuration.

ShanaryS: > All I did was finish a cache build, and then press the button again.

It's only when at least a card is built so it shows the elapsed time. The next time the button is pressed the elapsed counter is updated but this is fixed by setting the state to undefined on the button press.

> Leaving the "annotations" tab and coming back will reset the button state, even if a build is in progress, until the next build update is received.

Is there an easy fix for this? Otherwise it seems we would need to check the build on page load. But we won't get the updates in all scenarios, such as if they open the website in a new tab. If there isn't an easy fix it should be fine to leave as only the initial build should be long.

> The button is clickable even if card targeting is not completely set up (e.g. 0 word/sentence fields).
It can be disabled if no tracks are enabled, but empty fields are okay as that's what will clear the db on the next build. I also don't want anything too distracting if they don't configure it since it's valid to only use local tokens.

I'll address some of these in the local tokens PR and we can discuss more changes then.

killergerbah: >Is there an easy fix for this? Otherwise it seems we would need to check the build on page load.

Probably, just need to put the state somewhere outside the tab maybe

PR #853, opened 2026-01-09
feat(dictionary): support local and ignored tokens

### LINKED ISSUE #178: A function that can mark words we learnt
Could you develop a function that can mark words we learnt and when just have I+1 in sentence the movie will pause? I feel like when we have this function, the journey learning language is a game. I really like this function when I experience it in another project? I like Asbplayer 'cause it actives in so many websites. so that, I hope you will develop this function, it so usefull!!!
browser_pELpsrm2Wi

ShanaryS:
fixes #178

This allows collecting tokens locally without using Anki or to override statuses set by Anki.

* Priority for tokens is `Local > Anki word > Anki sentence`
* Local tokens are not per track unlike Anki (but still per profile), this makes it easier for users to switch between tracks and have local tokens behave consistently.
  * If per track local tokens are desired, it can be trivially added as an opt in later.
* Local tokens are mainly collected by hovering over the word and pressing the configured keyboard shortcut
  * Coloring must be enabled as it is used as the feedback for collection
  * Default keybinds are `Q+[0-5]` and `Q+I` for collecting and ignoring tokens
* Local tokens can also be collected using a clipboard import located in the `Annotations` tab
  * If users desire collecting without coloring or a token that was parsed differently they would use this feature
* Tokens can be exported and imported for backup, these only apply to local tokens as Anki tokens can always be regenerated
  * Imports are purely constructive, it will not delete existing tokens in the db and will use the highest status of existing or the import. The imported states are only used if there are currently no states on the token.
* Ignored tokens are treated as fully known, these are used for things such as names or places. Currently there is no value to ignored tokens but it will be useful when performing statistics.
  * If desired we can give ignored tokens their own color but at present I don't see a reason to
* The buildAnkiCache button is now disabled when all tracks are disabled with additional helper text if no Anki fields are configured
* `ensureStoragePersisted()` is used on any user interaction with this feature for additional assurance that the db will not be deleted.
* Treat `VideoPlayer` subtitles similar to `SubtitleController`, it will no longer split by line and instead use `white-space: pre-wrap`

In the future, it would be beneficial to have a way to browse the local words by their token/lemmas, states, status, etc. This would also be a good area to delete tokens, currently users must hover and mark uncollected or use the clipboard import with uncollected.

### DIFF
--- a/common/dictionary-db/dictionary-db.ts
+++ b/common/dictionary-db/dictionary-db.ts
@@ -135,13 +135,34 @@ export interface LemmaResults {
     }[];
 }
 
+export interface DictionarySaveRecordLocalResult {
+    savedTokens: DictionaryTokenKey[];
+    deletedTokens: DictionaryTokenKey[];
+}
+
+export interface DictionaryImportRecordLocalResult {
+    importedTokens: DictionaryTokenKey[];
+}
+
+export interface DictionaryExportRecordLocalResult {
+    exportedRecords: Partial<DictionaryTokenRecord>[];
+}
+
+export interface DictionaryDeleteRecordLocalResult {
+    deletedTokens: DictionaryTokenKey[];
+}
+
+export interface DictionaryDeleteProfileResult {
+    deletedMetas: DictionaryMetaKey[];
+    deletedTokens: DictionaryTokenKey[];
+    deletedAnkiCards: DictionaryAnkiCardKey[];
+}
+
 interface BuildAnkiCacheProgress {
     current: number;
     total: number;
     startedAt: number;
 }
-
-const command = 'dictionary-build-anki-cache-state';
 
 function hasDeck(dt: DictionaryTrack, cardDeck: string): boolean {
     if (!dt.dictionaryAnkiDecks.length) return true;
@@ -372,8 +393,8 @@ export class DictionaryDB {
         inputProfile: string | undefined,
         localTokenInputs: DictionaryLocalTokenInput[],
         applyStates: ApplyStrategy
-    ): Promise<[DictionaryTokenKey[], number]> {
-        if (!localTokenInputs.length) return [[], 0];
+    ): Promise<DictionarySaveRecordLocalResult> {
+        if (!localTokenInputs.length) return { savedTokens: [], deletedTokens: [] };
         const profile = this._getProfile(inputProfile);
         return this.db.transaction('rw', this.db.tokens, async () => {
             const tokenRecordMap = await this._getFromSourceBulk(
@@ -388,7 +409,7 @@ export class DictionaryDB {
             for (const localTokenInput of localTokenInputs) {
                 if (!HAS_LETTER_REGEX.test(localTokenInput.token)) {
                     console.error(`Cannot save local token with invalid token: "${localTokenInput.token}"`);
-                    return [[], 0];
+                    return { savedTokens: [], deletedTokens: [] };
                 }
                 const existingRecord = tokenRecordMap.get(localTokenInput.token);
                 if (existingRecord) {
@@ -403,10 +424,9 @@ export class DictionaryDB {
                             }
                             break;
                         case ApplyStrategy.REMOVE:
-                            for (const state of existingRecord.states) {
-                                const idx = localTokenInput.states.indexOf(state);
-                                if (idx !== -1) localTokenInput.states.splice(idx, 1);
-                            }
+                            localTokenInput.states = existingRecord.states.filter(
+                                (existingState) => !localTokenInput.states.includes(existingState)
+                            );
                             break;
                         case ApplyStrategy.REPLACE:
                             break;
@@ -422,7 +442,7 @@ export class DictionaryDB {
                             break;
                         default:
                             console.error(`Unsupported applyStates value: "${applyStates}"`);
-                            return [[], 0];
+                            return { savedTokens: [], deletedTokens: [] };
                     }
                 } else if (localTokenInput.status == null) {
                     localTokenInput.status = TokenStatus.UNCOLLECTED;
@@ -430,7 +450,7 @@ export class DictionaryDB {
                 localTokenInput.lemmas = localTokenInput.lemmas.filter((lemma) => HAS_LETTER_REGEX.test(lemma));
                 if (!localTokenInput.lemmas.length) {
                     console.error(`Cannot save local token with no lemmas: "${localTokenInput.token}"`);
-                    return [[], 0];
+                    return { savedTokens: [], deletedTokens: [] };
                 }
                 if (localTokenInput.status === TokenStatus.UNCOLLECTED && !localTokenInput.states.length) {
                     if (existingRecord) {
@@ -440,7 +460,7 @@ export class DictionaryDB {
                         console.error(
                             `Cannot save local token with uncollected status and no states: "${localTokenInput.token}"`
                         );
-                        return [[], 0];
+                        return { savedTokens: [], deletedTokens: [] };
                     }
                 }
                 recordsToAdd.push({
@@ -454,45 +474,53 @@ export class DictionaryDB {
                     cardIds: [],
                 });
             }
-            return Promise.all([
+            const res = await Promise.all([
                 this._saveRecordBulk(recordsToAdd),
                 this.deleteRecordLocalBulk(inputProfile, tokensToDelete),
             ]);
+            return { savedTokens: res[0], deletedTokens: res[1].deletedTokens };
         });
     }
 
-    async deleteRecordLocalBulk(inputProfile: string | undefined, tokens: string[]): Promise<number> {
-        if (!tokens.length) return 0;
+    async deleteRecordLocalBulk(
+        inputProfile: string | undefined,
+        tokens: string[]
+    ): Promise<DictionaryDeleteRecordLocalResult> {
+        if (!tokens.length) return { deletedTokens: [] };
         const profile = this._getProfile(inputProfile);
-        return this.db.tokens
-            .where('[token+source+track+profile]')
-            .anyOf(tokens.map((token) => [token, DictionaryTokenSource.LOCAL, LOCAL_TOKEN_TRACK, profile]))
-            .delete();
+        return this.db.transaction('rw', this.db.tokens, async () => {
+            const deletedTokens = await this.db.tokens
+                .where('[token+source+track+profile]')
+                .anyOf(tokens.map((token) => [token, DictionaryTokenSource.LOCAL, LOCAL_TOKEN_TRACK, profile]))
+                .primaryKeys();
+            await this.db.tokens.bulkDelete(deletedTokens);
+            return { deletedTokens };
+        });
     }
 
     /**
      * The only export we need is local tokens, Anki can be rebuilt as needed.
      * Since our needs are simple, we can avoid using the dexie-export-import package.
      */
-    async exportRecordLocalBulk(): Promise<Partial<DictionaryTokenRecord>[]> {
+    async exportRecordLocalBulk(): Promise<DictionaryExportRecordLocalResult> {
         return this.db.tokens
             .filter((record) => record.source === DictionaryTokenSource.LOCAL)
             .toArray()
-            .then((records) =>
-                records.map((r) => ({
+            .then((records) => ({
+                exportedRecords: records.map((r) => ({
                     profile: r.profile,
                     token: r.token,
                     status: r.status,
                     lemmas: r.lemmas.length ? r.lemmas : undefined,
                     states: r.states.length ? r.states : undefined,
-                }))
-            );
+                })),
+            }));
     }
 
     async importRecordLocalBulk(
         items: Partial<DictionaryTokenRecord>[],
         profiles: string[]
-    ): Promise<DictionaryTokenKey[]> {
+    ): Promise<DictionaryImportRecordLocalResult> {
         const defaultProfile = this._getProfile(undefined);
         if (!profiles.includes(defaultProfile)) profiles.unshift(defaultProfile);
         const fullyKnownStatus = getFullyKnownTokenStatus();
@@ -527,7 +555,7 @@ export class DictionaryDB {
                     for (const existingLemma of existingToken.lemmas) {
                         if (!item.lemmas.includes(existingLemma)) item.lemmas.push(existingLemma);
                     }
-                    if (existingToken.states.length) item.states = existingToken.states; // Treat the existing states as authoritative if any is present
+                    item.states = existingToken.states; // Treat the existing states as authoritative, TODO: expose ApplyStrategy for imports?
                 }
                 item.lemmas = item.lemmas.filter((lemma) => HAS_LETTER_REGEX.test(lemma));
                 if (!item.lemmas.length) continue; // Cannot import tokens with no lemmas, require a different method where a tokenizer is available
@@ -549,18 +577,22 @@ export class DictionaryDB {
                     cardIds: [],
                 });
             }
-            return this._saveRecordBulk(records);
+            return { importedTokens: await this._saveRecordBulk(records) };
         });
     }
 
-    async deleteProfile(profile: string): Promise<[number, number, number]> {
-        return this.db.transaction('rw', this.db.meta, this.db.tokens, this.db.ankiCards, () =>
-            Promise.all([
-                this.db.meta.where('profile').equals(profile).delete(),
-                this.db.tokens.where('profile').equals(profile).delete(),
-                this.db.ankiCards.where('profile').equals(profile).delete(),
-            ])
-        );
+    async deleteProfile(profile: string): Promise<DictionaryDeleteProfileResult> {
+        return this.db.transaction('rw', this.db.meta, this.db.tokens, this.db.ankiCards, async () => {
+            const deletedMetas = await this.db.meta.where('profile').equals(profile).primaryKeys();
+            const deletedTokens = await this.db.tokens.where('profile').equals(profile).primaryKeys();
+            const deletedAnkiCards = await this.db.ankiCards.where('profile').equals(profile).primaryKeys();
+            await Promise.all([
+                this.db.meta.bulkDelete(deletedMetas),
+                this.db.tokens.bulkDelete(deletedTokens),
+                this.db.ankiCards.bulkDelete(deletedAnkiCards),
+            ]);
+            return { deletedMetas, deletedTokens, deletedAnkiCards };
+        });
     }
 
     private async _getFromSourceBulk(
@@ -582,8 +614,11 @@ export class DictionaryDB {
             });
     }
 
-    private async _getAnkiCards(profile: string): Promise<DictionaryAnkiCardRecord[]> {
-        return this.db.ankiCards.where('profile').equals(profile).toArray();
+    /**
+     * primaryKeys() and keys() are faster than toArray(), we don't need all fields
+     */
+    private async _getAnkiCardKeys(profile: string): Promise<DictionaryAnkiCardKey[]> {
+        return this.db.ankiCards.where('profile').equals(profile).primaryKeys();
     }
 
     private async _getAnkiCardsByNoteIdBulk(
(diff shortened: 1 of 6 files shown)


common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -937,4 +990,53 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
+    }
+
+    static handleMouseOut(mouseEvent: MouseEvent, hoveredElement: HTMLElement | null): boolean {
+        if (!(mouseEvent.target instanceof HTMLElement)) return true;
+        return hoveredElement === mouseEvent.target;
+    }
+
+    static parseTokenFromElement(tokenEl: HTMLElement | null): { token: string; track: number } | null {
+        if (!tokenEl) return null;
+        if (tokenEl.tagName === 'RUBY') {
```
killergerbah: This could be simplified by using the top of this method to determine what `tokenEl` should be, and then checking the class list. This would de-duplicate the class list check for `ASB_TOKEN_CLASS`.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -937,4 +990,53 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
+        if (!(mouseEvent.target instanceof HTMLElement)) return null;
+        return mouseEvent.target;
+    }
+
+    static handleMouseOut(mouseEvent: MouseEvent, hoveredElement: HTMLElement | null): boolean {
+        if (!(mouseEvent.target instanceof HTMLElement)) return true;
+        return hoveredElement === mouseEvent.target;
+    }
+
+    static parseTokenFromElement(tokenEl: HTMLElement | null): { token: string; track: number } | null {
```
killergerbah: I wonder if all of these `static` functions + `_applyTokenStyle` should be in their own file, as they implement the rendering functionality for the subtitle coloring. Then you could export functions instead of using public static methods on this class.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -937,4 +990,53 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
+    static _extractTokenFromNode(node: Node): string {
+        if (node.nodeType === Node.TEXT_NODE) return node.textContent ?? '';
+        if (node.nodeType !== Node.ELEMENT_NODE) return '';
+
+        let token = '';
+        const el = node as HTMLElement;
+        if (el.tagName === 'RUBY') {
+            for (const child of el.childNodes) {
+                if (child.nodeType === Node.ELEMENT_NODE && (child as HTMLElement).tagName === 'RT') continue;
+                token += this._extractTokenFromNode(child);
```
killergerbah, line 1041: Is there a reason you're not early exiting if `this._extractTokenFromNode(child)` is nonempty?

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -937,4 +990,53 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
             this.subtitlesInterval = undefined;
         }
     }
+
+    static handleMouseOver(mouseEvent: MouseEvent): HTMLElement | null {
```
killergerbah: It took me a little while to understand why these static functions are grouped together in `SubtitleColoring`. Would it make sense to have a separate class for implementing this hover detection + token extraction logic? The state of the class should be the hovered element, and the behavior would be to extract the token? I think this would make the code easier to understand, and also relieve callers of the responsibility to track hovered element state themselves.

killergerbah:
Btw one thing I noticed when testing this, is that the default underline style makes it hard to tell the boundaries between consecutively underlined tokens. With this feature, I think it becomes more important to know exactly which token will be saved when the user uses the shortcut. Solvable with a highlight effect? Or spacing between consecutively underlined tokens?

common/subtitle-coloring/subtitle-coloring.ts
ShanaryS: I think it's necessary for ruby-normal-ruby tokens but I'm not sure. It should work for all parsing, is there a case it doesn't handle?

ShanaryS:
I added a highlight for the token but it doesn't seem to work for SidePanel. Also as a general note, the keybinds only work when the player is in focus. I think this is desirable but can be unintuitive if the user is multi tasking. This highlight effect is always there though when hovered, even if the player is not in focus.

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: Ok if there's a reason it might be necessary than let's keep it. Was just wondering since it seemed like you could stop recursion early.

killergerbah:
Hmm maybe highlight is not ideal then. Could just adjust the styles so that they have boundaries between - at least those styles are connected to the settings already.

common/dictionary-db/dictionary-provider.ts
```diff
@@ -1,21 +1,29 @@
...
 export interface DictionaryStorage {
     getBulk: (profile: string | undefined, track: number, tokens: string[]) => Promise<TokenResults>;
     getByLemmaBulk: (profile: string | undefined, track: number, lemmas: string[]) => Promise<LemmaResults>;
     saveRecordLocalBulk: (
         profile: string | undefined,
-        localTokenInputs: DictionaryLocalTokenInput[]
-    ) => Promise<DictionaryTokenKey[]>;
+        localTokenInputs: DictionaryLocalTokenInput[],
+        applyStates: ApplyStrategy
+    ) => Promise<[DictionaryTokenKey[], number]>;
```
killergerbah: I don't see this return value being used right now, but since you're returning something and sending it from extension to foreground scripts, can this be modeled using an object instead? As we can't easily change this return value, and an object can be extended later in a readable way.

common/dictionary-db/dictionary-provider.ts
```diff
@@ -1,21 +1,29 @@
...
     deleteRecordLocalBulk: (profile: string | undefined, tokens: string[]) => Promise<number>;
     deleteProfile: (profile: string) => Promise<[number, number, number]>;
+    exportRecordLocalBulk: () => Promise<Partial<DictionaryTokenRecord>[]>;
```
killergerbah: Similarly here and for `importRecordLocalBulk` below, while the return value is not likely to change, I think an object might be preferred, and its low cost to change that now.

common/settings/settings.ts
```diff
@@ -89,13 +89,24 @@ export enum TokenStatus {
...
-    MATURE = 5, // If ever adding more statuses, they should go after MATURE and getFullyKnownTokenStatus should be updated
+    MATURE = 5, // If ever adding more statuses, they should go last and getFullyKnownTokenStatus should be updated
 }
 
 export function getFullyKnownTokenStatus(): TokenStatus {
     return TokenStatus.MATURE; // If future statuses are optional, this logic may need to change
 }
 
+export enum TokenState {
+    IGNORED = 0, // If ever adding more states, they should go last (if adding colors for states, use a separate array from tokenStatusColors indexed by TokenState)
```
killergerbah, line 100: Is it accurate to say that "ignored" is effectively an alias for "fully known?" Asking just for my own information - I guess this is useful if we want to treat "ignored" differently from "fully known" in the future.

common/settings/settings.ts
```diff
@@ -89,13 +89,24 @@ export enum TokenStatus {
...
+}
+
+export enum ApplyStrategy {
```
killergerbah, line 103: If the user accidentally saves a local token, is it not possible to undo that operation right now? I noticed `REMOVE` is not currently used outside of the dictionary DB. If that's true then maybe we can solve it with a token browser UI. I think you mentioned that at some point already.

common/dictionary-db/dictionary-db.ts
```diff
@@ -387,26 +383,67 @@ export class DictionaryDB {
...
+                        if (!localTokenInput.lemmas.includes(existingLemma)) localTokenInput.lemmas.push(existingLemma);
                     }
+                    switch (applyStates) {
+                        case ApplyStrategy.ADD:
+                            for (const state of existingRecord.states) {
+                                if (!localTokenInput.states.includes(state)) localTokenInput.states.push(state);
+                            }
+                            break;
+                        case ApplyStrategy.REMOVE:
+                            for (const state of existingRecord.states) {
```
killergerbah: When the apply strategy is "remove," the input state is only removed if it's already on the existing record? So if the existing record doesn't have the state already, the state would be saved. Is that intended?

common/settings/settings.ts
killergerbah: Oh I see you wrote that in the PR description already, lol

common/dictionary-db/dictionary-db.ts
ShanaryS: It's not, I had the variables flipped. I'm using filter now which should be clearer.

common/settings/settings.ts
ShanaryS: Yes, they are completely equivalent currently. Their use will become important for statistics, so users can ignore places and names in the known words calculation and perhaps other usages. I opted not to give these their own customizable color due to this but we can add support for it later (and any other states that we add).

I designed the states using an array on the tokens even some may be mutually exclusive. Handling those complexities is left to the code to be flexible

common/dictionary-db/dictionary-db.ts
ShanaryS: It's not, I had the variables flipped. I'm using filter now which should be clearer.

common/settings/settings.ts
ShanaryS: Yes, they are completely equivalent currently. Their use will become important for statistics, so users can ignore places and names in the known words calculation and perhaps other usages. I opted not to give these their own customizable color due to this but we can add support for it later (and any other states that we add).

I designed the states using an array on the tokens even some may be mutually exclusive. Handling those complexities is left to the code to be flexible.

common/settings/settings.ts
ShanaryS: Yeah, the plan for managing tokens is essentially exposing the db to the users so they can read and update it directly (where applicable). So they type in a string and we search for tokens and lemmas with that substring and display results. It could be organized by source where local tokens can be deleted or have applicable fields editable, but it would also expose what tokens were processed from Anki. Essentially think of the Anki browser but for the dictionary db.

This is fairly complex (on the UI/UX side) and doesn't offer any real functionality. Uncollecting using hover and clipboard should be fine until we get around to that.

ShanaryS:
I updated all the functions to return objects and also am returning the keys instead of just the count for deletions. This gives more flexibility for future use.

I also changed how existingTokens are handled when importing. Now the existing states will always be preserved exactly as I think this makes the most sense. If the token already exists, then it should be safe to assume that's the state the user wants it in if they recently collected it. Users may end up requesting more control over the import logic, such as a destructive replace or exposing `ApplyStrategy`. But for now I think using the highest status and existing states is the best default.

> Hmm maybe highlight is not ideal then. Could just adjust the styles so that they have boundaries between - at least those styles are connected to the settings already.

I removed the highlight on hover for now, I think it maybe possible to detect if the element is in focus and only displaying it then, we can't just use the window focus as the `SubtitlePlayer` and `Video` have separate focuses for the keybinds. In general I don't think it will be a big deal as users should know enough to know what word they are collecting, and we can assume that adjacent words are unlikely to be the same status once users have settled in.

common/components/DictionarySettingsTab.tsx
```diff
@@ -267,12 +272,41 @@ const DictionarySettingsTab: React.FC<Props> = ({
         yomitanSectionRef.current?.scrollIntoView();
     };
 
+    const dictionaryDBFileInputRef = useRef<HTMLInputElement>(null);
+    const handleDictionaryDBFileInputChange = useCallback(async () => {
+        try {
+            const file = dictionaryDBFileInputRef.current?.files?.[0];
+            if (file === undefined) return;
+            await dictionaryProvider.importRecordLocalBulk(
+                JSON.parse(await file.text()),
```
killergerbah, line 281: Should we allow more flexibiity with the file import? As it is, the user would have to be aware of the JSON schema for the token records, and also pre-lemmatize everything. Meanwhile, import via clipboard uses one token per line. I think users would expect both import from file and import from clipboard to use the same data format. For file import, we _could_ support both - try json parse first then try split by newline. 

If we do this, it would make sense to merge the current file import with the clipboard import dialog, as we would want to re-use the lemmatization logic etc.

I can take on this change in a different PR if you agree.

common/components/DictionarySettingsTab.tsx
```diff
@@ -293,817 +327,869 @@ const DictionarySettingsTab: React.FC<Props> = ({
...
+                            i18nKey={`${dictionaryYomitanUrlError ? t('settings.annotationHelperText') : ''}${!extensionInstalled ? ` ${t('settings.annotationNoExtensionWarn')}` : ''}`.trim()}
+                            components={[
+                                <Link key={0} onClick={handleYomitanHelperTextClicked} href="javascript:void(0)" />,
+                            ]}
+                        />
+                    </Alert>
+                )}
+                <div>
+                    <Stack direction="row" spacing={1} alignItems="center">
+                        <DictionaryClipboardImport
```
killergerbah: Not for this PR, but I'm thinking it would be simpler UX to reduce the buttons to "import" and "export," where "import" opens a dialog with all the import options available. "Import from file" is just another button, so we could put that in the current `DictionaryClipboardImport` dialog. I can take on this change if you agree.

common/dictionary-db/dictionary-db.ts
```diff
@@ -417,54 +474,151 @@ export class DictionaryDB {
...
+                        if (!item.lemmas.includes(existingLemma)) item.lemmas.push(existingLemma);
+                    }
+                    item.states = existingToken.states; // Treat the existing states as authoritative, TODO: expose ApplyStrategy for imports?
+                }
+                item.lemmas = item.lemmas.filter((lemma) => HAS_LETTER_REGEX.test(lemma));
+                if (!item.lemmas.length) continue; // Cannot import tokens with no lemmas, require a different method where a tokenizer is available
+                let status = item.status;
+                if (item.status == null || item.status < TokenStatus.UNKNOWN) {
+                    if (!item.states.length) continue; // Status cannot be uncollected unless there is a state
+                    if (status == null) status = TokenStatus.UNCOLLECTED;
```
killergerbah, line 565: `status` could be `undefined` as well?

common/settings/settings.ts
killergerbah: Makes sense, can be for later

common/components/DictionarySettingsTab.tsx
ShanaryS: Yeah I think that makes sense, the original goal of the export/import is to be handled by asbplayer only and users would be able to covert to it with the exception of the lemma. It think giving the option to parse the lemmas make sense.

I want to clarify, the clipboard import does not split by newline, it parses whatever arbitrary text is there. But the user can dictate words by adding newlines so Yomitan won't see it as a word, we send the text as is. This means though that long text will be slow since we won't be batching though we could by splitting the newlines ourselves. I don't expect users to to paste more than say 100 words though.

common/dictionary-db/dictionary-db.ts
ShanaryS: It can't but if it was should be treated the same.

NovaKing001:
@ShanaryS Hey, I've been using this PR for the past week and have a few ideas/suggestions. 
### Add more shortcuts
Add shortcuts to set the status of all uncollected words in a subtitle line to either 'unknown' or 'known' this would help speed up the process of adding words.
### Treat ignore as its own status
This would allow for ignored words to have word readings, useful for languages such as Japanese, where proper names have ambiguous readings.  I have a names dictionary installed on yomitan, and having their readings show up would be a plus.
### Clipboard issues
I know that you said that you don't expect users to import more than 100 words ... I tried to import 10k known words, and it wouldn't successfully add them. It would preview for a few minutes and allow me to save; however, saving didn't do anything, and no words were actually added. I have to rely on having an Anki deck with my known words marked as suspended to pseudo-import my words. 
### Highlighting
The highlighting feature was great for identifying false positives. Making it a toggle in settings would be great. The more dictionaries you have installed in Yomitan, the weirder the parses work. 
For example, imagine trying to change the status of チャンク, but instead it changes 2チャン.  Highlighting would help identify these Yomitan false positives.
image
image
### Some characters are not able to have their status changed
Some characters, such as small よ　(ょ), are tokenized but are unable to have their statuses changed 
image 
In the picture above, all of the small characters aren’t able to be changed. 
### Future feature?
To mitigate false positives, you can manually select the characters and have them be tokenized/corrected 
for example 
from 
image
to 
image
This is currently how it works in [Lute](https://github.com/LuteOrg/lute-v3)
[Screencast_20260118_235313.webm](https://github.com/user-attachments/assets/02c2faf3-7df5-4f82-a714-4f75168b0435)
It would probably work by segmenting those characters from the subtitle line and sending them separately to yomitan however I don’t know how that would work with every instance of those characters. I dont know if it would be feasible/ worthwhile. Probably best to wait for this pr https://github.com/yomidevs/yomitan/pull/2254 to help with Yomitan’s false positives/ inaccurate parsing. 


Thank you for your hard work! I probably missed a few things, let me know what you think. Thanks!

ShanaryS:
@NovaKing001

> Add more shortcuts

This seems pretty niche and only gets less relevant as more words are marked.

> Treat ignore as its own status

From a technical point and don't think it's a good strategy, these flags on the cards should be independent of status. But we can add a toggle to show readings for ignored words.

> Clipboard issues

Yeah it's not meant for that big of an import, you use the import words button. It's meant for single words or sentences, maybe a paragraph. If you split it up yourself you be able to use it for that. But I find it hard to believe you have 10k words you can copy and they are all the same state. It sounds like you are using Anki for these words, so why do you need to import that many?

Also what is the structure of the text you are pasting? Does it have punctuation, spaces, or line breaks?

> Highlighting

Yeah this can be brought back in some way but some parts of the existing version was unintuitive.

> Some characters are not able to have their status changed

These all work fine for me and shouldn't be handled any different, perhaps you are on an outdated branch. The Yomitan parsing for this is incorrect though.

> To mitigate false positives, you can manually select the characters and have them be tokenized/corrected

I'm not sure how much value this would bring, if you know the tokenization is incorrect there is no need to mark it. We could only ever apply the override to that specific subtitle event, we won't be allow those characters to say always be parsed as a word. Language is too context sensitive for us to make any parsing decisions on our side so we will just have to rely on better parsers if this algorithm is insufficient.

NovaKing001:
@ShanaryS Thank you for responding 

> This seems pretty niche and only gets less relevant as more words are marked.

I have over 10,000 words marked, and I still find it tedious trying to mark all the false positives I'm getting. I think I just worded this wrong. What I really meant was adding some shortcuts related to bulk editing. I would hardly call this niche as [Migaku ](https://migaku.com/), [Lute](https://github.com/LuteOrg/lute-v3), and [Lingq](https://www.lingq.com/en/?referral=chloenguyen&gad_campaignid=23435229531) all have some features related to bulk editing on the spot without having to exit and access a separate word list. Maybe a shortcut that could highlight all uncollected words, and the user could then choose a status using their status shortcuts.

The reason I'm asking for these bulk shortcuts is that when someone inevitably adds dynamic autopausing based on word statuses, it'll autopause on known words that are incorrectly parsed as separate unknown words. It will improve the workflow to quickly bulk edit words without having to hover over each word. Of course, this is all my opinion as someone who has used similar tools in the past; saving a couple of seconds really does improve the user experience. 

> Yeah it's not meant for that big of an import, you use the import words button. It's meant for single words or sentences, maybe a paragraph. If you split it up yourself you be able to use it for that. But I find it hard to believe you have 10k words you can copy and they are all the same state. It sounds like you are using Anki for these words, so why do you need to import that many?

The words I'm importing are all in newlines, just words, no sentences. As for the import button, I noticed that it's for JSON files. Is it able to accept CSV/txt files? I tried to import a txt file before, and nothing seemed to happen. All of my known words come from LUTE, which can either export as a txt or a csv. I imported these words to Anki to be able to use them in conjunction with asbplayer. 

As for the weird Yomitan parsing, I have no clue why this is happening; it might be due to the number of dictionaries I have installed. I can send you an srt file and cross-check the parsing. 

Once again, thank you for adding the Yomitan API. It has greatly improved my workflow and opened up possibilities for many additional features I'd like to discuss in the near future.

ShanaryS:
> it'll autopause on known words that are incorrectly parsed as separate unknown words

Solving this is currently outside of the design scope. We rely completely on Yomitan parsing and overriding them is not currently planned. That's why even if you notice something off and manually try to fix it, it order for it to apply for the next subtitle or next session would require us making decisions about language parsing. The solution is to find alternatives to the standard Yomitan parser (there is a PR for exposing mecab) that's more accurate.

If you are fine with marking the incorrectly parsed words as known then it's fine. But bulk marking like this won't be a high priority as there is more pressing things to focus on.

> Is it able to accept CSV/txt files?

If it needs to be parsed a specific way then that's probably not going to be high priority. If it's just a text file as you say with words (either manually separated or just sentences) then that will be supported. #857 should help with that.

PR #857, opened 2026-01-22
Rearrange annotation settings

killergerbah:
Rearranged annotation settings mostly as discussed:
- Moved "import from file" into the "import clipboard" dialog and renamed the dialog.
- Rearranged the import/export and Anki cache sections so that they're below the "Annotation" section. The "colorize" setting felt more important, since it enables the word DB settings .
- Renamed them to "local word database" and "Anki word database" respectively. "Cache" felt like an implementation detail that might not be palatable to layman users.
- Changed wording of some helper text to (hopefully) be more clear.
- Linked some helper text to the settings sections they refer to.

https://github.com/user-attachments/assets/798de5f2-1b75-4975-8ae8-5aa828e27429

### DIFF
--- a/common/components/SettingsForm.tsx
+++ b/common/components/SettingsForm.tsx
@@ -367,7 +367,10 @@ export default function SettingsForm({
                     onSettingChanged={handleSettingChanged}
                     onViewKeyboardShortcuts={() => {
                         setTabIndex(tabIndicesById['keyboard-shortcuts']);
-                        keyboardShortcutsPanelRef.current?.scrollBy(0, 10000);
+                        setTimeout(
+                            () => keyboardShortcutsPanelRef.current?.scrollBy({ top: 10000, behavior: 'smooth' }),
+                            0
+                        );
                     }}
                 />
             </TabPanel>


killergerbah:
> the only reason I kept the Anki button at the top is to make it clear it applies to all tracks at once but it shouldn't be pressed often anyways.

Oh.. forgot about that. I'll move the buttons back up. Should be easy enough for the user should figure out that they need to enable stuff for the buttons to work.

> I didn't see any changes to have the lemmas from the json calculated if missing, is that for a later PR?

I also didn't know that we wanted to do this.. if importing JSON, then most likely the words were exported with lemmas in the first place, is this something we expect to happen often?

ShanaryS:
> I also didn't know that we wanted to do this.. if importing JSON, then most likely the words were exported with lemmas in the first place, is this something we expect to happen often?

I was referring to this comment: https://github.com/killergerbah/asbplayer/pull/853#discussion_r2701848121

By combing the clipboard and file import we can now parse any missing lemmas since the user can chose the track to parse with.

killergerbah:
@ShanaryS I've added the lemmatization but it will be done regardless of whether the record already has lemmas or not - I opted against adding an option for now since I thought it might be confusing given that the dialog already has a Yomitan API URL input, implying that parsing will be done.

Maybe we could add a checkbox to the effect of 'lemmatize only if record has no lemmas,' but this starts getting confusing since the behavior would only be affected in the case of "import from file" and not "import from clipboard." I guess this is the tradeoff with using the same dialog for two slightly different behaviors.

killergerbah:
I also updated the dialog UI slightly:
image

image

common/components/DictionaryImport.tsx
```diff
@@ -0,0 +1,372 @@
...
+        void ensureStoragePersisted();
+        setImportClipboardLoading(true);
+        setImportClipboardError(undefined);
+        setImportClipboardPreview(null);
+        try {
+            const dt = dictionaryTracks[importClipboardTrack];
+            const yomitan = new Yomitan(dt);
+
+            const tokenSet = new Set<string>();
+            for (const tokenParts of await yomitan.tokenize(importClipboardText)) {
```
ShanaryS: I was planning on adding this in a PR but we might as well do it here. It would be beneficial to split this text so we can take advantage of the concurrent tokenization. I think this regex should do to split the text: `/(?:\p{STerm}|\r?\n)+/u`

It splits by newlines and sentence terminating punctuation, other symbols are too messy and could be part of words in some languages. So this covers `.`, `!`, `?`, or even `。` for Japanese.

You can just pass in the the array to `yomitan.tokenizeBulk` then call `yomitan.tokenize` on each element of the array.

common/components/DictionaryImport.tsx
```diff
@@ -0,0 +1,372 @@
...
+            try {
+                text = await file.text();
+            } catch (e) {
+                console.error(e);
+                setImportClipboardError(e instanceof Error ? e.message : String(e));
+                return;
+            }
+
+            // Assume exported JSON file in DictionaryTokenRecord format first
+            let records: DictionaryTokenRecord[];
```
ShanaryS, line 203: This should be `Partial<DictionaryTokenRecord>[]`, export uses the interface:
```ts
export interface DictionaryExportRecordLocalResult {
    exportedRecords: Partial<DictionaryTokenRecord>[];
}
```

This is because we only need:
```
token
status
lemmas // Will now parse if empty or missing
states
```

common/components/DictionaryImport.tsx
```diff
@@ -0,0 +1,372 @@
...
+        async (records: DictionaryTokenRecord[]) => {
+            for (const record of records) {
+                const dt = dictionaryTracks[record.track ?? importClipboardTrack];
+                const yomitan = new Yomitan(dt);
+                record.lemmas = await yomitan.lemmatize(record.token);
+            }
+            return records;
+        },
+        [dictionaryTracks, importClipboardTrack]
+    );
```
ShanaryS, line 188: This should take in `records: Partial<DictionaryTokenRecord>[]` as the argument then check for `record.lemmas?.length` on each record. We should only need to lemmatize if this field is missing or empty, Yomitan doesn't cache the lemmatize call so it will be needlessly slow for asbplayer-export to asbplayer-import.

It also may be more efficient to create the Yomitan class once and call `yomitan.resetCache()` every 1000 iterations or so, or even not at all since the cache for lemmatize is small. Though I don't really expect the cache to be used since we are de-duplicating.

PR #866, opened 2026-01-26
Highlight hovered token if collectable

ShanaryS:
This lets the user know which token will be collected if the press a keybind. The highlight only shows when focus is correctly applied to the Video/SubtitlePlayer, only one of `VideoPlayer` or `SubtitlePlayer` for the `App` can receive the keybinds and likewise one of `Video` or `SidePanel` for the extension. Essentially, if the highlight is not present, the keybinds won't work.

This is a pure css implementation using `:hover` and `:focus-within`. To get the focus working needed `.asbplayer-token-container` applied in strategic places along with `tabIndex=-1` to allow them to be focusable. I needed to apply this to the `body` of the page when dealing with the subtitle overlay for arbitrary pages with the extension, this might have unintended effects.

Currently the color is a light blue, lighter than selecting text with a mouse but similar enough to be intuitive. This can be changed if desired.

The only issue I noticed is that clicking the title bar when using the `App` allows collecting tokens on the `SubtitlePlayer` but it won't be highlighted. This is a small issue though and we should prefer false negatives over false positives for highlighting for this feature. The other concern is that users may think they can collect custom tokens by manually highlighting with the mouse.


common/app/components/video-player.css
```diff
@@ -40,3 +40,7 @@
 .asbplayer-subtitles:hover .asbplayer-subtitle-rich {
     display: inline;
 }
+
+.asbplayer-token-container:focus-within .asb-token:hover {
```
killergerbah: Maybe we should rename this file, since its no longer being used by only the video player

killergerbah:
BTW, I guess this is because of focus-within but I am seeing the VideoPlayer have a blue outline:

image

ShanaryS:
I added the highlight as an option, defaults to `true`. To reiterate, the collection and highlighting of tokens is still dependent on coloring being enabled.

I couldn't replicate the highlight around VideoPlayer but I added something that might address it.

I also updated the names of some of the other dictionary settings to make them consistent. The reason I prefixed the settings with `dictionary` is that some of the settings could have meaning outside of a dictionary context, but now I just prefixed them all. This is a breaking change and will require an updated config. It shouldn't affect the translations though as those all already had the prefix.

killergerbah:
Awesome thanks, the blue outline looks to be addressed as well

PR #870, opened 2026-02-01
Rich text calculated from tokenization data model instead of prerendering + bug fixes

killergerbah:
- Subtitle model now has a structured tokenization field which can be used to calculate the rich text.
- This is used to pre-populate some tokens when "detect and display Netflix ruby" is enabled.
- The pre-populated tokens are respected and we just tokenize around them when we see them.
- Fixed various bugs encountered while working on this:
  - Fix color updates being applied for wrong subtitle file
  - Fix colors not loading if loading same subtitle file twice in a row
  - Fix colors disappearing when changing subtitle offset

### DIFF
--- a/common/subtitle-coloring/subtitle-coloring.ts
+++ b/common/subtitle-coloring/subtitle-coloring.ts
@@ -766,8 +766,8 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
         iterateOverStringInBlocks(
             fullText,
             (_, blockIndex) => externalTokens[blockIndex],
-            (left, right, token?: Token) => {
-                if (token === undefined) {
+            (left, right, existingToken?: Token) => {
+                if (existingToken === undefined) {
                     promise = promise.then(async () => {
                         const model = await this._tokenizationModel(fullText.substring(left, right), index, ts, left);
                         if (this.shouldCancelBuild) return;
@@ -786,7 +786,7 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                     });
                 } else {
                     promise = promise.then(async () => {
-                        const tokenText = fullText.substring(token.pos[0], token.pos[1]);
+                        const tokenText = fullText.substring(existingToken.pos[0], existingToken.pos[1]);
                         const trimmedToken = tokenText.trim();
 
                         const tokenToIndexes = this.tokenToIndexesCache.get(trimmedToken);
@@ -800,13 +800,21 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                             else this.tokenToIndexesCache.set(lemma, new Set([index]));
                         }
 
-                        token.states = ts.tokenStates.get(trimmedToken) ?? [];
-                        token.status =
-                            token.states.includes(TokenState.IGNORED) || !HAS_LETTER_REGEX.test(trimmedToken)
-                                ? getFullyKnownTokenStatus()
-                                : await this._tokenStatus(trimmedToken, ts);
-                        if (this.shouldCancelBuild) return;
+                        const states = ts.tokenStates.get(trimmedToken) ?? [];
+                        const token: Token = {
+                            pos: [existingToken.pos[0], existingToken.pos[1]],
+                            readings: existingToken.readings.map((r) => ({
+                                pos: [r.pos[0], r.pos[1]],
+                                reading: r.reading,
+                            })),
+                            states,
+                            status:
+                                states.includes(TokenState.IGNORED) || !HAS_LETTER_REGEX.test(trimmedToken)
+                                    ? getFullyKnownTokenStatus()
+                                    : await this._tokenStatus(trimmedToken, ts),
+                        };
                         if (token.status === null) this.erroredCache.add(index);
+                        if (this.shouldCancelBuild) return;
 
                         reconstructedTextParts.push(tokenText);
                         allTokens.push(token);
@@ -889,8 +897,8 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                     continue;
                 }
                 token.status = await this._tokenStatus(trimmedToken, ts);
-                if (this.shouldCancelBuild) return;
                 if (token.status === null) this.erroredCache.add(index);
+                if (this.shouldCancelBuild) return;
             }
 
             return { reconstructedText: reconstructedTextParts.join(''), tokenization: { tokens } };


common/app/components/Player.tsx
```diff
@@ -304,6 +304,8 @@ const Player = React.memo(function Player({
                 displayTime: timeDurationDisplay(s.originalStart + offset, length),
                 track: s.track,
                 index: i,
+                richText: s.richText,
```
killergerbah, line 308: Ensures colors don't disappear when offset changes

common/src/model.ts
```diff
@@ -20,6 +20,23 @@ export interface SubtitleTextImage {
...
+    pos: number[]; // Relative position inside parent text
+    states?: TokenState[];
+    status?: TokenStatus;
+    readings?: TokenReading[];
+}
+
+export interface Tokenization {
+    readonly tokens?: Token[];
+    readonly error?: boolean;
+}
```
ShanaryS, line 38: Is there a reason so many fields are optional? I feel like it reduces the type safety of it. Would `Tokenization` ever be created with tokens but not error or vice versa? Same with the fields for `Token`, it seems to be they would always be created together, with the exception of `status` and `readings`. It seems like `readings` could just be an empty array if only colors are enabled but `status` probably does need to be optional.

common/src/model.ts
```diff
@@ -40,6 +58,10 @@ export interface RichSubtitleModel extends IndexedSubtitleModel {
     richText?: string;
 }
 
+export interface TokenizedSubtitleModel extends RichSubtitleModel {
+    tokenization?: Tokenization;
```
ShanaryS, line 63: Same here, if we are using `TokenizedSubtitleModel`, it seems strange that `tokenization` would be optional.

common/src/model.ts
killergerbah: I was avoiding trying to serialize stuff over `sendMessage` instead of sending empty values. Maybe it's an over-optimization.

common/src/model.ts
killergerbah: Was blindly following existing pattern. The purpose of `RichSubtitleModel` seemed to be to make the `richText` mutable, so `TokenizedSubtitleModel` is doing the same. I don't feel strongly about it though because we can just send an empty array.

killergerbah:
Not my best work but I am using `__internal` ("internal" token)  and `__tokenized` (was tokenized "internally") flags to help distinguish between tokens that come from the subtitle reader vs tokens that come from the subtitle coloring.

common/app/components/Player.tsx
```diff
@@ -376,8 +378,11 @@ const Player = React.memo(function Player({
                         displayTime: timeDurationDisplay(s.start + offset, length),
                         track: s.track,
                         index: i,
+                        tokenization: s.tokenization,
```
ShanaryS, line 382: Doesn't this need `richText` as well?

common/subtitle-reader/subtitle-reader.ts
```diff
@@ -501,10 +509,24 @@ export default class SubtitleReader {
...
+        }
 
-        return text.replace(netflixRubyRegex, (_match, base, ruby) => `<ruby><rb>${base}</rb><rt>${ruby}</rt></ruby>`);
+        const tokens: Token[] = [];
+        node.text = node.text.replace(netflixRubyRegex, (_match, base, reading, offset) => {
+            tokens.push({
+                pos: [offset, offset + base.length],
+                readings: [{ pos: [0, reading.length], reading }],
+                states: [],
+                status: getFullyKnownTokenStatus(),
```
ShanaryS: This seems like a good usage for undefined `status` rather than known. We can then treat `undefined` as known when displaying. Or we could still use the actual status. The reading existing in the subtitle probably shouldn't matter for status.

common/src/model.ts
```diff
@@ -20,6 +20,23 @@ export interface SubtitleTextImage {
     readonly image: DimensionsModel;
 }
 
+export interface TokenReading {
+    pos: number[]; // Relative position inside parent token
+    reading: string;
+}
+
+export interface Token {
+    pos: number[]; // Relative position inside parent text
```
ShanaryS: Would `[number, number]` be a better type for both pos? Since it always will be `[index, length]`. Or it could be an object so it's self-descriptive.

common/subtitle-reader/subtitle-reader.ts
```diff
@@ -501,10 +509,24 @@ export default class SubtitleReader {
...
+        if (!node.text) {
+            return;
+        }
 
-        return text.replace(netflixRubyRegex, (_match, base, ruby) => `<ruby><rb>${base}</rb><rt>${ruby}</rt></ruby>`);
+        const tokens: Token[] = [];
+        node.text = node.text.replace(netflixRubyRegex, (_match, base, reading, offset) => {
+            tokens.push({
+                pos: [offset, offset + base.length],
+                readings: [{ pos: [0, reading.length], reading }],
```
ShanaryS: It might be due to how the Netflix parsing works but tokens are typically considered the entire word, not divided up by reading sections. Some words might be `kanji-hiragana-kanji` (<ruby>飛<rt>と</rt></ruby>び<ruby>切<rt>き</rt></ruby>り) which should ideally be parsed as 1 token with multiple readings. But the index always being 0 seems like it doesn't support this case. I'm unsure of how this will affect coloring/statistics since we are respecting this parsing.
Also, the internal readings are:
```ts
readings.push({
    pos: [currentPartOffset, currentPartOffset + part.text.length],
    reading: part.reading,
});
```
Which seems to use the text length instead of the reading length here. I'm also concerned if the offset works correctly since we are replacing left to right

extension/src/services/binding.ts
```diff
@@ -1445,13 +1458,10 @@ export default class Binding {
                         index,
                         originalStart: s.start,
                         originalEnd: s.end,
+                        tokenization: s.tokenization,
```
ShanaryS, line 1461: Should we pass `richText` here as well.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -65,8 +75,71 @@ function shouldUseAnyForm(s: TokenMatchStrategy): boolean {
     return s === TokenMatchStrategy.ANY_FORM_COLLECTED;
 }
 
+const readingsAreSame = (a: TokenReading, b: TokenReading) => arrayEquals(a.pos, b.pos) && a.reading === b.reading;
```
ShanaryS: Is there a benefit to using arrow functions versus regular functions or is it just a style thing?

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -65,8 +75,71 @@ function shouldUseAnyForm(s: TokenMatchStrategy): boolean {
...
+
+const tokenizationsAreSame = (a: Tokenization, b: Tokenization) => {
```
ShanaryS: I had to do something similar for `areDictionaryTracksEqual` and wanted to also make it type safe. I was concerned with future modifications not being compared correctly but it creates a bit more boilerplate. I think something like it would be worth it here if possible.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -65,8 +75,71 @@ function shouldUseAnyForm(s: TokenMatchStrategy): boolean {
...
+}
+
+interface InternalSubtitleModel extends TokenizedSubtitleModel {
+    text: string;
+    __tokenized?: boolean;
+}
+
 export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
-    private _subtitles: RichSubtitleModel[];
+    private _subtitles: InternalSubtitleModel[];
```
ShanaryS, line 120: This may no longer be true but at one iteration of `SubtitleColoring` these subtitles were shared with the external state so it could end up being `TokenizedSubtitleModel`. It probably wouldn't affect the correctness though as long as `_resetCache()` covers it.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -136,11 +209,22 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
+            // Preserve existing cache here so callers don't need to be aware of it
+            const subtitlesWithRichText: TokenizedSubtitleModel[] = [];
+
+            for (const s of subtitles) {
+                s.tokenization = this._subtitles[s.index].tokenization;
+                s.richText = this._subtitles[s.index].richText;
+                if (s.richText !== undefined) {
+                    subtitlesWithRichText.push(s);
+                }
+            }
```
ShanaryS, line 213: Should we also pass in `__internal` here? If not a comment might be helpful explaining why.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -161,7 +245,13 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
         this.ankiRecentlyModifiedTrigger = false;
         this.ankiRecentlyModifiedFirstCheck = true;
-        this._subtitles.forEach((s) => (s.richText = undefined));
+        this._subtitles.forEach((s) => {
+            s.__tokenized = undefined;
+            s.richText = undefined;
+            if (s.tokenization) {
+                s.tokenization.tokens = s.tokenization.tokens.filter((t) => !(t as InternalToken).__internal);
+            }
+        });
```
ShanaryS: Should we also reset `__internal` here? It seems like `_resetCache()` would be in a different state than the first time load which I believe I was trying to maintain, don't know if that's currently true though.

common/app/components/Player.tsx
killergerbah: This is just after coming out of subtitle reader so it won't have any rich text. The following call to `renderRichTextOntoSubtitles` will put the rich text on.

common/src/model.ts
killergerbah: I didn't know you could do that

common/subtitle-reader/subtitle-reader.ts
killergerbah: Netflix readings are always `word` followed by `(reading of word)` so we don't need to deal with the case where readings can be in different places inside the word

> Which seems to use the text length instead of the reading length here.

That was a mistake, will update to use reading length

> I'm also concerned if the offset works correctly since we are replacing left to right

I've re-read the logic and it looks correct to me. But I could be missing something.

extension/src/services/binding.ts
killergerbah: I don't think we should, ideally `richText` is never sent across chrome extension messaging API and passing `richText` here would hide that error if it happened

`subtitle-controller` will calculate the rich text if it sees a tokenization

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: There's no benefit in this case, it's a style thing.
But it does have actual functional differences, which I am forgetting right now.

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: I've put the exported function in `util` for now since the equality functions depends on other functions in `util`

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: Yeah I think it's fine. `InternalSubtitleModel` only adds optional field + makes one field mutable.

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: Right that make sense. Copying the `__tokenized` so that it doesn't get retokenized later (unless reset cache is called).

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: I think that's what's being reset already, unless you mean something else? The tokens list is being reset to all the tokens that don't have the __internal field - so all the Netflix readings will be retained but everything else gets blown away.

common/subtitle-reader/subtitle-reader.ts
killergerbah: I've made this change, but it means that the red line-through style will almost never be rendered. Which I'm fine with, but maybe I made the wrong change.

common/subtitle-reader/subtitle-reader.ts
ShanaryS: > Netflix readings are always `word` followed by `(reading of word)`

So it would show `飛び切り(とびきり)` parsed as <ruby>飛び切り<rt>とびきり</rt></ruby> or `飛び切り(とき)` parsed as <ruby>飛び切り<rt>とき</rt></ruby>?

In either case I think that means we should ignore the Netflix readings, they are worse that what we get parsing using Yomitan since we get the accurate readings per kanji grouping: <ruby>飛<rt>と</rt></ruby>び<ruby>切<rt>き</rt></ruby>り

So I think we can do away with `__internal` and `InternalToken` then overwrite the existing Netflix tokenization to improve it and make it consistent. It also drastically simplifies `SubtitleColoring` as we no longer need to handle them separately with `_tokenizationModelMergedWithExistingOne()` and others.

common/subtitle-reader/subtitle-reader.ts
ShanaryS: I didn't realize you changed the error state of `tokenStatus` from `null` to `undefined`. I think if we keep errors at null and do the strike though then it's fine.

Separately with my other comment about overwriting the Netflix readings we can also get the proper status for the token as well. As the more I think about it, the less I think existing readings should affect the status a user sees. I think this means `null` or `undefined` can be used since it will only apply to errors now, but I like how `null` is explicit as a "we attempted to parse but it failed" while `undefined` could mean "we didn't attempt parsing".

common/subtitle-coloring/subtitle-coloring.ts
ShanaryS: What I mean is that `__internal` only gets set on tokens after they've gone through this class. But after resetCache they are already populated which gives the opportunity for a behavior change.

ShanaryS: Oh I see, I didn't realize `__internal` is only on the tokens and so that's what's being reset here. I misunderstood what was happening.

common/subtitle-reader/subtitle-reader.ts
killergerbah: Yea but usually the only reason Netflix readings exist in the first place is to provide readings for words where they're actually necessary to know the correct reading. In these cases Yomitan would just provide an inaccurate reading most of the time. For example, I've been testing this with Jujutsu Kaisen subs and Netflix gives you まひと for the name 真人 while Yomitan would give you しんじん. If there is any logic that would ignore the reading even though the 'display netflix readings' setting is enabled, then we should probably just remove the setting altogether, which imo is a valid option.

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: After resetCache all the internal tokens are removed, by the call to `filter`

killergerbah: I replied too late lol

common/subtitle-reader/subtitle-reader.ts
killergerbah: Changing null to undefined might have been an accident, ill change it back

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -666,46 +829,38 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
...
             this.tokenRequestFailed = true;
             console.error(`Error colorizing subtitle text for Track${ts.track + 1}:`, error);
             this.erroredCache.add(index);
-            return this._applyTokenStyle([{ text, reading: '' }], null, ts.dt, {
-                validToken: false,
-                ignoredToken: false,
-            });
+            return { reconstructedText: fullText, tokenization: { tokens: [], error: true } };
+        }
+    }
```
ShanaryS: It doesn't seem like errors are propagated correctly for 3 reasons:
1. Per token errors are set to `undefined` for which `computeRichText` doesn't doesn't call `applyTokenStyle` for.
2. `textHasError` is never true since it checks for `tokenStatus === null` but it was coerced into `undefined` earlier.
3. `tokenization.error` doesn't seem to be used in `computeRichText` even when no tokens to strike through the entire line.

common/subtitle-reader/subtitle-reader.ts
ShanaryS: If the Netflix readings are only for stuff like names or even if it's just more accurate, then I think we should just prefer them. As accuracy is more important than readings being perfectly centered.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -1067,3 +1162,99 @@ export class HoveredToken {
         return token;
     }
 }
+
+export const renderRichTextOntoSubtitles = (subtitles: RichSubtitleModel[], dictionaryTracks?: DictionaryTrack[]) => {
+    for (const s of subtitles) {
+        if (s.tokenization && !s.richText) {
+            s.richText = computeRichText(s.text, s.tokenization, dictionaryTracks?.[s.track]);
+        }
```
ShanaryS, line 1241: When I modify a token locally, it doesn't seem to be updating the `richText`. I haven't looked into why but it's probably here or `_tokenizationModelMergedWithExistingOne()`. I have to update a setting for it to show the new colors.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -1067,3 +1162,99 @@ export class HoveredToken {
...
+        (left, right, token?: Token) => {
+            if (token === undefined) {
+                parts.push(fullText.substring(left, right));
+            } else {
+                parts.push(applyTokenStyle(fullText, token, dt));
+            }
+        }
+    );
+    return parts.join('');
+};
```
ShanaryS, line 1267: It seems that the readings aren't applied currently when the word starts with kanji and followed by hiragana. Such as 信じる which is displayed as: <ruby>信じ<rt>しん</rt></ruby>る

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: I don't entirely remember why I needed `Tokenization.error`. But with bringing `null` back to `Token.status` I think the original behavior is preserved.

common/subtitle-reader/subtitle-reader.ts
killergerbah: Even if we keep Netflix readings, tbh I think it's still kind of confusing. Like the Netflix reading will show up only the first time, since Japanese text usually only shows you readings the first time a word appears. Since the Netflix reading is missing for remaining instances of the word, Yomitan is used instead. So it might be good to solve this too by either removing Yomitan-generated readings for the same token or keeping a map of the Netflix readings and using those. But I think both of these options are not perfect.

killergerbah: Maybe for another PR

common/subtitle-coloring/subtitle-coloring.ts
killergerbah: Ok sorry for the laziness. It's been pretty long week. I think I've fixed up the `null` coercion and rendering of errored tokenizations.

ShanaryS (approved):
Rather than doing another review and to better help myself understand the code, I made some changes to address issues and revert some behavior:
* Updated reading `pos` calcuation to use the `base.length` as the end instead of `reading.length` to match the corrected construction
* Now keeping track of `originalText` before modifying `text` so that we can ensure that `needsReset` doesn't run erroneously.
* Also preserve `text` when `needsReset` is false
* Set `tokenization` to undefined for `untokenize()` if no tokens are present
* Use `__tokenized` instead of `tokenization` to determine if the event has already been processed when determining build cancellation
* Simplified `areTokenizationsEqual()` by moving checks inside the function
* Instead of `untokenize()` early when tokens need to be refreshed, remove the mutation on `existingTokenization` so `areTokenizationsEqual()` works correctly and set `richText` to `undefined` when `tokenization` is updated.
	* This fixes the text flashing uncollected when changing status/state since `untokenize()` was called too early
* Explicitly prevent `undefined` token status in `applyTokenStyle`, Netflix tokens must have their status determined rather than assuming fully known
	* This could cause these tokens to flash logic error (double line through instead of single) upon first load theoretically
* Fix error token returning the `token` object instead of `tokenText`

common/subtitle-reader/subtitle-reader.ts
ShanaryS: Yeah I agree, I think this is pretty much necessary as it will be too confusing for users to see the correct reading initially then the incorrect Yomitan readings later. I wasn't sure how the Netflix readings worked before but now I think we'll have to tackle this.

I was also thinking there might be a way to fix the unprecise readings. We could probably process the Netflix readings to only apply it to just the Kanji instead of being spread across the entire word.

killergerbah:
Sweet thanks @ShanaryS 
Will look again and maybe we can merge today or tomorrowish

ShanaryS:
I added reusing the external token readings. Any internal tokens will use the external token reading if it exists. I think this is better than just removing the readings for their subsequent appearance as the Yomitan readings are always shown, not just the first time.

I ended up not trying to parse the readings into their precise form into tokenParts, so that hiragana doesn't get redundant readings. I believe it is possible but not worth it given how infrequent these readings are and it's just a visual thing. I also think keeping them gives us a way to quickly distinguish the source of the word reading, if the word contains hiragana characters.

killergerbah:
I see, did you end up not keeping that change? I'm seeing that 真人 is still being parsed as しんじん
image

Btw I am seeing a bug, I can look into it when I have time today or tomorrow. For some reason the first 虎杖 does not get updated in realtime with local token color. I guess it might have to do with the external reading.


https://github.com/user-attachments/assets/fd551cb5-9b5d-4ee5-8563-fdb28383e55c

ShanaryS:
> For some reason the first 虎杖 does not get updated in realtime with local token color. I guess it might have to do with the external reading.

The issue was reusing the same token object from `existingTokeniation`, so `areTokenizationsEqual()` would always pass and not send the updated subtitles. The new method that I switched to a few commits ago is sensitive to the mutation of `existingTokenization` but I forgot to create new tokens as well.

> I see, did you end up not keeping that change? I'm seeing that 真人 is still being parsed as しんじん

It's still working, but it seems like your second 真人 was not parsed correctly, the ) and the following space should never be a part of a token unless Yomitan returned that. It's the only thing I can see being the issue. Hovering over it should highlight it as well if it's the problem.

It seems to be working for 虎杖 as they both have the same reading (assuming that is also external).

killergerbah:
> It's still working, but it seems like your second 真人 was not parsed correctly, the ) and the following space should never be a part of a token unless Yomitan returned that. It's the only thing I can see being the issue. Hovering over it should highlight it as well if it's the problem.

I verified it's working. I think I forgot to pull like a dumbass, lol. Merging now.

PR #873, opened 2026-02-04
feat: anki plugin (v1)

mwojick:
This adds an anki plugin for one click mining. In theory it could probably implement the other functionality from the proxy server as well but would need an http server.

To add it manually just copy the asbplayer-plugin folder over to where your addons are installed
(on Windows: C:\Users\name\AppData\Roaming\Anki2\addons21)

Also changing the config requires an Anki restart at the moment.

Note that the monkey patching to Collection.add_note is needed because apparently AnkiConnect doesn't trigger Anki's built in hooks like add_cards_did_add_note.

### DIFF
--- a/plugins/anki/asbplayer-plugin/__init__.py
+++ b/plugins/anki/asbplayer-plugin/__init__.py
@@ -43,7 +43,7 @@ def _handle_note_added(note: Note) -> None:
         },
     }
 
-    server.broadcast_message(command)
+    server.send_message(command)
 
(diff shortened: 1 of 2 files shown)


killergerbah:
Great, this basically replaces the existing WebSocket server. Once we have the HTTP server as well then I don't think there's a reason to prefer the Go-based one anymore.

killergerbah:
Thanks

PR #878, opened 2026-02-14
chore: update wxt and fix dev config

mwojick:
The dev and production versions were conflicting in weird ways which basically forced you to uninstall the production one before loading in the dev one. This allows them to be installed side by side (though you probably should only enable one at a time), and adds a Dev label to easily tell them apart. Chrome doesn't need an explicit key in dev mode since it seems to make a stable one based on the path name.

Added bonus is I found out about artifactTemplate and sourcesTemplate to override the default zip names that were pulling from the package.json before (i.e. projectextension).

### DIFF
--- a/extension/src/services/active-tab-permission-request.ts
+++ b/extension/src/services/active-tab-permission-request.ts
@@ -1,13 +1,20 @@
 const key = 'tabRequestingActiveTabPermission';
 
+interface SavedTab {
+    tabId: number;
+    url: string;
+    src: string;
+}
+
 export const getTabRequestingActiveTabPermission = async () => {
     const result = await browser.storage.session.get(key);
-    const savedTab = result ? result[key] : undefined;
+    const tab = result ? result[key] : undefined;
 
-    if (savedTab === undefined) {
+    if (!tab) {
         return undefined;
     }
 
+    const savedTab = tab as SavedTab;
     const currentTabInfo = await tabInfo(savedTab.tabId);
 
     if (currentTabInfo === undefined || currentTabInfo.url !== savedTab.url) {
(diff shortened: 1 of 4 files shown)


extension/wxt.config.ts
```diff
@@ -46,11 +50,15 @@ export default defineConfig({
         },
     },
     manifest: ({ browser, mode }) => {
+        const version = '1.14.0';
+        const isDev = mode === 'development';
+        const name = `${extName}${isDev ? ' (Dev)' : ''}: Language-learning with subtitles`;
```
killergerbah: Extract this part into variable? 
```
isDev ? ' (Dev)' : ''
```
IMO that would be more readable

extension/wxt.config.ts
```diff
@@ -46,11 +50,15 @@ export default defineConfig({
...
+
         let manifest: UserManifest = {
-            name: 'asbplayer: Language-learning with subtitles',
+            name,
             description: '__MSG_extensionDescription__',
-            version: '1.14.0',
-            action: { default_title: 'asbplayer' },
+            version,
+            action: { default_title: `${extName}${isDev ? ' (Dev)' : ''}` },
```
killergerbah: Can use that variable here too

extension/wxt.config.ts
```diff
@@ -156,22 +164,32 @@ export default defineConfig({
...
                 ...manifest,
                 host_permissions: ['<all_urls>'],
                 commands,
                 browser_specific_settings: {
                     gecko: {
-                        id: '{e4b27483-2e73-4762-b2ec-8d988a143a40}',
-                        update_url: 'https://killergerbah.github.io/asbplayer/firefox-extension-updates.json',
+                        id: geckoId,
+                        ...(isDev
+                            ? {}
```
killergerbah: In general wondering if we can group these `isDev` branches together. Since we have `geckoId` assigned using `isDev` ternary expression, and `update_url` is also assigned using `isDev` ternary expression, just in a different place. In case it's possible to make things more readable without a lot of effort.

killergerbah (commented):
Thanks, yea this will be useful. I've probably wasted a lot of time uninstalling the dev version so I could install the prod version and vice versa.

extension/wxt.config.ts
mwojick: yeah I think I have something that's a bit cleaner now, though with the nested conditionals it's still not perfect

PR #879, opened 2026-02-14
Display word frequency under token

ShanaryS:
This adds a new option `TokenFrequencyAnnotation` which can be `Always`, `Uncollected only`, and `Never`.

Currently we need an API request to yomitan per token so it's slow and cannot be bulked. To mitigate this, fetching the frequency is non-blocking and will trickle in over time. A full solution will require changes in Yomitan which should hopefully be resolved.

Users will need to have frequency dictionaries installed and they need to be `rank-based`. Checking for this will also be present the Yomitan PR.

### DIFF
--- a/common/subtitle-coloring/subtitle-coloring.ts
+++ b/common/subtitle-coloring/subtitle-coloring.ts
@@ -120,6 +120,12 @@ function originalTokenization(tokenization: Tokenization | undefined): Tokenizat
     };
 }
 
+function resetYomitan(ts: TrackState) {
+    if (!ts.yt) return;
+    ts.yt.resetCache();
+    ts.yt = undefined;
+}
+
 export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
     private _subtitles: InternalSubtitleModel[];
     private readonly dictionaryProvider: DictionaryProvider;
@@ -148,7 +154,7 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
     private colorCacheBuilding: boolean;
     private colorCacheBuildingCurrentIndexes: Set<number>;
     private shouldCancelBuild: boolean; // Set to true to stop current color cache build, checked after each async call
-    private tokenRequestFailed: boolean;
+    private tokenRequestFailedForTracks: Set<number>;
 
     private readonly subtitleColorsUpdated: (updatedSubtitles: RichSubtitleModel[], dt: DictionaryTrack[]) => void;
     private readonly getMediaTimeMs?: () => number;
@@ -190,7 +196,7 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
         this.colorCacheBuilding = false;
         this.colorCacheBuildingCurrentIndexes = new Set();
         this.shouldCancelBuild = false;
-        this.tokenRequestFailed = false;
+        this.tokenRequestFailedForTracks = new Set();
     }
 
     get subtitles() {
@@ -252,7 +258,7 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
         if (this.colorCacheBuilding) this.shouldCancelBuild = true;
         this.profile = null;
         this.anki = undefined;
-        this.trackStates.forEach((ts) => ts.yt?.resetCache());
+        this.trackStates.forEach(resetYomitan);
         this.trackStates = [];
         this.ankiRecentlyModifiedCardIds.clear();
         this.ankiRecentlyModifiedTrigger = false;
@@ -529,14 +535,8 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
             }
 
             try {
+                for (const subtitle of subtitles) this.colorCacheBuildingCurrentIndexes.add(subtitle.index);
                 await this._buildTokenAndLemmaMap(profile, subtitles);
-            } catch (e) {
-                console.error('Error building token and lemma map:', e);
-                this.trackStates.forEach((ts) => {
-                    if (!ts.yt) return;
-                    ts.yt.resetCache();
-                    ts.yt = undefined; // Propagate error so that subtitles are error styled
-                });
             } finally {
                 this.colorCacheBuildingCurrentIndexes.clear();
             }
@@ -606,15 +606,11 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                 updateThresholds = false;
             }
         } finally {
-            if (this.tokenRequestFailed) {
-                this.tokenRequestFailed = false;
-                this.trackStates.forEach((ts) => {
-                    if (!ts.yt) return;
-                    ts.yt.resetCache();
-                    ts.yt = undefined;
-                });
+            if (this.tokenRequestFailedForTracks.size) {
                 tokensRefreshed = [];
                 updateThresholds = false;
+                for (const track of this.tokenRequestFailedForTracks) resetYomitan(this.trackStates[track]);
+                this.tokenRequestFailedForTracks.clear();
             } else if (!this.shouldCancelBuild) {
                 this.initialized = true;
             }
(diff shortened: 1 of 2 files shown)


common/yomitan/yomitan.ts
```diff
@@ -98,6 +114,39 @@ export class Yomitan {
...
+        if (!this.supportsTokenizeFrequency) return;
+        let minFrequency: number | undefined;
+        for (const headwords of tokenizeHeadwords) {
+            for (const headword of headwords) {
+                for (const source of headword.sources) {
+                    if (source.originalText !== token) continue;
+                    if (!source.isPrimary) continue;
+                    if (source.matchType !== 'exact') continue;
+                    if (source.matchSource !== 'term' && preferTermSource) continue; // Frequency of this exact form, don't promote rare kanji
+                    for (const f of headword.frequencies) {
```
killergerbah, line 146: I think this might be where I was hitting an error, since `frequencies` was missing. Should we do a defensive check here to ensure it's defined before iterating?

common/yomitan/yomitan.ts
```diff
@@ -15,17 +15,29 @@ export class Yomitan {
...
+    private readonly frequencyCache: Map<string, number | null>;
+    private supportsTokenizeFrequency: boolean;
+    private tokensWereModified?: (token: string) => void;
+    private tokensWereModifiedPromise?: Promise<void>;
 
-    constructor(dictionaryTrack: DictionaryTrack, fetcher = new HttpFetcher()) {
+    constructor(
+        dictionaryTrack: DictionaryTrack,
+        fetcher = new HttpFetcher(),
+        tokensWereModified?: (token: string) => void
```
killergerbah, line 27: Is it a problem that this function cannot be unsubscribed? Thinking of a possible issue where the frequency call queue takes a while to drain and `tokensWereModified` is called even though this Yomitan instance has been let go.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -815,6 +843,14 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                         };
                         if (token.status === null) this.erroredCache.add(index);
                         if (this.shouldCancelBuild) return;
+                        if (
+                            ts.dt.dictionaryTokenFrequencyAnnotation === TokenFrequencyAnnotation.ALWAYS ||
+                            (token.status === TokenStatus.UNCOLLECTED &&
+                                ts.dt.dictionaryTokenFrequencyAnnotation === TokenFrequencyAnnotation.UNCOLLECTED_ONLY)
+                        ) {
+                            token.frequency = await ts.yt!.frequency(trimmedToken);
```
killergerbah: Wouldn't this block until this token gets its turn to be included in a frequency query?

killergerbah: Ah never mind I see it is non blocking if the `tokensWereModified` callback is provided

common/yomitan/yomitan.ts
ShanaryS: There is a gate at the top of this function that requires `this.supportsTokenizeFrequency`. The issue you had is that since you had a custom build of Yomitan the version was `0.0.0.0` which will assume you support all the features for the versions of Yomitan asbplayer cares about. If `this.supportsTokenizeFrequency === true`, then `headword.frequencies` is guaranteed to exist (pending PR).

common/yomitan/yomitan.ts
ShanaryS: I added unsubscribing keeping track of timestamps.

ShanaryS:
I changed the in flight loading to use a new class `AsyncSemaphore` which is from one of my other projects. This maintains the FIFO ordering so that relevant frequencies are calculated first. I also added `this.initialized` which is set to `true` on the small initial cache build. This ensures that we only start getting frequencies after the main tokenize work is done (if frequencies aren't supported in tokenize).

> What are you planning for the styling? At first glance it feels like using numbers to express the frequency might make the text feel crowded for the average user.

I've only ever thought of using numbers for this. I think using another abstraction like color (something that covers ranges of frequencies) is actually more complex since the user will need to remember what it means along with token status. It would now require a lot more settings too so a user can customize what each level means. Also if we ever add pitch accent we will need to use a token styling for that and so it can become extremely infromation dense. 

I think the frequency is useless for everything except uncollected tokens, the only real value of it is to know if something is worth adding to Anki. So if they use uncollected only, they can hover+shift to check the frequencies in Yomitan as needed for everything else to reduce the clutter. This feature only brings value if it can be more convenient than the already convenient Yomitan lookup (aside from future features using frequency for automations).

common/util/util.ts
```diff
@@ -521,60 +521,92 @@ export function iterateOverStringInBlocks<T, B extends Block>(
...
-            return false;
+    }
+
+    acquire(): Promise<number> {
+        return new Promise<number>((resolve) => {
+            if (this.permits > 0) {
+                this.permits--;
+                const id = this.getNextId();
+                this.acquired.add(id);
+                if (this.lifetimeMs) {
```
killergerbah: Looks like this check can be removed since it's guaranteed positive by the constructor

common/util/util.ts
```diff
@@ -521,60 +521,92 @@ export function iterateOverStringInBlocks<T, B extends Block>(
...
+
+    release(id: number): void {
+        if (!this.acquired.has(id)) return;
+        this.acquired.delete(id);
+        clearTimeout(this.timers.get(id)!);
+        this.timers.delete(id);
+
+        if (this.waiting.length > 0) {
+            const newId = this.getNextId();
+            this.acquired.add(newId);
```
killergerbah: Resolving the next promise feels like an opportunity to deduplicate some code. These lines look similar to the ones in `acquire`

common/yomitan/yomitan.ts
```diff
@@ -13,19 +13,39 @@ export interface TokenPart {
...
+    ) {
         this.dt = dictionaryTrack;
         this.fetcher = fetcher;
+        this.asyncSemaphore = new AsyncSemaphore({ permits: 1, lifetimeMs: 10000 });
```
killergerbah: Is there a reason to limit to 10s? feels like we could have an effectively infinite time limit unless there are performance concerns

killergerbah:
Interesting, expressing frequency with numbers make sense then

common/util/util.ts
ShanaryS: Do you mean `if (this.permits > 0)`? It's necessary to know if we should acquire immediately or we need to wait for a slot.

common/yomitan/yomitan.ts
ShanaryS: The lifetime is to prevent deadlocks, it only matters if callers may forget to call `release()` in a try/finally block. I've removed it since our usage of it is simple.

common/util/util.ts
killergerbah: I meant lifetimeMs. It's always positive right?

ShanaryS: This check is for if it's undefined.

killergerbah: Oh right im a dumbo

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -590,31 +608,33 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
         } finally {
             if (this.tokenRequestFailed) {
                 this.tokenRequestFailed = false;
-                trackStates.forEach((ts) => (ts.yt = undefined));
+                this.trackStates.forEach((ts) => {
```
killergerbah: This works because only one instance of `buildColorCache` can be executing at any given time? Just want to make sure that it's not possible to reset a `Yomitan` that belongs to a different track state instance.

common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -899,6 +931,18 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                 token.status = await this._tokenStatus(trimmedToken, ts);
                 if (token.status === null) this.erroredCache.add(index);
                 if (this.shouldCancelBuild) return;
+                if (
+                    ts.dt.dictionaryTokenFrequencyAnnotation === TokenFrequencyAnnotation.ALWAYS ||
+                    (token.status === TokenStatus.UNCOLLECTED &&
+                        ts.dt.dictionaryTokenFrequencyAnnotation === TokenFrequencyAnnotation.UNCOLLECTED_ONLY)
+                ) {
+                    if (ts.yt.getSupportsTokenizeFrequency() || this.initialized) {
+                        token.frequency = await ts.yt.frequency(trimmedToken);
```
killergerbah: Suggest to deduplicate this block since the same one is on line 838

common/subtitle-coloring/subtitle-coloring.ts
ShanaryS: Yeah only one instance of `buildColorCache` can execute at a time but it was resetting all`Yomitan` instances if any failed. I updated as we are now caching a lot more.

common/yomitan/yomitan.ts
```diff
@@ -147,13 +208,82 @@ export class Yomitan {
...
+        if (!HAS_LETTER_REGEX.test(token)) {
+            this.frequencyCache.set(token, null);
+            return;
+        }
+        if (this.tokensWereModified) {
+            void (async () => {
+                const now = Date.now();
+                const semaphoreId = await this.asyncSemaphore.acquire();
+                try {
+                    if (now <= this.lastCancelledAt) {
```
killergerbah, line 227: Is the direction of the comparison correct? Otherwise it looks like `lastCancelledAt` is not being used to cancel

ShanaryS: The goal is to cancel anything that was started before `resetCache()`. The time is recorded before trying to acquire which could take minutes.

killergerbah: Oh yeah my brain is really shrinking

ShanaryS:
I think we should merge and release with 1.14. It's pretty low stakes since it loads it asynchronously and if it's kept as `NEVER` then it has no effect on the existing features. It also contains a few bug fixes like the Yomitan reset and `trackStates` not updating properly causing duplicate requests at startup.

PR #882, opened 2026-02-14
Fix some issues related to Netflix readings

killergerbah:
- If multiple Netflix readings appear in same subtitle, their offsets should render correctly now. Tested with "無為‎転変" in the attached file.
- If colorization is disabled, then returning token text as-is from `applyTokenStyle`. Without this, words with Netflix readings show up with the error style.

[呪術廻戦 S02E45 変身 - ja - 日本語 [CC].nfvtt.txt](https://github.com/user-attachments/files/25314637/S02E45.-.ja.-.CC.nfvtt.txt)

### DIFF
--- a/common/subtitle-coloring/subtitle-coloring.ts
+++ b/common/subtitle-coloring/subtitle-coloring.ts
@@ -1272,17 +1272,17 @@ const LOGIC_ERROR_STYLE = `style="text-decoration: line-through red 3px double;"
 const applyTokenStyle = (fullText: string, token: Token, dt?: DictionaryTrack) => {
     const tokenText = applyReadingAnnotation(fullText, token, dt);
     if (token.status === null) return `<span ${ERROR_STYLE}>${tokenText}</span>`;
-    if (!dt?.dictionaryColorizeSubtitles) return tokenText;
-    if (token.status === undefined) {
+    if (token.status === undefined && dt && dictionaryTrackEnabled(dt)) {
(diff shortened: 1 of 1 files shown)


common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -1271,11 +1271,11 @@ const LOGIC_ERROR_STYLE = `style="text-decoration: line-through red 3px double;"
 
 const applyTokenStyle = (fullText: string, token: Token, dt?: DictionaryTrack) => {
     const tokenText = applyReadingAnnotation(fullText, token, dt);
+    if (!dt?.dictionaryColorizeSubtitles) return tokenText;
```
ShanaryS: The only thing is that we won't see the error displayed if the user only has annotations or frequency enabled. I think the null check should be before this, only the undefined check shouldn't. `token.status` probably doesn't get set since `dictionaryStatusEnabled()` is probably false unless you use the right value of reading or frequency display.

killergerbah: Makes sense, `dictionaryColorizeSubtitles` only refers to colorization not to the other annotation features

ShanaryS (approved):

ShanaryS:
I realize the really issue is that `applyTokenStyle()` has been decoupled from `SubtitleColoring` so it can be called when `dictionaryTrackEnabled()` is false. Checking that should be the only thing that's needed as `token.status` shouldn't be `undefined` even if just using readings and frequencies.

Issue #374, opened 2024-02-22
Add timestamps bookmark function for subless listening experience similar to some youtube bookmark extensions


phamthaison1207:
When I am watching anime or anything without subs, I want to be able to save a timestamp of where I couldn't decode what they were saying and noticed there isn't such a feature yet in asbplayer like a timestamp bookmark extension for it.
So I think suggesting this feature will be really useful for any listening based learner like me.
![image](https://github.com/killergerbah/asbplayer/assets/63585568/bb563c13-b944-4b6b-9647-a7efcfebb39b)

phamthaison1207:
ah yeah, for my training, I don't stop and record audio (it spoils the content for my brain) but just mark the timestamp so I have to use listening to comprehend instead of making a habit of relying on the spoilers from asbplayer

phamthaison1207:
like: from a to z: rely purely on listening to decode the content : hit or miss type of training, no stopping and turning on subs
![image](https://github.com/killergerbah/asbplayer/assets/63585568/5537e91e-9358-4b20-81fa-c683c6af910f)

killergerbah:
In that case you could use the 'mine subtitle' keyboard shortcut (not: 'mine subtitle and open anki dialog'). Asbplayer should copy the subtitle into your mining history and continue playing. You will have to try to not look at the notification that pops up however.

naiza-asaad:
Hi @killergerbah, is there a button to save to the mining history? I'm thinking of buying a Windows tablet for portability but I'm not sure how to save the sentence without relying on the keyboard shortcut.

killergerbah:
@naiza-asaad If you want to mine from streaming video, you'll just need to confirm that you can install some version of Chromium that supports chrome extensions onto your tablet. Assuming asbplayer can be installed then you should be able to use the side panel OR video overlay (enabled by default or via 'streaming video' settings) to mine using a button. You can give it a try on desktop to see if that's something that would work for you.

If you want to mine local video files, the + button that appears next to every subtitle in the subtitle list should suffice.

naiza-asaad:
It's just going to be a Windows tablet PC so I guess it'll have a similar look as on my laptop right now? I didn't notice that the button at the side panel also adds to the mining history so this will work for me, thanks! Although currently it always opens the export Anki dialog. Is there/will there also be a button equivalent to Mine current subtitle (without opening the Anki dialog)?

naiza-asaad:
[...]

PR #885, opened 2026-02-15
Improve freeflow immersion: Jump To seeks playback and keep scroll position

L-M-Sherlock:
My current workflow of Free-Flow Immersion:
1. I hide the subtitle during watching.
2. When I don't understand the current dialogue, I use "Mine current subtitle" to add the current line to the mining history. I learn this method from  https://github.com/killergerbah/asbplayer/issues/374#issuecomment-1960879291
3. When I finish the entire episode, I open the mining history and mine the unknown cards.

The problems of the current workflow:
1. The "jump to" doesn't affect the video. To use it, I have to show the subtitle list and locate the video via clicking the subtitle located by "jump to".
2. After I mine the unknown word via yomitan, the web-socket-server help me add the picture and sentence audio. But this action refresh the mining history with a new log, which let me have to locate the next line I want to jump to again. When I have 30+ logs in the mining history, it's annoying because I have to spend 50% time to re-locate the next line.

This PR:
1. The "jump to" will locate the subtitle in the list and the video in the same time, which save me an extra click.
2. New mining history log will not refresh the side panel if my current position is not the bottom.
3. I add `/scripts/web-socket-server/asbplayer-ws` to the `.gitignore` because I made the server as a background service which is based on the binary file.

Now I can skip the step 2 shown in the following screenshot:

image

Update:

1. 35395fa `scripts/web-socket-server/build` builds main.go into asbplayer-ws in the same directory

### DIFF
--- a/common/app/components/App.tsx
+++ b/common/app/components/App.tsx
@@ -776,6 +776,10 @@ function App({
             try {
                 let { subtitleFiles, videoFile } = extractSources(files);
 
+                if (videoFile || subtitleFiles.length > 0) {
+                    setJumpToSubtitle(undefined);
+                }
+
                 setSources((previous) => {
                     let videoFileUrl: string | undefined = undefined;
 


common/app/components/CopyHistoryList.tsx
```diff
@@ -245,16 +245,32 @@ export default function CopyHistoryList({
...
 
+    const handleScroll = useCallback(() => {
+        const el = listContainerRef.current;
+        if (!el) {
+            return;
+        }
+
+        const threshold = 20;
+        const distanceToBottom = el.scrollHeight - el.scrollTop - el.clientHeight;
+        setShouldAutoScroll(distanceToBottom <= threshold);
```
killergerbah: After testing this it looks like `setShouldAutoScroll` is causing the list to re-render when you get near the bottom. Which would normally be fine but it's causing a snappy behavior which feels a little unexpected. Instead of using React state for this you can just calculate whether to scroll or not when a new element is added.

common/app/components/CopyHistoryList.tsx
```diff
@@ -245,16 +245,32 @@ export default function CopyHistoryList({
     onAnki,
 }: CopyHistoryListProps) {
     const classes = useStyles();
-    const scrollToBottomRefCallback = useCallback((element: HTMLElement | null) => {
-        if (element) {
-            element.scrollIntoView();
-        }
-    }, []);
+    const listContainerRef = useRef<HTMLDivElement | null>(null);
+    const [shouldAutoScroll, setShouldAutoScroll] = useState<boolean>(true);
```
killergerbah: As mentioned below, suggest to not use React state for this since you can just calculate the same flag at the point when an element is added.

common/app/components/CopyHistoryList.tsx
```diff
@@ -245,16 +245,32 @@ export default function CopyHistoryList({
...
+    const scrollToBottomRefCallback = useCallback(
+        (element: HTMLElement | null) => {
+            if (element && shouldAutoScroll) {
```
killergerbah: This code is working:

```
        if (!element || !listContainerRef.current) {
            return;
        }

        const listElement = listContainerRef.current;

        const threshold = 20;
        const distanceToBottom =
            listElement.scrollHeight - listElement.scrollTop - listElement.clientHeight - element.clientHeight;
        const shouldAutoScroll = distanceToBottom <= threshold;
        if (shouldAutoScroll) {
            element.scrollIntoView();
        }
```

I think I see why you are using state, since by this point it's difficult to calculate the distance from bottom. Because the newest element is already inserted. But compensating for this by subtracting `element.clientHeight` seems to work.

.gitignore
```diff
@@ -32,3 +32,6 @@ yarn-error.log*
 .pnp.*
 
 .idea/
+
+# local binaries
+/scripts/web-socket-server/asbplayer-ws
```
killergerbah, line 37: If you're including this can you also add a convenience  script that builds `main.go` to a binary named `asbplayer-ws`? Otherwise for someone else to benefit from this line they'd have to be coincidentally using the same binary name `asbplayer-ws`.

common/app/components/SubtitlePlayer.tsx
```diff
@@ -693,37 +693,53 @@ export default function SubtitlePlayer({
...
-        let i = 0;
+        const jumpToIndex = findJumpToIndex();
+        const target = jumpToIndex !== -1 ? subtitles[jumpToIndex] : jumpToSubtitle;
 
-        for (let s of subtitles) {
-            if (s.originalStart === jumpToSubtitle.originalStart && jumpToSubtitle.text.includes(s.text)) {
-                jumpToIndex = i;
-                break;
-            }
+        if (target.start !== undefined) {
```
killergerbah: No need to check for `target.start !== undefined`, `start` is already a `number` and must be defined

common/app/components/SubtitlePlayer.tsx
```diff
@@ -693,37 +693,53 @@ export default function SubtitlePlayer({
...
+            if (s.originalStart === jumpToSubtitle.originalStart && jumpToSubtitle.text.includes(s.text)) {
+                return i;
+            }
+            ++i;
+        }
+
+        return -1;
+    }, [jumpToSubtitle, subtitles]);
+
+    useEffect(() => {
```
killergerbah: Was there a reason you put this seek call in a separate effect? Seems like you could just put it in the existing one that's setting `highlightedJumpToSubtitleIndex`

.gitignore
L-M-Sherlock: Sure!

common/app/components/SubtitlePlayer.tsx
L-M-Sherlock: I split it into a separate effect to avoid re-seeking when hidden toggles (the previous combined effect caused a seek whenever the subtitle list visibility changed). If you’re OK with it, I can fold it back into the existing effect and gate the seek so it only runs on jumpToSubtitle changes.

L-M-Sherlock:
OK, now it works well in my test.

common/app/components/SubtitlePlayer.tsx
killergerbah: Oh got it. Yeah if thats easy to do then I think that would be preferred, but not a big deal. It just feels strange to be calculating the subtitle index twice.

L-M-Sherlock:
I fixed a small bug to avoid jumping when loading new subtitle file.

PR #886, opened 2026-02-16
Fix update last preserve html newlines

L-M-Sherlock:
Problem:

  - When updateLastCard runs, multiline field values are converted from \n to `<br>`.
  - Yomitan definitions include HTML (e.g., <style> blocks).
  - The conversion breaks CSS inside <style>, producing `<br>` between CSS rules.

  Fix:

  - Detect HTML content in field values and skip newline-to-`<br>` conversion for those values.
  - Plain-text multiline fields are still converted to `<br>` as before.

Before:

image

After:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ yarn-error.log*
 .pnp.*
 
 .idea/
+
+# local binaries
+/scripts/web-socket-server/asbplayer-ws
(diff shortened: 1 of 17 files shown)


killergerbah:
So basically, "if already HTML do not try to convert to HTML"
I think that makes sense, just slightly worried about unintended side effects. HTML can be inherited by the subtitle farther up in the file but this should only be coming from Yomitan, which should have inserted the breaklines already anyway.

killergerbah (approved):

PR #887, opened 2026-02-18
Player: add notifications when changing speed & offset

artjomsR:


common/app/components/VideoPlayer.tsx
```diff
@@ -542,11 +548,21 @@ export default function VideoPlayer({
         );
     }, []);
 
+    useEffect(() => {
+        playModeRef.current = playMode;
```
killergerbah: `useEffect` is not necessary for this, can just right `playModeRef.current = playMode` directly without `useEffect`

killergerbah (commented):
Thanks, seems like this has been missing for a while

common/app/components/VideoPlayer.tsx
artjomsR: thanks :)

killergerbah (approved):
Awesome thanks

Issue #742, opened 2025-07-08
Ability to toggle subtitle timing features per subtitle track


StoannP:
I would also like to submit a request for the ability to toggle subtitle timing features per subtitle track, or restrict it to only the first track, as mentioned in https://github.com/killergerbah/asbplayer/issues/402 and https://github.com/killergerbah/asbplayer/issues/385.

Right now if you have autopause enabled and two different subtitles loaded, the video pauses twice for what is usually the same sentence. The same goes for seek to previous/next subtitle.

killergerbah:
Linking this so I don't forget, since it's rleated https://github.com/killergerbah/asbplayer/issues/674

killergerbah:
Resolved by https://github.com/asbplayer/asbplayer/pull/890

Issue #402, opened 2024-03-31
Auto pause does not behave well with more than one subtitle track loaded


killergerbah:
If two tracks are loaded simultaneously, then auto-pause will trigger for each track. A possible solution is to intelligently guess whe subtitles on two different tracks represent the same line of dialog.

killergerbah:
Related to https://github.com/killergerbah/asbplayer/issues/385
It might be worth considering having a setting that determines which tracks should be considered for subtitle-timing-dependent features.

killergerbah:
Resolved by https://github.com/asbplayer/asbplayer/pull/890

Issue #385, opened 2024-03-12
Seek to previous or next subtitle line takes into account both imported subtitles.


avc1657:
I've observed that when I import both Japanese and English subtitles, the seeking function takes into account both, even though the primary subtitle is the Japanese one. For instance, during the opening song where there's an English translation but no Japanese transcription, pressing the hotkey to jump to the next subtitle line doesn't go to the next Japanese line. Instead, it goes through the English secondary subtitle since it contains the translation of the opening. Shouldn't the seeking be focused on the main subtitle?

killergerbah:
Maybe, some people will want to seek through one track or both tracks. Fixing this problem for everyone would probably mean implementing a setting somewhere that controls which subtitle tracks are navigated via the seek binds.

killergerbah:
Resolved by https://github.com/asbplayer/asbplayer/pull/890

PR #890, opened 2026-02-21
feat: setting which tracks should be considered for seek and autopause

rodrigo-suarezmajor:
Added a setting for deciding which subtitles track should be "seekable" through keyboard shortcuts as well as auto-paused on, as mentionend in #742 #385 #402 

Default is to consider only the first track, which imo should be the expected behavior. I didn't manage to implement it in the other playback modes easily, there the previous behavior remains.

Screenshot 2026-02-21 at 21 24 14


killergerbah:
Thanks for this, will review once 1.14 is out

common/src/model.ts
```diff
@@ -310,3 +310,5 @@ export enum ControlType {
     subtitleOffset = 1,
     playbackRate = 2,
 }
+
+export type SeekableTracks = [boolean, boolean, boolean]; // [firstTrack, secondTrack, thirdTrack]
```
killergerbah: Any reason why you picked a list of booleans vs a list of numbers? At first thought a list of numbers would generally be more compact, easier to understand and extensible, even if we never extend the list of configurable tracks beyond 3.

common/subtitle-collection/subtitle-collection.ts
```diff
@@ -62,7 +62,7 @@ export class SubtitleCollection<T extends SubtitleModel> {
         return SubtitleCollection.emptySubtitleCollection as SubtitleCollection<S>;
     }
 
-    subtitlesAt(timestamp: number): SubtitleSlice<T> {
+    subtitlesAt(timestamp: number, seekableTracks?: SeekableTracks): SubtitleSlice<T> {
```
killergerbah: I can see why `seekableTracks` is a convenient parameter but from the perspective of just this class it's not completely obvious why `seekable` should have the same meaning as `showing`.

Instead of having this parameter I suggest keeping the track seekability logic outside of this class. If it feels like you're repeating a lot of code then it can be extracted into a function.

extension/src/controllers/subtitle-controller.ts
```diff
@@ -102,6 +103,7 @@ export default class SubtitleController {
     refreshCurrentSubtitle: boolean;
     _preCacheDom;
     dictionaryTrackSettings?: DictionaryTrack[];
+    seekableTracks: () => SeekableTracks;
```
killergerbah: All the other subtitle settings are exposed as public fields, curious why you went with a function just for this one?

killergerbah (commented):
Really appreciate the PR, fixes a pretty bad issue. Left some minor comments.

Sima-Qian-TaiShiGong:
Hello, this feature is amazing, I've waited for this for a few years now.

I don't see it yet in the extension settings though, perhaps it isn't released yet ?

killergerbah (approved):
Thanks @rodrigo-suarezmajor for your initial work
I've taken over the PR and merging as this feature is highly requested

christopherball:
As of version 1.16.0, I do not see this feature merged.

image

christopherball:
Since this is marked merged, not sure if my last comment will notify folks, so tagging. @killergerbah @rodrigo-suarezmajor

ShanaryS:
Merged doesn't mean released, this is in the v1.17.0 milestone. New features will never show up in an already released version, not just for this project but any software.

Also github will still notify for new comments post merge.

christopherball:
> Merged doesn't mean released, this is in the v1.17.0 milestone. New features will never show up in an already released version, not just for this project but any software.

I didn't realize how recent the merge was.  You could have simply stated it was coming as part of v1.17.0 without the additional commentary.  Thanks.

ShanaryS:
My bad, I assumed you were new to GitHub instead of assuming you commented without reading.

PR #891, opened 2026-02-22
feat: add configurable mining screenshot timing

khajiitvaper2017:
### Description
Often the beginning of a subtitle corresponds to an image of the previous speaker or scene. In anime especially, scenes tend to start from the feet and pan up to the head.

To address this, I added configurable timing for mined subtitle screenshots with the following options:
- Beginning (default)
- After beginning (1 second delay, or 30% of duration if the subtitle is too short)
- Middle
- Before ending (1 second earlier, or 30% of duration if the subtitle is too short)
- Ending

### Changes
**Adds configurable timing for mined subtitle screenshots and wires it through the full mining pipeline.**
- Introduces a new setting, `miningScreenshotCaptureTimestamp`, with options ranging from subtitle start to end.
- Adds a shared timestamp helper and applies it in both the app (`VideoPlayer`) and the extension mining flow.
- Updates screenshot delay handling so captures align with the selected subtitle timestamp.
- Adds the new setting to defaults, the settings import/export schema, and tests.
- Adds a Mining settings UI selector and English locale strings for the new options.

### DIFF
--- a/common/app/components/VideoPlayer.tsx
+++ b/common/app/components/VideoPlayer.tsx
@@ -28,6 +28,7 @@ import {
     DictionaryTrack,
     TokenState,
     ApplyStrategy,
+    MiningScreenshotCaptureTimestamp,
     miningScreenshotTimestamp,
 } from '@project/common/settings';
 import {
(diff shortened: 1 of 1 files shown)


killergerbah:
This change breaks the existing "image screenshot delay" setting and replaces it with `miningScreenshotCaptureTimestamp`. While it provides some additional functionality I'm not sure it's a strict upgrade, as there is less flexibility for controlling the actual delay value. So I'm hesitant to review this change as-is.

khajiitvaper2017:
"image screenshot delay" was gated behind extension, so I moved it to Mining tab

killergerbah:
Screenshot delay was under the streaming video section because the feature only makes sense for streaming video where playing the subtitle being mined is part of the normal mining UX. This isn't true for the local video player where screenshots are taken instantly from the timestamp the user is at. What are you addressing by moving "image screenshot delay" out of the streaming video section?

khajiitvaper2017:
> This isn't true for the local video player where screenshots are taken instantly from the timestamp the user is at. 

This is false, I think? It takes screenshot from the begging of the subtitle for me.
I mine using this button and it doesn't use this delay.
image

Maybe, I am missing something since I use only local player.

I want to press this button and have screenshot either using delay or, with this feat, from the end of a spoken line.

killergerbah:
> This is false, I think? It takes screenshot from the begging of the subtitle for me.
> I mine using this button and it doesn't use this delay.

I see, yeah you're right. When you use that button it takes screeshot from the beginning of the subtitle.

Would this address your use-case?

- Instead of introducing a new setting, just allow negative values for the screenshot capture delay. For example, -1000 could mean "1000ms from the end of the subtitle."
- Also apply screenshot capture delay to local videos <-- I think this is what you've already started doing

With this type of configuration it feels like we are doing more with less. Same setting, just additional and more flexible behavior.

khajiitvaper2017:
created separate PR #897

Issue #840, opened 2025-12-08
API - Get Currently Loaded Subtitle


bpwhelan:
## **Is your feature request related to a problem? Please describe.**
There are certain applications that interface with asbplayer, but need to load the subtitle independently to work properly. One such example is https://github.com/mcgrizzz/Yomine

## **Describe the solution you'd like**
A simple http endpoint to grab the currently loaded subtitle. Pretty much the same thing as "Download Subtitles as SRT" option, but via HTTP.

## **Describe alternatives you've considered**
At the moment you can just download via the button, which is fine, an http endpoint would just be a 100% convenience thing.

## **Additional context**
I may or may not work on this if it's even possible in the first place. Should just be an HTTP endpoint and websocket message/reply. Please let me know if this would be denied if PR'd.

mcgrizzz:
+1

One thing to keep in mind: asbplayer can have multiple tabs active at the same time (e.g. Crunchyroll in one tab and YouTube in another, each with its own subtitles on the video element). The sidebar can only ever show one track at a time, though.

Because of that, I could see two possible scopes for this endpoint:

- **Sidebar-scoped:** the endpoint always returns the subtitles currently loaded in the sidebar (simple but may cause confusion if a user is going back and forth between tabs) 
- **Tab/video-scoped:** the endpoint takes some kind of tab/video identifier so clients can ask for the subtitles for a specific active tab.

In the latter case, there'd probably need to be a convenience method to grab a list of active tabs. 

Happy to help with this.

killergerbah:
I think whatever you do it should match the `load-subtitles` logic for consistency. By default it targets the active tab and assumes there's only one video element on the tab. That logic may need to be improved.

mcgrizzz:
I've been thinking about this a bit more after being away for a bit. 

The easiest implementation would be to mirror `load-subtitles` behavior. If no target is specified, it uses the current/default active-tab behavior.

But, as **killergerbah** hinted at, there is probably room for improvement on this behavior.

My concern is ambiguity. If there are multiple tabs/video elements, it might be useful to also have a WebSocket command like `get-targets` / `get-video-targets` that returns the known targets and whether each one has subtitles loaded. It looks like you can easily grab this information from the tab registry. 

Then a caller could either use the default behavior or explicitly pass a target to `get-loaded-subtitles`. 

This also opens the door for the `load-subtitles` command to be improved and allow targeting specific videos.

So for the general flow I’m imagining this:

[...]

PR #893, opened 2026-02-22
Feat: Chinese character converter

Luc-Mcgrady:
This would probably have been better as a git issue than as a pull request; but I was wondering if you would be ok with/interested in an option to automatically convert Chinese characters between simplified and traditional.

image

image
image

This change seems to increase the size of the extension by >1mb which isn't fantastic...

Anyway let me know what you think. If it's ok with you I will try to make a UI for it.

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -35,6 +35,7 @@
         "i18next": "^25.3.2",
         "m3u8-parser": "^7.2.0",
         "mpd-parser": "^1.3.1",
+        "opencc-js": "^1.0.5",
         "react": "^19.1.0",
         "react-dom": "^19.1.0",
         "react-i18next": "^15.6.1",
@@ -45,6 +46,7 @@
     },
     "devDependencies": {
         "@types/jest": "^30.0.0",
+        "@types/opencc-js": "^1.0.3",
         "@wxt-dev/module-react": "^1.1.5",
         "jest": "^30.0.5",
         "ts-jest": "^29.4.0",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
         ]
     },
     "devDependencies": {
-        "@types/opencc-js": "^1.0.3",
         "@typescript-eslint/parser": "^8.38.0",
         "eslint": "^9.32.0",
         "eslint-plugin-react": "^7.37.5",
@@ -25,8 +24,5 @@
     "packageManager": "yarn@3.2.0",
     "resolutions": {
         "lamejs@1.2.0": "patch:lamejs@npm:1.2.0#.yarn/patches/lamejs-npm-1.2.0-b0315d05aa.patch"
-    },
-    "dependencies": {
-        "opencc-js": "^1.0.5"
     }
 }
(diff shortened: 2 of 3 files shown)


killergerbah:
Appreciate the PR but hesitant to put this in for these reasons:
- This is a language-specific feature while currently asbpayer has little awareness of the user's target language. I think something like this makes more sense once we have more language-specific features.
- It would be convenient for asbplayer to convert the characters but it feels like asbplayer is providing very little value here other than making that function call. It might make more sense or as some kind of external plugin that converts subtitle files and loads them via asbplayer's WebSocket interface.
- As you said it increases the size of the bundle significantly. If not for this then I'd lean closer towards just including it.

Luc-Mcgrady:
Makes sense to me. Thanks for the feedback.

> It might make more sense or as some kind of external plugin that converts subtitle files and loads them via asbplayer's WebSocket interface.

I'd like this but my current problem would be that it is quite cumbersome to download the subtitle files, for example, with youtube-dl.

Maybe there could be some way to download the currently selected subtitle to re-upload a modified version via the websocket? Maybe there could even be a way for the websocket to announce that a new subtitle has been loaded, then any interested listeners could make the expensive call to fetch the subtitles.

Is this something you might be keen on including in the `scripts` folder?

> As you said it increases the size of the bundle significantly. If not for this then I'd lean closer towards just including it.

I could try and make it so that the user uploads whatever dictionary they would like to use. I would imagine this would solve the size problem.

Anyway, If all else fails I can just keep using my fork :)

killergerbah:
>Maybe there could be some way to download the currently selected subtitle to re-upload a modified version via the websocket? Maybe there could even be a way for the websocket to announce that a new subtitle has been loaded, then any interested listeners could make the expensive call to fetch the subtitles.

Actually this seems like a really great use case for the websocket server. We could provide a way to query the currently loaded subtitles. Then anybody interested could just poll for changes and transform the subs however they wish.

killergerbah:
If we do this i'd prefer to extend the anki plugin that @mwojick wrote though

Luc-Mcgrady:
> Actually this seems like a really great use case for the websocket server. We could provide a way to query the currently loaded subtitles. Then anybody interested could just poll for changes and transform the subs however they wish.

Just going to check, do you mean the file or the currently displayed subtitles. Would the web-socket from the client send a message to the server only once when the subtitles are loaded, or once per new subtitle node is displayed (and have the subtitle text updated as the video is playing)?

> If we do this i'd prefer to extend the anki plugin that @mwojick wrote though

So we keep a constant store of the current subtitles in the anki plugin, let it be polled by third party apps, and then when the subtitles are updated send them back to the client?

killergerbah:
> Just going to check, do you mean the file or the currently displayed subtitles. Would the web-socket from the client send a message to the server only once when the subtitles are loaded, or once per new subtitle node is displayed (and have the subtitle text updated as the video is playing)?

I was just thinking of providing a way to query the current subtitle file and its contents via an HTTP endpoint. Pushing subtitle events could be cool too but currently the only supported client of the websocket server is asbplayer itself. Not sure how complicated it would be to support other types of clients.

> So we keep a constant store of the current subtitles in the anki plugin, let it be polled by third party apps, and then when the subtitles are updated send them back to the client?

Yeah something like that but unsure of the details atm

killergerbah:
Looks like @bpwhelan created an issue for this already: https://github.com/killergerbah/asbplayer/issues/840

PR #897, opened 2026-02-23
Screenshot delay

khajiitvaper2017:
## Changes
- Moves `Screenshot Capture Delay` from `Streaming Video` settings to `Mining` settings, so that local player users can access it
- Allow negative numbers for delay (takes screenshot from the end + delay)
- Delay now works with local media
- Updated description

I have not tested extension.


killergerbah:
Great work, thank you

killergerbah (approved):

Issue #444, opened 2024-06-21
Blur widget for hardcoded subtitles


cristian428:
I understand there's a planned update to add a blur for subtitle tracks and I think that's great but something I've thought would be very useful before is having a dedicated blur widget to deal with hardcoded subs. Especially with Chinese language content, there's often hardcoded subs. In today's case I'm watching a show with subs in both ZH and EN. English subtitles are not wanted so having the option to hide them would be great. On windows PC there's a cool program that lets you put a blur bar over windows and I do use that on PC when needed but there's no equivalent of that for phones and tablets that I know of. At least for me, hardcoded subs can be very distracting.

![Screenshot_20240621_235131_YouTube](https://github.com/killergerbah/asbplayer/assets/36088384/41d49a41-aeb7-4577-a8e0-0c7dafa28db7)

artjomsR:
I wonder if this can be achieved to a satisfactory degree using the existing blur settings.

- load your subtitles twice, one displayed at the top, other - at the bottom
- Top subtitles are your actual subtitles that you want. Configure as needed
- Bottom subtitles are your blur widget. Enable the blur setting for them, increase the font size (increasing the "blur widget's" size as a result) and adjust the spacing at the bottom as needed

If you find that hardcoded subs are still occasionally visible, then it must be that the timings on hard and soft subs don't match. So you could create a sub file which would display a long subtitle throughout the whole video, e.g.
``` blur.srt
WEBVTT

1
00:00:00.000 --> 99:99:99.000
[...]

PR #898, opened 2026-02-25
feat: Add blur overlay option

saifkaral:
https://github.com/user-attachments/assets/eba4d0ef-2a36-4f4f-827c-25174ff4fec0

This feature has been requested for a long time, and it’s something I would personally use. It applies a blur over the video currently playing in the player.

Why is this useful? Some videos come with hardcoded subtitles that can’t be disabled. If a user wants to use their own subtitles instead, the blur can be applied to the area where the hardcoded subtitles appear, making them unreadable while keeping the rest of the video clear.

As shown in the demo video, this allows users to rely on their preferred subtitle track without the distraction of overlapping text.

it fixes this issue: https://github.com/killergerbah/asbplayer/issues/444


killergerbah:
Thanks for doing this
I played around with the overlay and it has some minor UI/UX quirks:
- If I click and drag, the handles disappear after a couple seconds, while I'm dragging. Feels a bit unintuitive - I would expect the handles to either be visible or invisible the whole time.
- If I click outside of the blur overlay, the handles don't disappear.
- It feels like the top/left and bottom/right handles don't need to be handles - they might as well take up the whole edge of the box so that they're easier to grab.

saifkaral:
> Thanks for doing this I played around with the overlay and it has some minor UI/UX quirks:
> 
> * If I click and drag, the handles disappear after a couple seconds, while I'm dragging. Feels a bit unintuitive - I would expect the handles to either be visible or invisible the whole time.
> * If I click outside of the blur overlay, the handles don't disappear.
> * It feels like the top/left and bottom/right handles don't need to be handles - they might as well take up the whole edge of the box so that they're easier to grab.

I addressed the issues you mentioned.

For the handles on the top/bottom/left/right edges though, I didn’t extend them to cover the entire edge because I felt it looked worse visually. Instead, I increased their size so they’re easier to grab.
They’re also now dynamic so they resize depending on the size of the blur box to avoid overlapping with the other handles.

common/app/components/BlurOverlay.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    // Clicking the overlay restores handles if they were hidden
+    const handleMouseClick = useCallback(() => {
+        setShowHandles(true);
+        resetAutoHideTimer();
+    }, [resetAutoHideTimer]);
+
+    return (
+        <div
+            data-blur-overlay="true"
```
killergerbah: What was the reason for using a data attribute, rather than using a `ref`?

common/app/components/BlurOverlay.tsx
```diff
@@ -0,0 +1,258 @@
...
+    const classes = useStyles();
+
+    // Position/size of the overlay within the container
+    const [pos, setPos] = useState<Position>({ x: 50, y: 300 });
+    const [size, setSize] = useState<Size>({ width: 400, height: 80 });
+    const [showHandles, setShowHandles] = useState<boolean>(true);
+
+    // Mutable refs used during drag/resize so we don't re-render on every mousemove
+    const dragging = useRef(false);
+    const resizing = useRef<ResizeDir | null>(null);
```
killergerbah: Nitpick, but intuitively would expect `dragging` and `resizing` to be the same type `boolean` but since they're not maybe there's a better name for `resizing`.

common/app/components/BlurOverlay.tsx
```diff
@@ -0,0 +1,258 @@
...
+        // Start auto-hide timer and wire global listeners for drag/resize
+        resetAutoHideTimer();
+
+        const onMouseMove = (e: MouseEvent) => {
+            const dx = e.clientX - dragStart.current.x;
+            const dy = e.clientY - dragStart.current.y;
+
+            if (dragging.current) {
+                setPos(() => {
+                    const nx = posStart.current.x + dx;
```
killergerbah: Why a function instead of just setting the value?

common/app/components/BlurOverlay.tsx
```diff
@@ -0,0 +1,258 @@
...
+    const [size, setSize] = useState<Size>({ width: 400, height: 80 });
+    const [showHandles, setShowHandles] = useState<boolean>(true);
+
+    // Mutable refs used during drag/resize so we don't re-render on every mousemove
+    const dragging = useRef(false);
+    const resizing = useRef<ResizeDir | null>(null);
+    const dragStart = useRef<Position>({ x: 0, y: 0 });
+    const posStart = useRef<Position>({ x: 0, y: 0 });
+    const sizeStart = useRef<Size>({ width: 0, height: 0 });
+    const lastActivityTimestamp = useRef<number>(Date.now());
```
killergerbah: As this is only used in `resetAutoTimer` a ref doesn't seem necessary to store this value

common/app/components/BlurOverlay.tsx
```diff
@@ -0,0 +1,258 @@
...
+        [pos, size, resetAutoHideTimer]
+    );
+
+    useEffect(() => {
+        // Start auto-hide timer and wire global listeners for drag/resize
+        resetAutoHideTimer();
+
+        const onMouseMove = (e: MouseEvent) => {
+            const dx = e.clientX - dragStart.current.x;
+            const dy = e.clientY - dragStart.current.y;
```
killergerbah: The mouse position should be clamped by the container bounds. Otherwise, if mouse is dragging off the container, then the overlay will continue to grow in the opposite direction.

saifkaral:
Sorry this took forever. I didn’t get a notification for your reply, so I genuinely thought I was getting ghosted 😭, and then I completely forgot about it. Just pushed changes to address the comments.

saifkaral:
was trying to test oops and then pushed to the wrong branch

killergerbah:
Thanks for the updates, I was trying the feature again and it looks like it doesn't really handle full screen transitions well, or showing/hiding the subtitle list. The overlay doesn't reposition relative to the size of the video. I think this should be fixed, since those are relatively common things for a user to do and it would be painful to have to reposition the overlay every time.

saifkaral:
The feature works great now. It was a difficult change, honestly. I want to note that I used AI quite a bit in this commit, but I reviewed the code carefully and it looks good to me.

saifkaral:
https://github.com/user-attachments/assets/83378d8d-730d-46e4-9b2c-33cb9ad521c0

killergerbah:
Great work, thanks for being patient

PR #912, opened 2026-03-01
feat: Align local video/subtitle split by source aspect ratio on drop and theater toggle

khajiitvaper2017:
This PR improves local playback layout when dropping a video + subtitle file.

On drop, the app reads source video metadata and initializes the split so the video pane matches the video aspect ratio as closely as possible, maximizing visible video area and reducing black side bars.

The same one-time recalculation also runs when Toggle Theater Mode changes layout (app bar hidden/shown), so split initialization stays correct after theater mode transitions.

After initialization, manual subtitle panel resizing remains user-controlled.

https://github.com/user-attachments/assets/fa5d7a97-0529-4897-9fa1-f299775fac48


killergerbah:
Thanks for doing this but it's not really obvious to me that this behavior would benefit everyone. I think some users would like the subtitle list to be a certain size and for the app to remember that. I'm open to alternate solutions if you can think of any.

khajiitvaper2017:
> Thanks for doing this but it's not really obvious to me that this behavior would benefit everyone. I think some users would like the subtitle list to be a certain size and for the app to remember that. I'm open to alternate solutions if you can think of any.

Add setting like "Video-Subtitle Split Behavior" with options: "Remember my split position" or "Auto-maximize video"?

killergerbah:
>Add setting like "Video-Subtitle Split Behavior" with options: "Remember my split position" or "Auto-maximize video"?

I see, yeah a setting feels like the natural way to do this. My preference would be to default to "Remember my split position".

Only other alternative I can think of is to add a new button to the video player or subtitle list but I don't know if there's a suitable icon for the button and where it makes sense to put it.

khajiitvaper2017:
> > Add setting like "Video-Subtitle Split Behavior" with options: "Remember my split position" or "Auto-maximize video"?
> 
> I see, yeah a setting feels like the natural way to do this. My preference would be to default to "Remember my split position".

Added the setting.

common/settings/settings.ts
```diff
@@ -7,8 +7,15 @@ export enum PauseOnHoverMode {
...
 
+export enum VideoSubtitleSplitBehavior {
+    rememberSplitPosition = 'rememberSplitPosition',
+    autoMaximizeVideo = 'autoMaximizeVideo',
+}
+
 export interface MiscSettings {
     readonly themeType: 'dark' | 'light';
+    readonly videoSubtitleSplitBehavior: VideoSubtitleSplitBehavior;
+    readonly subtitlePlayerWidth: number;
```
killergerbah: The split position was already being remembered by the app, within the same play session, but it looks like you've done more than that and are persisting the value to settings. As this is a purely visual, client-specific value, which shouldn't be affected by settings profiles, I would suggest either persisting this value using `PlaybackPreferences` instead, or simply keep the original behavior.

common/app/components/Player.tsx
```diff
@@ -232,6 +240,27 @@ const Player = React.memo(function Player({
     const classes = useStyles({ appBarHidden, appBarHeight });
     const calculateLength = () => trackLength(channelRef.current, subtitlesRef.current);
 
+    useEffect(() => {
```
killergerbah: Is it possible to avoid loading the video element unless it's necessary? The aspect ratio is not required if the setting is `remember my split`. One idea would be to move this into a hook exported from `video-subtitle-split`.

khajiitvaper2017:
fixed scrollbars appearing on video player when toggling theater mode

common/app/components/subtitles.css
```diff
@@ -1,3 +1,18 @@
+html,
+body {
```
killergerbah, line 1: Why was this added to `subtitles.css`?

khajiitvaper2017: When theater mode was toggled back on, the iframe document was ending up a bit taller than the viewport, so the browser showed a scrollbar. This problem is also present on live app, This fixes it.

khajiitvaper2017: Though it's mostly visible when using `yarn run start`, when using `preview`, it's rare

khajiitvaper2017: I kept only one `overflow: hidden;`, should be enough to fix this.

killergerbah: Can this be addressed in a separate change then? It doesn't seem critical to fix, if it is only dev-facing. Also, I'm not completely convinced it's the right fix, if it's really the iframe that's taller then setting overflow to `hidden` would just hide the part of the iframe that's overflowing.

khajiitvaper2017: Sure, feel free to remove it, just didn't seem to be worthy of separate pull request. In my first post's video you can see scrollbars appearing for a split second when toggling theater mode, that's what I am addressing. On live version this problem also exists but layout changes faster than scrollbar can appear and it jerks a little instead.

https://github.com/user-attachments/assets/4eefbf4f-c724-42ef-8969-77b406299d1c


I tried to log it and for some reason, htmlScrollHeight is taller than viewport:
bodyClientHeight = 847
bodyScrollHeight = 847
htmlClientHeight = 847
htmlScrollHeight = 865
If you can find anything better than just slapping `overflow: hidden;`, that would be great.

killergerbah:
I was looking into merging this just now but the auto maximize apparently only works the first time a video is loaded

khajiitvaper2017:
> I was looking into merging this just now but the auto maximize apparently only works the first time a video is loaded

uhhh, it works on my machine...

khajiitvaper2017:
I am using chrome

killergerbah:
It looks like this check
```
        if (appliedInitialWidthRef.current === clampedInitialWidth) {
            return;
        }
```
is preventing the video from maximizing on my computer, because my screen is usually too small and `clampedInitialWidth` is always 0. So the maximize behavior only applies the first time any sufficiently wide video is loaded.

common/app/components/SubtitlePlayer.tsx
```diff
@@ -1012,18 +1015,35 @@ export default function SubtitlePlayer({
...
-        const listener = () => {
-            lastKnownWidth = undefined;
-            setWidth(calculateInitialWidth());
-        };
-        screen.orientation.addEventListener('change', listener);
-        return () => screen.orientation.removeEventListener('change', listener);
-    }, [setWidth]);
+        // Small delay to allow layout to settle
+        const timer = setTimeout(() => {
+            scrollToCurrentSubtitle();
```
killergerbah, line 1031: Why was this additional scroll-to-subtitle added? It doesn't seem like obviously better behavior.

khajiitvaper2017: after resize it would go to the beginning or end of the subtitle list, so I added this

khajiitvaper2017: there is auto scroll on switching tabs, so I guess, trying to keep some specific position which is not a current subtitle isn't a priority

killergerbah: Ok, thanks for the explanation

killergerbah:
Appreciate the addition @khajiitvaper2017

killergerbah:
I believe the new setting can be presented with fewer words so I will make that change after merging

PR #913, opened 2026-03-02
fix: captureStream usage in Firefox 149

mwojick:
From https://www.firefox.com/en-US/firefox/149.0beta/releasenotes/:

"Enabled the spec-compliant [HTMLMediaElement.captureStream()](https://w3c.github.io/mediacapture-fromelement/#html-media-element-media-capture-extensions) API."

Practically speaking this means that the new captureStream API is now independent of the video's volume, which means that we have to manually map the video's volume to the MediaStream using a gain node, and keep it up to date when the user adjusts the volume.

### DIFF
--- a/extension/src/entrypoints/sidepanel/index.html
+++ b/extension/src/entrypoints/sidepanel/index.html
@@ -3,7 +3,7 @@
     <head>
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
-        <meta name="manifest.open_at_install" content="true" />
+        <meta name="manifest.open_at_install" content="false" />
         <meta
             name="manifest.default_icon"
             content="{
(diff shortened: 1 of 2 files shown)


extension/src/entrypoints/sidepanel/index.html
```diff
@@ -3,15 +3,15 @@
     <head>
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
-        <meta name="manifest.open_at_install" content="true" />
+        <meta name="manifest.open_at_install" content="false" />
```
mwojick, line 6: This is an unrelated change but I found that this causes the sidepanel to reopen every time the extension is refreshed, not just on initial install. Doesn't seem to be easy to disable this just in dev mode, but I don't think it's a big deal to disable it in general.

PR #915, opened 2026-03-02
Blur ruby for extension subtitles if enabled

### LINKED ISSUE #899: [Bug] Subtitle blur does not apply to ruby text (word readings)
https://github.com/user-attachments/assets/c40aca5b-8718-4335-971e-11a5097a0419

ShanaryS:
fixes #899

### DIFF
--- a/extension/src/entrypoints/video.content/video.css
+++ b/extension/src/entrypoints/video.content/video.css
@@ -27,6 +27,7 @@
 }
 
 .asbplayer-subtitles-blurred {
+    display: inline-block; /* Blur ruby annotations as well */
     filter: blur(10px);
 }
 


extension/src/entrypoints/video.content/video.css
```diff
@@ -27,6 +27,7 @@
 }
 
 .asbplayer-subtitles-blurred {
+    display: inline-block; /* Blur ruby annotations as well */
```
killergerbah, line 30: Does some change need to be made in `video.css` for the extension?

ShanaryS: It already handled it correctly, the website that is. Only the extension had this issue.

killergerbah: Got it..  my brain is shrinking by the day...

PR #916, opened 2026-03-02
Reduce batch size for Anki and Yomitan

### LINKED ISSUE #900: New annotations struggle with big anki decks
im using vivaldi browser btw. i tried to connect my main anki deck to asb player but it just kept rebuilding the database for a while and then timed out. My anki deck is really big, i have 19k cards on it and it's using a pretty heavy note type with custom css and html so im guessing it just takes too long for asb player and anki to process so it just defaults to time out. it probably needs some optimizing for big decks

ShanaryS:
fixes #900

* It seems Anki decks that have complicated `css` dramatically increases the data being sent which breaks things along the way (or causes crashes). The deck size is not the issue but how much cards we batch at a time. This is only the `cardsInfo` endpoint which is only used for modified cards (all for first time build). This makes the process slower but is negligible since future syncing will be small and in the background.
* Added a new build message for the start of a build so users know that it's Anki being queried.
* Reduced the batching for Yomitan as well as we can trade off a bit of speed there for less RAM.

I'm not sure if these will fixes all the issues for everyone but these changes are necessary at least for some people. We can just design around the lowest common denominator as it comes up.

---

Did a deeper dive on the API results and the three problematic fields are `question`, `answer`, and `css`. These field accounted for `45kb` per card for me out of `45.5kb`. So for the users having issues, this is probably a big reason, wasted RAM causing crashes.

### DIFF
--- a/common/anki/anki.ts
+++ b/common/anki/anki.ts
@@ -206,16 +206,18 @@ export class DuplicateNoteError extends Error {
     }
 }
 
+// Optional fields are unused thus deleted to save memory
 export interface CardInfo {
-    answer: string;
-    question: string;
+    answer?: string;
+    question?: string;
     deckName: string;
     modelName: string;
     fieldOrder: number;
     fields: { [fieldName: string]: { value: string; order: number } };
-    css: string;
+    css?: string;
     cardId: number;
     interval: number;
+    factor: number;
     note: number;
     ord: number;
     type: number;
@@ -225,6 +227,8 @@ export interface CardInfo {
     lapses: number;
     left: number;
     mod: number;
+    nextReviews: [string, string, string, string];
+    flags: number;
 }
 
 export interface NoteInfo {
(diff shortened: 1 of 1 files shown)


killergerbah:
@ShanaryS Feel free to merge whenever you're ready. For what its worth I tried the branch before you pushed the field deletion change and I didn't notice any difference in speed of initial import at all.

ShanaryS:
@killergerbah Yeah I tested it a few days ago on my fast PC at it was a difference of `23s` to `27s`. I just tested it on a slower PC and it went from `2m` to `2m30s` so negligible. For 0 modified cards, it takes about `5s` to `7s` for both. This is all excluding the parsing of the fields, just the getting the notes and cards.

But I think the biggest difference will be in the memory saved as I think most people were simply running out of ram. Without these fields everyone's card should be a bytes to a few kb. We could also prune the fields not configured but we can save that for later if it's needed.

PR #918, opened 2026-03-03
Support token color alpha and ability to disable any status

ShanaryS:
This was reported in the discord to have the ability to ignore arbitrary status. Some users may want to treat Young+ as known without the underline or other styles appearing. I also added ability to set the alpha which is mainly useful for the background style.

### DIFF
--- a/common/settings/settings.ts
+++ b/common/settings/settings.ts
@@ -96,11 +96,37 @@ export function getFullyKnownTokenStatus(): TokenStatus {
     return TokenStatus.MATURE; // If future statuses are optional, this logic may need to change
 }
 
-// Any future field added will likely need to be optional for app/extension version mismatch and added to dictionaryTrackComparators
+// Any future field added will likely need to be optional for app/extension version mismatch
 export interface TokenStatusConfig {
     readonly display: boolean;
     readonly color: string;
     readonly alpha: string;
+}
+
+const tokenStatusConfigComparators: {
+    [K in keyof TokenStatusConfig]: (a: TokenStatusConfig[K], b: TokenStatusConfig[K]) => boolean;
+} = {
+    display: (a, b) => a === b,
+    color: (a, b) => a === b,
+    alpha: (a, b) => a === b,
+};
+
+export function compareTokenStatusConfigField<K extends keyof TokenStatusConfig>(
+    key: K,
+    a: TokenStatusConfig,
+    b: TokenStatusConfig
+): boolean {
+    return tokenStatusConfigComparators[key](a[key], b[key]);
+}
+
+export function areTokenStatusConfigsEqual(a: TokenStatusConfig, b: TokenStatusConfig): boolean {
+    if (a === b) return true;
+    for (const key in tokenStatusConfigComparators) {
+        if (!compareTokenStatusConfigField(key as keyof TokenStatusConfig, a, b)) {
+            return false;
+        }
+    }
+    return true;
 }
 
 export enum TokenState {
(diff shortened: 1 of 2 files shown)


common/settings/settings.ts
```diff
@@ -157,8 +157,10 @@ export interface DictionaryTrack {
     readonly dictionaryAnkiTreatSuspended: TokenStatus | 'NORMAL';
     readonly dictionaryTokenStyling: TokenStyling;
     readonly dictionaryTokenStylingThickness: number;
-    readonly dictionaryColorizeFullyKnownTokens: boolean;
+    readonly dictionaryColorizeFullyKnownTokens: boolean; // Deprecated in favor of dictionaryTokenStatusDisplays
+    readonly dictionaryTokenStatusDisplays: boolean[]; // Indexed by TokenStatus
     readonly dictionaryTokenStatusColors: string[]; // Indexed by TokenStatus (if adding colors for states, use a separate array indexed by TokenState)
+    readonly dictionaryTokenStatusAlphas: string[]; // Indexed by TokenStatus (if adding alpha for states, use a separate array indexed by TokenState)
```
killergerbah: Wonder if we should just bite the bullet and migrate into an array of objects instead of two additional arrays? Just need to make sure the migration is very safe.

ShanaryS: We would have to store it as a separate field regardless. The big problem is dealing with mismatched version between extension and app, so we will always need access to the current format to serve with.

I think it would make future changes to this (if any) easier as they could just be marked optional.

ShanaryS: I used the object as it makes it cleaner. The migration is handled by `ensureDictionaryTracksConsistency()` which adds the field if missing and populates with the default values (migrating the colors). It also prunes from the ends if there are too many in the array. This is checked on every read and write so it should migrate as well as ensure the array is always populated correctly.

killergerbah: Nice, yea I think we'll probably benefit from using an object in the long run.
Now I think you rightly pointed out that version mismatch is going to be the big problem. I think the worst combination is newer app + older extension.

common/settings/settings-provider.ts
```diff
@@ -433,7 +442,59 @@ const deepEquals = (a: any, b: any) => {
     return true;
 };
 
+const ensureDictionaryTracksConsistency = ({ dictionaryTracks }: Partial<AsbplayerSettings>) => {
```
killergerbah, line 445: If app is newer than extension then I think this should work for reading the old settings. But when writing to old extension, the old field will not get the latest value, so the extension popup will not reflect the change. I think to do this completely we'll need some knowledge of what version of the schema the extension is actually on.

killergerbah: Might need to introduce a schema version number or something. Then when writing to an older schema, we'd need to double write both in new and old format. When writing to latest schema, we can just write in the new format and cleanup the old data.

common/settings/settings.ts
killergerbah: Now that I'm thinking about it, the complexity of the migration is high enough where just adding in the new fields might have been the better choice... I won't push back if you decide to go back to the original method.

common/settings/settings-provider.ts
ShanaryS: I think it should be in fine? I'm updating both the old and new fields when saving using either UI. I kept running into issues with my designs so I just essentially made `dictionaryTokenStatusColors` and `dictionaryColorizeFullyKnownTokens` the absolute truth in `ensureDictionaryTracksConsistency()`. This covers first time migrations and future updates (from any UI version) updates both fields. Even if they were to be swaping between different versions all the time it should be covered.

Essentially, `1.14` only used the old settings. `1.15` will write to both old and new regardless of version mismatch. `ensureDictionaryTracksConsistency()` makes sure the new settings always match the old settings.

common/settings/settings.ts
ShanaryS: I don't think the objects significantly change the complexity since I'm always updating the old and the new.

common/settings/settings-provider.ts
killergerbah: Oh ok got it. If double writing no matter what then all good.

ShanaryS: I just retested everything, the migration works fine and updating always works as expected regardless of any version mismatch.

killergerbah: Ok thanks, I was being too paranoid

killergerbah:
Awesome, thanks for doing this
I might adjust the UI a bit before we release. I think the switches might look better on the right, and if possible I want the color picker + alpha inputs to be more together and compact.

killergerbah (approved):

Issue #917, opened 2026-03-02
Exported files need underscores to be removed from the front of the filename


killergerbah:
From jtkehler on Discord 

>The way special characters are replaced in filenames causes most of them to start with an underscore, which means that Anki will not delete them when checking for unused media. For example `[SubsPlease] Seihantai na Kimi to Boku - 07 (1080p) [C004A6E2]` becomes `_SubsPlease__Seihantai_n_350092_y0xLWDQK.mp3`

mwojick:
The reason for this is due to: https://docs.ankiweb.net/templates/styling.html#add-font-to-media-folder

> Adding an underscore will tell Anki that this file will be used on a template, and should not be deleted when checking for unused media.

I think just adding a prefix like `asbp_` would work and would also allow filtering files created by asbp

killergerbah:
Nice yea that's a better solution

PR #919, opened 2026-03-03
prefix file names

mwojick:
closes https://github.com/killergerbah/asbplayer/issues/917

### DIFF
--- a/common/anki/anki.ts
+++ b/common/anki/anki.ts
@@ -595,7 +595,10 @@ export class Anki {
         // Sanitize unsafe URL characters for AnkiWeb compatibility.
         name = this._sanitizeUnsafeURLChars(name);
         // Sanitize for file system compatibility on various operating systems.
-        return sanitize(name, { replacement: replacement });
+        name = sanitize(name, { replacement: replacement });
+        // Prefix to allow filtering by asbplayer created files,
+        // and to prevent Anki from skipping cleanup of files that start with an underscore.
+        return 'asbp_' + name;
     }
 
     private async _storeMediaFile(name: string, base64: string, ankiConnectUrl?: string) {


common/anki/anki.ts
```diff
@@ -595,7 +595,10 @@ export class Anki {
         // Sanitize unsafe URL characters for AnkiWeb compatibility.
         name = this._sanitizeUnsafeURLChars(name);
         // Sanitize for file system compatibility on various operating systems.
-        return sanitize(name, { replacement: replacement });
+        name = sanitize(name, { replacement: replacement });
+        // Prefix to allow filtering by asbplayer created files,
+        // and to prevent Anki from skipping cleanup of files that start with an underscore.
+        return 'asbp_' + name;
```
killergerbah, line 601: Only reservation with this change, is that IIRC there was a file length limit on Android that caused me to make the file names really short, like <= 32 characters. With this change it might be possible to go over that limit. One way to work around this would be to add the `asbp_` prefix in `makeFileName` in `audio-clip.ts` and `image.ts` instead.

killergerbah: Or maybe reducing `maxPrefixLEngth` in those files by 5

mwojick: Does the problem on Android have to do with creating cards on Android? Or just viewing them on AnkiDroid? Seems like viewing is fine for me, but I've never tried making them on Android. Either way I think we would already have an issue if 32 was the limit since you have the prefix (24) plus the timestamp + random string (18)

mwojick: Actually I just tried loading it into Firefox Nightly on Android using web-ext (such a pain to setup) with Ankiconnect Android, and the audio file seems fine too (img still doesn't get captured on Android FF). Though there's a random number at the end instead of the random string. Does Ankiconnect create that and is that related to the issue?

killergerbah: >Does the problem on Android have to do with creating cards on Android? Or just viewing them on AnkiDroid?

It was the former

>Either way I think we would already have an issue if 32 was the limit since you have the prefix (24) plus the timestamp + random string (18)

Yeah tbh I don't remember what the actual limit was. It's possible I chose a very small number just to be very safe.

killergerbah: >Though there's a random number at the end instead of the random string. Does Ankiconnect create that and is that related to the issue?

Hmm yeah must be AnkiConnect by process of elimination. I suppose that would increase the filename length but if it's working for you I guess we're fine. A limit in the 40s feels extraordinarily small for an OS.

killergerbah: From Googling, the smallest number I could fine was 50. Does the filename you observed in AniiConnect less than 50?

mwojick: With the new prefix I'm getting in total between like 56-58 (excluding file ext) depending on the length of the timestamp. The number at the end (idk if it's a random num or what) is around 20 by itself. I searched around as well and heard that the limit can be between 63-255 depending on encoding but who knows. Maybe it was less on older Android versions too. Either way I'll leave it up to you

killergerbah: Sounds like we were already above 50 so the limit being more like 63 seems more plausible. Let's just yolo it and we can always lower it back down if it's a problem.

PR #923, opened 2026-03-05
Adjust maturity styling settings UI

killergerbah:
Adds a preview and arranges the UI elements so they line up a nice-feeling grid-like formation.

https://github.com/user-attachments/assets/83dfbc5f-cd27-476b-997b-3e21cca0b941


ShanaryS:
I went ahead and implemented the change. The readings and frequency are shown based on their settings. I opted to just use `Reading` as the text for all status, languages that have reading can localized it for each status name.

image

killergerbah:
Sweet, yea the readings and the frequency were a great addition. Now it's basically a complete preview.

Issue #574, opened 2024-11-27
Can we add a GIF Capture feature instead of SS


MuhammadShahzeb123:
If we will have GIF then it will add more fun learning the card we create. Static images are right but sometimes, you just want to know what's happening in the scenes when two people are talking to describe the meaning of a Kanji/Word with Context. 

**Is your feature request related to a problem? Please describe.**
NO

**Describe the solution you'd like**
So, when we press the capture Shortcut... Ctrl + Shift + X -> It should do the same things as before (get subtitles and open a dialog) but instead of taking a Screenshot (SS) we can make it capture GIF. If, for some reasons, GIF cannot be captured, we can take a short video in mp4 and convert into GIF all while hiding the subtitles shown by asbplayer;

**Describe alternatives you've considered**
Nope

**Additional context**
I don't know much of JS but if for some reason converting mp4 to GIF or recording mp4 video in sync is not  possible, what we can do is use any other extension like AwesomeScreenshot or something like that and use that to create a video and convert it into GIF

ありがとう

killergerbah:
A feature like this is likely very difficult but I acknowledge that this is a possible improvement.

MuhammadShahzeb123:
> A feature like this is likely very difficult but I acknowledge that this is a possible improvement.

I am working on it myself... The problem is that I cannot sink the capturing function with the video being played along with the subtitles... 

Let's see how it turns out

Issue #187, opened 2022-12-13
Request: capture low quality GIF from the video


YinnVI:
Can we have Feature to capture low quality GIF from the video

killergerbah:
It is possible, likely wouldn't come soon.

LeonelRFF:
Hello, in case this GIFs thing comes in the future, it could be replaced momentarily by taking more than one screenshot, in some cases, and trying to adjust the milliseconds so that the screenshot is taken at the precise moment, but it is like doing it blindly, if the function of taking 3,4 or more screenshots is added, it can be even more assured that the exact screenshot is taken, when mining a sentence in the pop-up window, those images that are not useful can be deleted, or leave them all, it can also be the case

PR #924, opened 2026-03-06
feat: support WebM media fragments in Anki exports for local video

khajiitvaper2017:
## Summary
At first I tried adding GIF capture (#574 #187). It worked, but GIFs were too heavy and took too long to generate, so I switched to WebM instead. WebM ended up working surprisingly well. I also parallelized video and audio capture, and since they take about the same amount of time, it doesn't really feel any slower in practice (I used a laptop disconnected from power and with power saving on for a week with this). I tested this mainly on the latest Chrome.

The main caveat is that the Anki note template needs to support the `<video>` tag.

- replaced the old image-only capture path with a generalized `MediaFragment` pipeline that supports both JPEG screenshots and generated WebM clips
- added app-side mining settings for capture format and clip trim start/end, while keeping the extension UI on the existing JPEG-only path for now
- updated Anki export, preview, download, and saved-state flows so media fragments can be rendered, downloaded, and exported as either images or looping WebM clips
- tightened JPEG/WebM rendering lifecycle handling around retries, cancellation, hidden-tab pauses, video readiness, and resource cleanup
- avoided doing media encoding work for `updateLast` when there is no recent note to update
- added tests

## Details
- introduced `common/src/media-fragment.ts` and split file-backed rendering into `jpeg-file-media-fragment-data.ts` and `webm-file-media-fragment-data.ts`
- added JPEG compression
- added automatic WebM codec selection, minimum clip duration enforcement, trim-range resolution, canvas capture / `MediaRecorder` handling, and fallback scheduling for browsers without `requestVideoFrameCallback`
- extended settings sync/import/export/defaults with `mediaFragmentFormat`, `mediaFragmentTrimStart`, and `mediaFragmentTrimEnd`
- updated `AnkiDialog`, preview components, and hooks to handle both still images and video previews; clipboard copy is still image-only
- updated Anki export to store WebM fragments as media files and embed them with `<video autoplay loop muted playsinline ...>`, while preserving compatibility with existing serialized `image` payloads

Assisted-by: Codex:gpt-5.3


killergerbah:
Hey sorry I'm dragging my feet on this. Will try to review carefully on the weekend.

common/anki/anki.ts
```diff
@@ -476,40 +505,27 @@ export class Anki {
...
-                    this._appendField(fields, this.settingsProvider.audioField, `[sound:${fileName}]`, false);
-                } else {
-                    params.note['audio'] = {
-                        filename: sanitizedName,
-                        data,
-                        fields: [this.settingsProvider.audioField],
-                    };
-                }
-            }
+        const recentNotes = updateLast ? await this.findNotes('added:1', ankiConnectUrl) : [];
```
killergerbah, line 519: Why did this line move out of the `updateLast` case below? With this change, the other export modes won't work unless a recent card is in the deck.

common/src/webm-file-media-fragment-data.ts
```diff
@@ -0,0 +1,963 @@
...
+            });
+
+            const blob = await this._stopAndFlushRecorder(mediaRecorder, stopRecorder);
+            if (!blob || blob.size <= 0) {
+                throw new Error('Could not encode WebM from local video');
+            }
+
+            mediaRecorder = undefined;
+            stopRecorder = undefined;
+            return blob;
```
killergerbah: The `blob` is coming from the `MediaRecorder`... what is the purpose of writing to the `canvas` in `runFrameLoop`? Unless I'm misreading this, it looks like only the `MediaRecorder` is required.

common/anki/anki.ts
khajiitvaper2017: The line moved out of the updateLast switch case so updateLast can fail before doing media encoding. Before you would waste 5 seconds recording audio and video before it tells you that there is no note to update. Other modes shouldn't be affected by this.

common/src/webm-file-media-fragment-data.ts
khajiitvaper2017: The Blob is produced by MediaRecorder, but MediaRecorder is recording the canvas stream, not the `<video>` element directly. runFrameLoop is what copies each decoded video frame onto the canvas and triggers frame emission, which is also how we resize to the target output dimensions before encoding. We could switch to video.captureStream(), but no resizing then. I tried it, there were no noticeable difference in encoding speed, though it removed 100 lines from this file. But 1080p webm is around 620 kb and 720p one is 190kb.

common/anki/anki.ts
killergerbah: Okay got it, I missed that `updateLast` is being checked

common/src/webm-file-media-fragment-data.ts
killergerbah: I see now, thanks

common/components/AnkiDialog.tsx
```diff
@@ -444,8 +451,46 @@ const AnkiDialog = ({
             return;
         }
 
-        setImage(Image.fromCard(card, settings.maxImageWidth, settings.maxImageHeight));
-    }, [card, open, settings.maxImageWidth, settings.maxImageHeight]);
+        setImage((previousImage) => {
+            previousImage?.dispose();
+
+            if (settings.mediaFragmentFormat === 'webm' && card.file) {
```
killergerbah: If you expose `endTimestamp` on `MediaFragment` then this branch may not be necessary, since similar logic is already in `MediaFragment.fromCard`.

common/components/MiningSettingsTab.tsx
```diff
@@ -214,6 +220,25 @@ const MiningSettingsTab: React.FC<Props> = ({ settings, onSettingChanged }) => {
...
+                    label="Capture format"
+                    value={mediaFragmentFormat}
+                    onChange={(event) =>
+                        onSettingChanged(
+                            'mediaFragmentFormat',
+                            event.target.value as AsbplayerSettings['mediaFragmentFormat']
+                        )
+                    }
+                >
+                    <MenuItem value="jpeg">JPEG screenshot</MenuItem>
```
killergerbah: This should be localized. I think it would be fine to simplify to simply "screenshot" and "video clip." Most people won't care about the encoding.

common/components/MiningSettingsTab.tsx
```diff
@@ -242,34 +267,72 @@ const MiningSettingsTab: React.FC<Props> = ({ settings, onSettingChanged }) => {
...
-                value={screenshotDelayInput}
-                color="primary"
-                onChange={(event) => {
-                    const value = event.target.value;
-                    setScreenshotDelayInput(value);
+            {showWebmMediaFragmentSettings && mediaFragmentFormat === 'webm' && webmCaptureSupported && (
+                <>
+                    <TextField
+                        type="number"
+                        label="Clip trim start"
```
killergerbah: Same here, should create a localization key for this

common/settings/settings-provider.ts
```diff
@@ -106,6 +106,9 @@ export const defaultSettings: AsbplayerSettings = {
     audioPaddingEnd: 500,
     maxImageWidth: 0,
     maxImageHeight: 0,
+    mediaFragmentFormat: 'jpeg',
+    mediaFragmentTrimStart: 200,
```
killergerbah, line 113: Trimming makes sense, but I wonder if it would be simpler and also safer to simply specify a max length instead, positioned around the center of the subtitle? It's possible for a subtitle file to have a really long subtitle for whatever reason, so would prefer a length-limiting setting that still works for most users.

common/src/media-fragment.ts
```diff
@@ -0,0 +1,441 @@
...
+        }
+
+        const sourceBlob = await this.blob();
+        const canvas = document.createElement('canvas');
+        const ctx = canvas.getContext('2d');
+        if (!ctx) {
+            throw new Error('Failed to create canvas context for PNG conversion');
+        }
+
+        if (typeof createImageBitmap === 'function') {
```
killergerbah: Seems like unnecessary backwards compatibility code, based on browser support for `createImageBitmap`

common/src/webm-file-media-fragment-data.ts
```diff
@@ -0,0 +1,963 @@
...
+
+        const canvas = this._canvas;
+        if (typeof canvas.captureStream !== 'function') {
+            throw new Error('WebM capture stream is not supported in this browser');
+        }
+
+        canvas.width = width;
+        canvas.height = height;
+
+        if (!this._ctx || this._ctx.canvas !== canvas) {
```
killergerbah: Is `this._ctx.canvas !== canvas` possible?
If not, would it make sense to avoid defensive code, since `_ctx` is private to this class?

common/src/webm-file-media-fragment-data.ts
```diff
@@ -0,0 +1,963 @@
...
+        });
+        const stopRecorder = new Promise<Blob>((resolve, reject) => {
+            recorder.onstart = () => {
+                resolveRecorderStarted?.();
+                resolveRecorderStarted = undefined;
+                rejectRecorderStarted = undefined;
+            };
+            recorder.onerror = (event) => {
+                const error = event.error;
+                const recorderError = error instanceof Error ? error : new Error('Could not encode WebM');
```
killergerbah: If `error instanceof Error` is false, would it make sense to still include `error` in `new Error(...)`, so we don't lose it in the logs?

common/src/webm-file-media-fragment-data.ts
```diff
@@ -0,0 +1,963 @@
...
+        const drawFrame = () => {
+            ctx.drawImage(video, 0, 0, width, height);
+            captureTrack?.requestFrame();
+        };
+        const frameSchedulerMode: FrameSchedulerMode =
+            typeof video.requestVideoFrameCallback === 'function'
+                ? 'video-frame-callback'
+                : typeof requestAnimationFrame === 'function'
+                  ? 'animation-frame'
+                  : 'timeout';
```
killergerbah: It might be reasonable for `requestVideoFrameCallback` to be missing but I don't think we need a fallback after `requestAnimationFrame`. I wouldn't be surprised if asbplayer doesn't even work on browsers that old.

common/src/webm-file-media-fragment-data.ts
```diff
@@ -0,0 +1,963 @@
...
+
+    private _prepareVideoForCapture(video: HTMLVideoElement) {
+        // Video element is owned by this class instance; capture can mutate playback state.
+        video.muted = true;
+        video.volume = 0;
+        video.playbackRate = 1;
+
+        const videoWithPreservesPitch = video as HTMLVideoElement & { preservesPitch?: boolean };
+        if (typeof videoWithPreservesPitch.preservesPitch === 'boolean') {
+            videoWithPreservesPitch.preservesPitch = false;
```
killergerbah: What is the reason for setting this, even though the volume is 0, and the playback rate is 1?

killergerbah (commented):
Impressive work

common/settings/settings-provider.ts
khajiitvaper2017: The main purpose of trimming is to remove abrupt transitions from one scene to another, and a max length would not help with this. Also, from my experience, long clips usually feel better when you are doing your cards since there is less abrupt movement when switching from the last frame to the first.

killergerbah: I see, but isn't the current metho vulnerable to very long subtitles? e.g. >1 minute or more. Would not happen often, but it is plausible

common/src/webm-file-media-fragment-data.ts
khajiitvaper2017: There isn’t a strong reason. That was probably left over from earlier experiments with non-1x playback during capture of gifs. Removed.

common/components/AnkiDialog.tsx
khajiitvaper2017: Done

common/settings/settings-provider.ts
khajiitvaper2017: Should we make max length hardcoded or a setting?

killergerbah: A setting would be ideal but for this purpose I think hardcoding it is also fine, as this will just prevent the app from automatically recording long clips accidentally. With a long limit (like 10 seconds) most users wouldn't notice.

khajiitvaper2017: added setting

common/src/media-fragment.ts
```diff
@@ -0,0 +1,444 @@
...
+        };
+    }
+
+    return {
+        startTimestamp: desiredStartTimestamp,
+        endTimestamp,
+    };
+};
+
+export const createVideoElement = async (blobUrl: string): Promise<HTMLVideoElement> =>
```
killergerbah, line 117: It seems that in moving the video element creation to this function, some logic got lost.

Originally, to create a video element for screenshot from a file, this logic existed:
```
    private async _videoElement(file: FileModel): Promise<HTMLVideoElement> {
        if (this._video) {
            return this._video;
        }

        return new Promise((resolve, reject) => {
            const video = document.createElement('video');
            video.src = file.blobUrl;
            video.preload = 'auto';
            video.autoplay = false;
            video.volume = 0;
            video.controls = false;
            video.pause();
            const t0 = Date.now();
            const interval = setInterval(() => {
                if (
                    (video.seekable.length > 0 && video.seekable.end(0) === video.duration) ||
                    Date.now() - t0 >= 5_000
                ) {
                    this._video = video;
                    clearInterval(interval);
                    resolve(video);
                }
            }, 100);
        });
    }
```
The `seekable` check was intended to make file screenshots more precise. Without this, the screenshot that gets taken is often very far from where the user intends. Can this logic be restored?

common/src/model.ts
```diff
@@ -78,6 +78,7 @@ export interface CardModel extends CardTextFieldValues {
     readonly subtitleFileName: string;
     readonly url?: string;
     readonly image?: ImageModel;
+    readonly mediaFragment?: MediaFragmentModel;
```
killergerbah: This additional field doesn't seem necessary, because video clips all come from `FileModel` anyway. Even with this change, the only base64-serializable media format is image.

common/src/model.ts
```diff
@@ -95,17 +96,22 @@ export interface CopyHistoryItem extends CardModel {
...
-export interface ImageModel {
+export interface MediaFragmentModel {
     readonly base64: string;
-    readonly extension: 'jpeg';
-    readonly error?: ImageErrorCode;
+    readonly extension: 'jpeg' | 'webm';
+    readonly error?: MediaFragmentErrorCode;
 }
 
+// Backward-compatible aliases while callers migrate from image to mediaFragment naming.
```
killergerbah: There is no need for a migration, since video clips only come from the `file` field.

common/src/model.ts
```diff
@@ -142,6 +148,7 @@ export interface AnkiUiState extends CardTextFieldValues {
     readonly url?: string;
     readonly source: string;
     readonly image?: ImageModel;
+    readonly mediaFragment?: MediaFragmentModel;
```
killergerbah: Same here, unnecessary field

common/src/model.ts
```diff
@@ -171,6 +178,7 @@ export interface AnkiUiSavedState {
     text: string;
     definition: string;
     image?: ImageModel;
+    mediaFragment?: MediaFragmentModel;
```
killergerbah: Unnecessary

common/src/media-fragment.ts
```diff
@@ -0,0 +1,446 @@
...
+    static fromCard(
+        card: CardModel,
+        maxWidth: number,
+        maxHeight: number,
+        mediaFragmentFormat: MediaFragmentFormat = 'jpeg',
+        mediaFragmentTrimStart: number = 0,
+        mediaFragmentTrimEnd: number = 0,
+        mediaFragmentMaxClipLength: number = 0
+    ) {
+        if (card.mediaFragment) {
```
killergerbah: This `if` block can be removed, it's not being used

extension/src/ui/components/AnkiUi.tsx
```diff
@@ -307,6 +305,7 @@ export default function AnkiUi({ bridge }: Props) {
             subtitle,
             surroundingSubtitles,
             url,
+            mediaFragment: serializedImage,
```
killergerbah: Not necessary

killergerbah:
Appreciate the patience maintaining this PR
It's been taking extra long for me since it's a big one and there's already a lot of change happening on main
I'll give it another pass and maybe merge hopefully in the next week or so

killergerbah (approved):
Final request: If assisted by AI can you include that either in this PR or a commit message, like this?

```
Assisted-by: AGENT_NAME:MODEL_VERSION
```

Example:
```
Assisted-by: Claude:claude-3-opus
```

I'll include it in the squashed commit before merging

khajiitvaper2017:
added timeout since it would sometimes stall with weirdly encoded videos

killergerbah:
Great work, thanks

PR #929, opened 2026-03-08
Allow collecting tokens with no dictionary entry

### LINKED ISSUE #927: Korean annotations
I've been using annotations for Korean and most of the time it works great, but sometimes there's this issue.

Image

차 here is the word and 가 is the particle. But ASB player grouped them both as if they're one word together. I'm also unable to use Q+any shortcut when I come across these. 

Even if I add both of these separately to my Anki deck (차 and 가), the status will stay uncollected. 

**What browser?**
FireFox
Local video

ShanaryS:
fixes #927

The Yomitan parser simply tries to find the longest dictionary entry possible left to right in order to tokenize. It's possible for some tokens to be non-dictionary entries (until the next dictionary entry-able set of characters appear) which of course won't have any lemmas when we look for them. These entries should just be created with their own token as a lemma to allow the user to collect them.

I modified the strategy logic to fall back to exact if the lemma is missing. We cannot just always add the token as the lemma as we want the failure in all other use cases of `lemmatize()`.

### DIFF
--- a/common/subtitle-coloring/subtitle-coloring.ts
+++ b/common/subtitle-coloring/subtitle-coloring.ts
@@ -327,6 +327,7 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
         if (!ts || !dictionaryTrackEnabled(ts.dt) || !ts.yt) return;
 
         const lemmas = await ts.yt.lemmatize(token);
+        if (!lemmas.length) lemmas.push(token); // Allow collecting ungrouped segments (no dictionary entry)
         await this.dictionaryProvider.saveRecordLocalBulk(profile, [{ token, status, lemmas, states }], applyStates);
         this.tokensForRefresh.add(token);
         for (const lemma of lemmas) this.tokensForRefresh.add(lemma);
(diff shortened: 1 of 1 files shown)


common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -671,7 +672,12 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                         .map((p) => p.text)
                         .join('')
                         .trim();
-                    if (shouldQueryExactForm && !ts.collectedExactForm.has(token)) forExactFormQuery.add(token);
+                    if (
+                        (shouldQueryExactForm || this.tokensForRefresh.has(token)) &&
```
killergerbah, line 676: What's the reason for this particular change? Having a trouble understanding how it fits into the diff

killergerbah: Oh never mind, I think I see. When collecting a token with no lemma, this is needed to cause a refresh to happen.

ShanaryS: We need to update the exact strategy cache in case the user does not have it selected. But we don't need to populate the cache as normal, only for refreshed tokens which is triggered by collecting it.

killergerbah:
I just tested this with the Japanese YT [video](https://www.youtube.com/watch?v=dV4IVaVNwBY) mentioned in the original issue and the ー character in ずーっと is not being rendered on this version of the code

image

killergerbah:
Ah never mind I'm idiot

NovaKing001:
Yeah that was a miscommunication on my part I made those subtitles using whisper.cpp lol

PR #930, opened 2026-03-08
Consistent handling of no token lemmas

ShanaryS:
After thinking about it some more there is some inconsistencies on how no lemmas are being handled. In #929 we need to always build the exact cache in case we are in a new session where the user hasn't collected the token yet, not just for token refreshes.

The Anki database was allowing junk tokens by allowing ones without lemmas to be stored. Even if these tokens were valid it would only match with the exact strategy and not the others which could lead to confusion. So it's best to just ignore these. I don't think it's worth trying to prune the existing entries for users as these shouldn't appear in subtitles and will naturally get pruned over time if the card/note is ever modified.

### DIFF
--- a/common/subtitle-coloring/subtitle-coloring.ts
+++ b/common/subtitle-coloring/subtitle-coloring.ts
@@ -327,7 +327,6 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
         if (!ts || !dictionaryTrackEnabled(ts.dt) || !ts.yt) return;
 
         const lemmas = await ts.yt.lemmatize(token);
-        if (!lemmas.length) lemmas.push(token); // Allow collecting ungrouped segments (no dictionary entry)
         await this.dictionaryProvider.saveRecordLocalBulk(profile, [{ token, status, lemmas, states }], applyStates);
         this.tokensForRefresh.add(token);
         for (const lemma of lemmas) this.tokensForRefresh.add(lemma);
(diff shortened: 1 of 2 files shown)


common/subtitle-coloring/subtitle-coloring.ts
```diff
@@ -672,12 +669,7 @@ export class SubtitleColoring extends SubtitleCollection<RichSubtitleModel> {
                         .map((p) => p.text)
                         .join('')
                         .trim();
-                    if (
-                        (shouldQueryExactForm || this.tokensForRefresh.has(token)) &&
-                        !ts.collectedExactForm.has(token)
-                    ) {
-                        forExactFormQuery.add(token);
-                    }
+                    if (!ts.collectedExactForm.has(token)) forExactFormQuery.add(token); // Always query as we may need to fallback to exact form
```
killergerbah: I guess this is fine because if the exact form is in the database, that's a more accurate representation of their knowledge of the token than the lemma form?

Just wondering if this might alter the existing behavior of the coloring strategies

killergerbah: Ah never mind, it just populates the cache, which can be used as a fallback when the a token has no lemmas

ShanaryS:
I considered just always appending the token if no lemmas but I didn't want it for the other use cases outside of `SubtitleColoring`. I should have just added a flag to control it in the first place rather than trying to change the strategy processing.

> Would it be more consistent to just consider any token with no lemmas as having itself as a lemma?

No we should generally not allow non dictionary entries into the database, at least now there is much reason to. So that when we parse Anki cards or user imports we can ignore these words. But we can allow it when they manually collect it on a subtitle for better UX essentially.

> Or is the intention to only let these types of tokens exist in DB as local tokens picked by the user?

Yes that's the goal. Only local tokens that the user specifically picked out.

killergerbah:
I see, the entire diff including the last commit became a lot simpler with this change

PR #934, opened 2026-03-10
add option to use mecab parser

### LINKED ISSUE #933: Add support for mecab parser via the new parser option in yomitan’s api
New parser option for Yomitan’s [api]( https://github.com/yomidevs/yomitan-api/blob/master/docs/api_paths/tokenize.md)

Add support to toggle between Yomitan’s default parser and [mecab](https://github.com/yomidevs/yomitan-api/blob/master/docs/api_paths/tokenize.md) (A faster/more accurate parser for Japanese) 

Time and parser comparison located [here](https://github.com/yomidevs/yomitan/pull/2254#event-23405668274)

ShanaryS:
fixes #933

Added a new Yomitan option to choose between the `scanning-parser` and `mecab` from a dropdown. When mecab is selected we will verify if the user has it correctly installed by sending a test tokenize string and inspecting the response. This is now possible is mecab is Japanese only.

We should wait for https://github.com/yomidevs/yomitan-mecab-installer/pull/12 to be merged otherwise users may have to install twice.


common/components/DictionaryImport.tsx
```diff
@@ -109,11 +109,12 @@ const DictionaryImport: React.FC<Props> = ({
...
         setImportClipboardPreview(null);
         try {
-            const dt = dictionaryTracks[importClipboardTrack];
-            const yomitan = new Yomitan(dt);
+            const yomitan = new Yomitan(dictionaryTracks[importClipboardTrack]);
+            await yomitan.version();
 
             const tokenSet = new Set<string>();
             for (const tokenParts of await yomitan.splitAndTokenizeBulk(importClipboardText, async (progress) => {
+                yomitan.resetCache();
```
killergerbah, line 117: What's the reason to reset cache here?

common/components/DictionarySettingsTab.tsx
killergerbah: Should add `t` here

common/yomitan/yomitan.ts
```diff
@@ -139,6 +187,33 @@ export class Yomitan {
...
     }
 
+    private filterDictionaries(tokenizeRes: TokenizeResult[]): TokenizeResult[] {
+        if (this.dt.dictionaryYomitanParser !== 'mecab') return tokenizeRes;
+
+        // Mecab can return multiple dictionaries, prefer unidic or the first one if unidic is not present
+        const results: TokenizeResult[] = [];
+        const indexCompleteSet = new Set<number>();
+        for (const res of tokenizeRes) {
+            if (indexCompleteSet.has(res.index)) continue;
```
killergerbah: The usage of `indexCompleteSet` seems to imply that `res.index` can repeat values, doesn't this mean that the `results` array can have holes in it? Is that ok?

common/components/DictionaryImport.tsx
ShanaryS: To prevent memory from growing. The cache for `/tokenize` only matters for subtitles since we frequently reprocess events, but any other case will be useless since it requires an exact sentence match.

common/yomitan/yomitan.ts
ShanaryS: I changed this to do more but the principle is the same for `index`. It corresponds to the index of the input array when there is multiple from batching. This is necessary since Yomitan uses all MeCab dictionaries that a user has for all parsing so each index repeats `* number of MeCab dictionaries`. It can never be sparse, if it is then it would be from this function when we are choosing the best dictionary.

ShanaryS:
I added more checks and instructions in case a user has an outdated installer or a dictionary that doesn't support the lemma from `/tokenize` (will fallback to `/termEntries` which is much slower but accuracy should be the same).

Prefer newer `UniDic` versions for forward comaptability. Currently Yomitan only provides `unidic-mecab-translate` as a default which will probably be what most users use.

common/components/DictionarySettingsTab.tsx
```diff
@@ -223,11 +228,37 @@ const DictionarySettingsTab: React.FC<Props> = ({
...
+            if (!supportsDictionaryYomitanMecab) {
+                setDictionaryYomitanMecabError(undefined);
+                return;
+            }
+            if (selectedDictionary.dictionaryYomitanParser !== 'mecab' || yomitan.getSupportsMecabLemma()) {
+                setDictionaryYomitanMecabError(undefined);
+                return;
+            }
+            if (yomitan.getSupportsMecab()) {
+                const str = t('settings.dictionaryYomitanMecabLemmaNotSupportedError').replace(
```
killergerbah: You can change `dictionaryYomitanMecabError` to a `React.ReactNode` and do something like the below to get the link into the helper text.

```
           setDictionaryYomitanMecabError(
                    <Trans
                        i18nKey="settings.dictionaryYomitanMecabLemmaNotSupportedError"
                        components={[<Link key={0} href="yomitanMecabInstallerUrl" sx={{ cursor: 'pointer' }} />]}
                    />
                );
```

common/components/DictionarySettingsTab.tsx
```diff
@@ -238,7 +269,7 @@ const DictionarySettingsTab: React.FC<Props> = ({
                 setDictionaryYomitanUrlError(String(e));
             }
         }
-    }, [selectedDictionary]);
+    }, [selectedDictionary, t]);
```
killergerbah: Dependency array needs `supportsDictionaryYomitanMecab` now

common/yomitan/yomitan.ts
```diff
@@ -294,17 +428,74 @@ export class Yomitan {
...
         const version: string = (await this._executeAction('yomitanVersion', {}, yomitanUrl)).version;
         if (version === '0.0.0.0') {
-            this.supportsTokenizeFrequency = true;
+            if (this.dt.dictionaryYomitanParser === 'mecab') {
+                await this.verifyMecabSupport(yomitanUrl);
+            } else {
+                this.supportsMecab = false;
+                this.supportsMecabLemma = false;
+            }
+            // this.supportsTokenizeFrequency = true;
```
killergerbah, line 442: It looks like `supportsTokenizeFrequency` is never set to `true` now. Is that intentional?

common/yomitan/yomitan.ts
```diff
@@ -105,46 +169,112 @@ export class Yomitan {
...
+                const month = match?.groups?.month ? parseInt(match.groups.month) : 0;
+                return { year, month };
+            }
+            if (lower === 'ipadic-neologd') return { year: 0, month: 0 };
+            return { year: 1, month: 0 };
+        };
+
+        for (const res of tokenizeRes) {
+            const curr = indexDictMap.get(res.index);
+            const pref = preference(res.dictionary);
```
killergerbah, line 237: Should this `preference` computation be cached?

common/yomitan/yomitan.ts
ShanaryS: Yes we need Yomitan to support it and defaulting to `true` then setting false later is no longer possible if they are using `mecab` tokenize since it won't have `headwords`. It's also safer to just wait for the PR to be merged and support it in the next asbplayer version rather than having it happen automatically. We won't really benefit from it until statistics since it doesn't block anything.

Also, if the user is using `mecab` then it's pointless as those responses can never contain it, even Yomitan would need to call `/termEntries` internally (`scanning-parser` does this which is why it's free).

PR #940, opened 2026-03-14
Dev/vite8

mwojick:
https://vite.dev/blog/announcing-vite8

Still experimental obviously and needs more testing, but basic functions seem to work after fixing 2 broken packages that didn't work with rolldown. Had to switch to videojs-vtt.js which is a fork of vtt.js that's used in video.js. Seems like it has the same base API though. Might also want to consider switching to https://github.com/ajv-validator/ajv for validating, but just updating jsonschema seems to be ok for now.

I'm getting about a 3x initial build speed improvement (30s -> 10s) with this which is amazing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "packageManager": "yarn@3.2.0",
     "resolutions": {
         "lamejs@1.2.0": "patch:lamejs@npm:1.2.0#.yarn/patches/lamejs-npm-1.2.0-b0315d05aa.patch",
-        "vite": "8.0.0"
+        "vite": "8.0.0",
+        "@vitejs/plugin-react": "^6.0.1"
     }
 }
--- a/client/decs.d.ts
+++ b/client/decs.d.ts
@@ -1,1 +1,0 @@
-declare module 'videojs-vtt.js';
--- a/common/decs.d.ts
+++ b/common/decs.d.ts
@@ -1,3 +1,2 @@
-declare module 'videojs-vtt.js';
 declare module 'lamejs';
 declare module 'sanitize-filename';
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -23,6 +23,8 @@
     },
     "packageManager": "yarn@3.2.0",
     "resolutions": {
-        "lamejs@1.2.0": "patch:lamejs@npm:1.2.0#.yarn/patches/lamejs-npm-1.2.0-b0315d05aa.patch"
+        "lamejs@1.2.0": "patch:lamejs@npm:1.2.0#.yarn/patches/lamejs-npm-1.2.0-b0315d05aa.patch",
+        "vite": "8.0.0",
+        "@vitejs/plugin-react": "^6.0.1"
```
mwojick, line 28: Temporary overrides until WXT makes these the default

killergerbah:
I haven't noticed any issues so just going to merge this

Btw how is jsonschema related to this upgrade?

mwojick:
> I haven't noticed any issues so just going to merge this
> 
> Btw how is jsonschema related to this upgrade?

It had to do with this issue: https://github.com/tdegrunt/jsonschema/issues/395
tbh I'm not sure why we just now ran into this since it's been there for a while. Maybe rolldown is more strict about not allowing node deps in the browser.

The vtt issue was different and was fixed here: https://github.com/videojs/vtt.js/pull/14
again, not sure why this was just now surfaced, but maybe rolldown is also more strict about how modules are defined. Like in this case: https://github.com/mozilla/vtt.js/blob/42ca104e8845bfc4485e8579d62cac6feaa00de9/lib/vtt.js#L1344
you can't assume 'this' refers to the window anymore with the way rolldown wraps dependencies, and becomes undefined instead.

PR #941, opened 2026-03-14
  Add subtitle detection for SVT Play and UR Play

jprostko:
### Summary

This adds subtitle detection support for the Swedish streaming sites [SVT Play](https://www.svtplay.se) and [UR Play](https://urplay.se).

SVT Play is the official streaming platform of Sveriges Television (SVT), whereas UR Play is the official streaming platform of UR, Sweden’s public-service educational broadcaster. Together with Sveriges Radio (SR), they are part of Sweden’s public-service broadcasting system.

### Implementation

The SVT Play implementation was pretty straightforward. It makes a simple REST call for each video, and so the implementation just intercepts that fetch, and then re-requests the same URL to extract the subtitle and title data.

The UR Play implementation was more involved (which is probably obvious from the number of comments in its entry point file) because its page data goes stale during client-side navigation, and _serie_ pages and _program_ pages have different URL structures that require separate handling.

The code passes Prettier, ESLint, and `yarn workspace @project/extension compile`. Both the Firefox and Chrome extensions have been tested and confirmed as working as intended on both sites.

Some test URLs with one set of subtitles (as of this pull-request being submitted) that are globally-accessible:

SVT Play: https://www.svtplay.se/video/e9qaywe/leif-och-billy/1-en-ny-tjej-i-byn
UR Play (serie link): https://urplay.se/serie/244870-utraknad
UR Play (program link): https://urplay.se/program/244872-utraknad-overvakade

Some test URLs with multiple sets of subtitles (as of this pull-request being submitted) that are globally-accessible:

SVT Play: https://www.svtplay.se/video/ePB955J/detektiverna/henrik-och-de-misstankta
UR Play (serie link): https://urplay.se/serie/243411-dead-mans-dog
UR Play (program link): https://urplay.se/program/243414-dead-mans-dog-the-dead-man

### Limitations

Affiliated sites like svt.se and svtbarn.se do not gain subtitle support from the SVT Play implementation. They use entirely different backends, and cannot be easily added by minor modification to the SVT Play implementation.  All content on both sites generally ends up on SVT Play anyway in some form, so there may not be a lot of value in implementing them in the future.

Also, content under the [Kanaler section](https://www.svtplay.se/kanaler) on SVT Play cannot have subtitles retrieved, even if a given channel is streaming a show that exists in its entirety on the site elsewhere. Subtitles are delivered in a totally different way as fragmented segments, and therefore a standalone VTT file cannot be intercepted. If one wants the subtitles, it is best to just try to find a particular show episode elsewhere on the site and get the subtitles from there.  Additionally, short-form content like trailers that have /klipp in the URL do not have subtitle support. The videos of this type have hardcoded subtitles in the cases where they have subtitles at all.  Finally, while not common, full shows can have hardcoded subtitles as well, and naturally those cannot be intercepted, since there is nothing to intercept.  These shows will display subtitles no matter what with no option to shut them off.

All videos on UR Play have similar restrictions as SVT Play.  Basically, if there happens to be hardcoded subtitles, or else no subtitles at all, then no subtitles will be intercepted.  This situation is somewhat rare though on UR Play due to its education-focused nature, but it is being mentioned in case it is encountered.

### DIFF
--- a/extension/src/entrypoints/ur-play-page.ts
+++ b/extension/src/entrypoints/ur-play-page.ts
@@ -2,6 +2,7 @@ import { inferTracks } from '@/pages/util';
 
 export default defineUnlistedScript(() => {
     const originalFetch = window.fetch.bind(window);
+    let latestBuildId: string | undefined;
     let latestProductData: any = undefined;
     let latestProductDataPath: string = '';
     let latestProductDataSlug: string = '';
@@ -19,11 +20,12 @@ export default defineUnlistedScript(() => {
                     ? input.href
                     : '';
 
-        const nextDataMatch = url.match(/\/_next\/data\/[^/]+\/program\/([^/]+)\.json/);
+        const nextDataMatch = url.match(/\/_next\/data\/([^/]+)\/program\/([^/]+)\.json/);
         if (nextDataMatch) {
-            // Capture pathname now, since window.location may change before the promise resolves.
+            // Capture buildId and pathname now, since window.location may change before the promise resolves.
+            latestBuildId = nextDataMatch[1];
             const requestPathname = window.location.pathname;
-            const slug = nextDataMatch[1];
+            const slug = nextDataMatch[2];
             const promise = originalFetch(...args);
             promise.then(async (response) => {
                 try {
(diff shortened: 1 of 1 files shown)


extension/src/entrypoints/ur-play-page.ts
```diff
@@ -0,0 +1,166 @@
...
+            latestProductDataSlug = programSlug;
+        }
+
+        return data;
+    };
+
+    const getProductData = async (): Promise<any> => {
+        try {
+            const el = document.getElementById('__NEXT_DATA__');
+            if (!el) return undefined;
```
killergerbah: I'm not sure if it matters, as the script depends on the website being a next js app, but it looks like `__NEXT_DATA__` is not required to be able to get the final data. Does it make sense to instead skip to the fallbacks below which `fetch` the data, instead of early returning?

jprostko: Thank you for the code review.

Good point. The code was bailing out too early when `__NEXT_DATA__` was absent. I will remove that early return and let the code continue to the existing fallback fetch path, using a `buildId` captured from intercepted `/_next/data/` requests when available.

I think that should handle everything.

I'm not sure if I will get to it today (I just started picturing the fix in my head), but I'll submit a fix when I get a chance.

killergerbah:
Amazing work, thank you

PR #942, opened 2026-03-16
Allow mining/line selection from App while streaming

bpwhelan:
Need to test a bit more with the app standalone to make sure there are no regressions, but this allows for selecting lines and mining from the app when it's connected to an active video element in another tab.

I'm not sure if this is desired, but it would be extremely useful for dual monitor setups, cutting out unecessary recording/re-recording that comes with a hotkey-only workflow.

Very useful with the `When loading subtitles, also open subtitle list via the app in a separate tab` option

https://github.com/user-attachments/assets/927b18be-234d-4a2d-ba81-0f9826de7e7e


bpwhelan:
Closing since I didn't really end up having time to test

PR #948, opened 2026-03-23
Add 'Select card to update' action for targeted Anki card updates

### LINKED ISSUE #776: Select specific card to update
Sometimes I want to update a card that wasn't last added.

A UI to search and update the specific card I'm looking for. [Nadeshiko](https://nadeshiko.co/) has a nice interface for this:

Image

I don't think an alternative is possible.

thntx:
Closes #776

## Summary

- Adds a new **"Select card to update"** post-mine action that opens a search dialog when mining, letting the user pick which existing Anki note to update with audio/screenshot instead of always targeting the most recently added card.
- Adds a new `update-selected-card` keyboard shortcut (Chrome extension command) that triggers this action directly.
- The dialog shows the last 50 notes added in the past 30 days, displays the note's **key field** (field with `order === 0`) as the primary label and the configured sentence field as secondary, and supports live search filtering across both.
- A **multi-select** checkbox (off by default, persisted via `localStorage`) lets the user select multiple cards and apply the update to all of them at once via an "Apply (N)" button.
- Fullscreen is gracefully handled: the dialog exits fullscreen, shows, then re-enters fullscreen on close.
- Locale fix: bundled strings are now deep-merged as the base with cached remote translations on top, so new locale keys are never missing even when an older cached version is present.

## Test plan

- [ ] Set "Post-mine action" to "Select card to update" in extension settings
- [ ] Mine a subtitle line — dialog should appear listing recent Anki notes
- [ ] Search filters correctly by key field and sentence field
- [ ] Single-click on a note updates it with audio/screenshot and closes dialog
- [ ] Enable multi-select checkbox → select multiple notes → "Apply (N)" updates all of them
- [ ] Multi-select preference persists across dialog opens
- [ ] Dialog appears and closes correctly when video is in fullscreen
- [ ] New keyboard shortcut appears after "Update last card" in `chrome://extensions/shortcuts`
- [ ] All UI strings display correctly (no raw i18n keys)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


common/anki/anki.ts
```diff
@@ -562,6 +566,46 @@ export class Anki {
                 }
 
                 throw new Error('Could not update last card because the card info could not be fetched');
+            case 'updateSpecific':
```
killergerbah: There's almost no difference between this and the previous block aside from the note that's being updated. Can the code be deduplicated?

extension/src/services/localization-fetcher.ts
```diff
@@ -12,16 +12,33 @@ export interface Localization {
...
+    const bundled = (await bundledStringsForLang(lang)) ?? (await bundledStringsForLang('en'))!;
+    const cached = await cachedStringsForLang(lang);
+
+    if (!cached) {
+        return bundled;
+    }
+
+    // Use bundled as the base so new keys are always available,
+    // then overlay cached strings (which may have newer remote translations).
+    return { lang: cached.lang, strings: deepMerge(bundled.strings, cached.strings) };
```
killergerbah: Can this be reverted? It's not related to this change, and the real solution is to deploy the latest strings before releasing new extension versions.

extension/src/ui/components/CardSelectUi.tsx
```diff
@@ -0,0 +1,386 @@
...
+    }, [notes, search, sentenceField]);
+
+    if (!open) return null;
+
+    return (
+        <StyledEngineProvider injectFirst>
+            <ThemeProvider theme={theme}>
+                <CssBaseline />
+                {/* Backdrop */}
+                <Box
```
killergerbah: Can the UI be split out into its own component? That way the same feature can be implemented for the webapp, eventually.

extension/src/ui/components/CardSelectUi.tsx
```diff
@@ -0,0 +1,386 @@
...
+import { useTranslation } from 'react-i18next';
+import type { PaletteMode } from '@mui/material/styles';
+
+const MULTI_SELECT_KEY = 'cardSelectMultiSelect';
+
+interface Props {
+    bridge: Bridge;
+}
+
+export default function CardSelectUi({ bridge }: Props) {
```
killergerbah: I think that this entire component is unnecessary and should have been written as an extension of the existing `AnkiUi` component for a couple reasons:
- If it were merged with the `AnkiUi` component instead, there would be an opportunity to implement a UI-only way of accessing the card selector.
- A lot of the existing `AnkiUi` code could be used, instead of duplicating it here.

common/anki/anki.ts
thntx: Done — both branches now go through a shared private helper `_updateNoteFields(noteId, params, tags, ankiConnectUrl)`. `updateLast` resolves the noteId from `findNotes('added:1')`; `updateSpecific` takes it as a parameter. Also folded `updateLast || updateSpecific` into a single `isUpdate` boolean for the media-store guard.

extension/src/services/localization-fetcher.ts
thntx: Reverted — `fetchLocalization` is back to the original `cached ?? bundled ?? en` chain and the `deepMerge` helper is gone.

extension/src/ui/components/CardSelectUi.tsx
thntx: Done — the picker UI now lives at `common/components/CardSelectView.tsx` with a props-only API (`anki`, `sentenceField`, `disabled`, `onSelect`, `onCancel`). No bridge knowledge, framework-agnostic enough for the webapp to import it directly when that work happens.

extension/src/ui/components/CardSelectUi.tsx
thntx: Done — `CardSelectUi.tsx`, the standalone `CardSelectController`, the dedicated entrypoint and the `card-select-ui.js` bundle entry are all removed. `AnkiUiState['type']` got a new `'cardSelect'` variant, `AnkiUiController.showCardSelect()` reuses the existing `anki-ui.js` iframe, and `AnkiUi.tsx` renders `CardSelectView` when the state type matches — sharing the bridge, Anki client, settings, theme and Snackbar with the regular dialog flow.

extension/src/ui/components/AnkiUi.tsx
```diff
@@ -56,6 +59,7 @@ const blobToDataUrl = async (blob: Blob): Promise<string> => {
 
 export default function AnkiUi({ bridge }: Props) {
     const [open, setOpen] = useState<boolean>(false);
+    const [uiType, setUiType] = useState<AnkiUiState['type']>('initial');
```
killergerbah: Instead of implementing a UI type I suggest having the keyboard shortcut open both the Anki dialog and the card select dialog with the card select dialog on top, as this would be consistent with the UX as if the user had opened the card select dialog from the Anki dialog (whenever that gets implemented). That would keep the UX consistent.

extension/src/ui/components/AnkiUi.tsx
```diff
@@ -347,6 +352,137 @@ export default function AnkiUi({ bridge }: Props) {
...
+            serializedAudio,
+            file,
+            word,
+            source,
+            url,
+            customFieldValues,
+        ]
+    );
+
+    const handleCardSelectSelect = useCallback(
```
killergerbah, line 352: I believe all these additions that export the card are not necessary. The code to export cards is already in this file, the target Anki card just needs to be injected in the right place.

extension/src/ui/components/AnkiUi.tsx
thntx: Done — dropped the `uiType` state. The picker is now controlled by a `cardSelectOpen?: boolean` flag on `AnkiUiInitialState`. `AnkiUi.tsx` renders `AnkiDialog` unconditionally and overlays `CardSelectView` (with `zIndex: theme.zIndex.modal + 1`) on top when the flag is true. Both entry points — the keyboard shortcut now, and the eventual "open from AnkiDialog" button — would land on the same UI state.

extension/src/ui/components/AnkiUi.tsx
thntx: Done — `AnkiDialog` now exposes a `buildExportParams(mode, noteId?)` helper on its `stateRef` (its own `handleProceed` also routes through this helper, so there's still one source of truth inside the dialog). The card-select handler is now a small loop: `for (const noteId of noteIds) await anki.export(dialogStateRef.current.buildExportParams('updateSpecific', noteId))`, then the existing `card-updated-dialog` + `resume` flow. Side benefit: any edits the user makes in the dialog before picking a note carry over.

killergerbah:
Thanks @thntx for the highly desired feature and appreciate the patience
I will likely adjust the UX post merge

PR #952, opened 2026-04-04
Statistics UI

killergerbah:
First pass on UI that just implements the basic flows:

- Statistics have been moved out of the settings form into its own component
- Statistics can be accessed in popup or in side panel
- The popup will only show statistics section if a video has subtitles loaded

Still a lot of TODOs:
- [x] Support a disabled UI state if the anki word database isn't created/enabled
- [x] In-app statistics without the extension
- [x] Update keybinds to support new statistics UI locations - side panel by default but may need to open new window for browsers that don't support the side panel
- [x] Mining/seeking from sentences popups work
- [x] Update the "good sentences" UI. Inside the side panel, it's currently too cramped.
- [x] Statistics UI should respond in real time to newly-loaded subtitles
- [x] Statistics overlay 
- [x] Polish/cleanup
- [x] Won't fix ~Bug - app statistics can show statistics from extension media and vice versa~ won't fix~


ShanaryS (commented):
Everything looks good. The only things I would like are:
* Keeping a way to view statistics on a separate page. Maybe we can support a popout that we can freely resize and move to a second monitor. This allows keep the side panel subtitles while still having statistics on the second monitor. This is mainly useful for mining workflows (which will get more in depth in the future likely from the stats menu).
  * This could be added as a setting but I think it's fine to default to side panel and only open popup through a button.
* An overlay on the video or in the header for App views that shows a few stats (clicking on it also opens stats):
  * Comprehension % (using the same colors but no band, maybe we color the % in stats view as well)
  * 1 Uncollected count where clicking on it brings us to that filtered sentence view
  * I think these are the only stats that need high visibility

killergerbah:
@ShanaryS I think this is basically ready
Some minor concerns that can be addressed in another diff:
- when statistics auto generate is not enabled, the stats overlay doesnt make as much sense, since generating the statistics involves opening the statistics window.
- thinking to add a setting to disable the overlay.

ShanaryS (approved):
Looks good!

> when statistics auto generate is not enabled, the stats overlay doesnt make as much sense, since generating the statistics involves opening the statistics window.

> thinking to add a setting to disable the overlay.

I think both of these could be addressed by tying the overlay defaulting to on with the auto generate setting and adding a button next to the popout to toggle the overlay. I don't think we need a setting as if they want auto generate, they probably also want the overlay (we can decouple it later if requested).

I also would like this and the popout button should be present when view the stats from the extension popup.

When you're ready to merge I'd prefer it squashed since the commit history is already in this PR.

PR #953, opened 2026-04-08
Show mining history count and fix stale jump target after seek

L-M-Sherlock:
## Use case

I record how many lines I mine for each episode for stats and analysis. It's also helpful for estimating the difficulty of anime.

  ## Summary
  - show the current mining history count in the mining history drawer header
  - show the count in the side panel mining history entry points
  - add a badge to the top history action and keep `0` visible
  - fix a stale jump target bug where adjusting subtitle offset after jumping from mining history could seek back to the previously selected subtitle

 ## Screenshots

image

image


common/app/components/CopyHistory.tsx
```diff
@@ -18,9 +25,32 @@ interface CopyHistoryProps {
 }
 
 export default function CopyHistory(props: CopyHistoryProps) {
+    const { t } = useTranslation();
+    const appBarHeight = useAppBarHeight();
+
     return (
-        <Drawer showBackButton open={props.open} drawerWidth={props.drawerWidth} onClose={props.onClose}>
-            <CopyHistoryList {...props} />
+        <Drawer showBackButton={false} open={props.open} drawerWidth={props.drawerWidth} onClose={props.onClose}>
```
killergerbah: I think you'd get more reusability if the label becomes a property of the `Drawer` instead of drawing it here

extension/src/ui/components/SidePanel.tsx
```diff
@@ -585,18 +588,26 @@ export default function SidePanel({ dictionaryProvider, settingsProvider, settin
...
-                        forceShowDownloadOptions={true}
-                        onClose={handleCloseCopyHistory}
-                        onDelete={deleteCopyHistoryItem}
-                        onDeleteAll={deleteAllCopyHistoryItems}
-                        onAnki={handleAnki}
-                        onClipAudio={handleClipAudio}
-                        onDownloadImage={handleDownloadImage}
-                        onSelect={handleJumpToSubtitle}
-                    />
+                    <div style={{ display: 'flex', flexDirection: 'column', height: '100%', minHeight: 0 }}>
```
killergerbah: If you make the label a property of `Drawer` you can use the `CopyHistory` component (which contains `Drawer`) here with `showBackButton` set to false

common/app/components/CopyHistory.tsx
L-M-Sherlock: get it

killergerbah:
Awesome, thanks as always

PR #956, opened 2026-04-12
Fix Disney+ subtitle seek and sync via the player API

Ayase-the-Dark:
## Summary

Drive Disney+ subtitle seek and sync through the site's internal player API instead of MSE timestamp math.

Direct writes to `video.currentTime` are ignored by the Disney+ player (it snaps back to internal state), and `video.currentTime` does not correspond to content time, so the previous `SourceBuffer.timestampOffset` approach could not seek or stay in sync. This reworks the integration to mirror the Netflix path:

- **Player API access (`disney-plus-page.ts`, MAIN world):** locate the Disney+ player on the React fiber tree above `<video>` (object exposing `seek`/`scrub`/`play`/`pause`) and handle `asbplayer-disney-plus-seek/play/pause`. Push the player's true content time (`timeline.info.playheadPositionMs`) to the content script via `asbplayer-disney-plus-time`.
- **`binding.ts`:** `seek`/`play`/`pause` dispatch to the player API when `_isDisneyPlus()` (page-config driven). `contentTimeMs` returns the player's content time (interpolated while playing) for Disney+. Seek completion is awaited before mining (`_waitForDisneyPlusSeek`) since Disney+ does not fire `seeked`.

### Changed files

| File | Change |
|------|--------|
| `extension/src/entrypoints/disney-plus-page.ts` | Reverse-engineered player-API accessor + seek/play/pause handlers + content-time push |
| `extension/src/services/binding.ts` | Disney+ branches for seek/play/pause, `contentTimeMs` from player content time, seek-completion wait |
| `extension/src/services/disney-plus-mse-offset.ts` / `.test.ts` | `isDisneyPlusPageConfigKey` helper (page-config host detection) |
| `extension/src/services/video-change-handler.ts` / `.test.ts` | Debounced src-change gating for loadedmetadata |

## Test plan

- [x] Unit tests pass
- [x] Disney+: subtitle-row click seeks the video
- [x] Disney+: overlay + side-panel subtitle list stay in sync with playback
- [x] Disney+: Anki mining captures audio at the correct position
- [ ] Verify no regression on YouTube, Netflix, or local file playback


extension/src/entrypoints/disney-plus-page.ts
```diff
@@ -2,6 +2,169 @@ import { inferTracks } from '@/pages/util';
...
+        const offsetSeconds = state.videoOffsetSeconds ?? state.audioOffsetSeconds;
+
+        if (offsetSeconds === undefined || !validOffsetSeconds(offsetSeconds)) {
+            return undefined;
+        }
+
+        return Math.round(offsetSeconds * 1000);
+    };
+
+    const currentOffsetDetail = () => {
```
killergerbah: Is there a specific reason why this logic to determine the MSE offset is in this page script rather than in the content script? I can see benefits to putting this logic in a content script instead:
- Would avoid needing to use `document.dispatchEvent` for communication
- Technically, the offset being published by this script may not be for the same video that the receiving `Binding` instance is bound to. At least, it's not enforced by the code.

If there isn't a specific reason, then I would suggest moving this logic into a file included by `binding` only for Disney Plus hosts, which can be determined by something like `currentPageDelegate()?.config?.key === 'disneyPlus'`

extension/src/services/binding.ts
```diff
@@ -96,6 +97,9 @@ document.addEventListener('asbplayer-netflix-enabled', (e) => {
 document.dispatchEvent(new CustomEvent('asbplayer-query-netflix'));
 
 const youtube = /(m|www)\.youtube\.com/.test(window.location.host);
+const disneyPlus = window.location.host === 'www.disneyplus.com';
```
killergerbah: While there is precedent for this method of determining the current host, I would suggest using the page delegate -> page config, so that all page-specific behavior is driven mostly by the config file `pages.json`.

extension/src/services/binding.ts
```diff
@@ -1593,6 +1611,63 @@ export default class Binding {
...
+        if (!disneyPlus || !this.hasPageScript) {
+            return this._pageScriptMseBaseOffsetMs;
+        }
+
+        document.dispatchEvent(new CustomEvent(requestDisneyPlusMseOffsetEventName));
+        return this._pageScriptMseBaseOffsetMs;
+    }
+
+    private _detectMseBaseOffset(): number {
+        const pageScriptOffsetMs = this._requestDisneyPlusMseBaseOffset();
```
killergerbah: Why are there two separate offset values? How are they different?

extension/src/services/binding.ts
```diff
@@ -1593,6 +1611,63 @@ export default class Binding {
...
+            this._pageScriptMseBaseOffsetMs = offsetMs;
+            this._mseBaseOffsetMs = offsetMs;
+            this.subtitleController.mseBaseOffsetMs = offsetMs;
+
+            const subtitles = this.subtitleController.subtitles;
+
+            if (subtitles.length > 0) {
+                const currentOffset = subtitles[0].start - subtitles[0].originalStart;
+                const userOffset = currentOffset - previousMseBaseOffsetMs;
+                this.subtitleController.offset(userOffset + offsetMs, true);
```
killergerbah: I don't think `this.subtitleController.offset` should be used to apply the MSE offset. `this.subtitleController.offset` is used to apply the user-specified offset and by combining the two values into one, you're needing to compensate in other places.
Instead, I would suggest modifying how video's current timestamp is exposed to the app. The rest of the code should behave correctly according to that.

Ayase-the-Dark:
Thanks for the detailed review! After more testing I found the MSE `timestampOffset` approach can't work on Disney+ at all: direct writes to `video.currentTime` are ignored by the player (it snaps back to its internal state), so subtitle-click seek never actually moved playback regardless of the offset math. I've reworked the PR to drive Disney+'s internal player API directly, mirroring the existing Netflix integration. How this maps to your points:

- **page script vs content script:** the new logic must run in the page's MAIN world — the Disney+ player object (exposing `seek()`/`play()`/`pause()` and `timeline.info.playheadPositionMs`) lives on the React fiber tree in page context and isn't reachable from a content script. That's the same reason `netflix-page.ts` runs in-page and talks over `document.dispatchEvent`, so I followed that precedent (`asbplayer-disney-plus-seek/play/pause/time`).
- **host detection:** now page-config driven — `_isDisneyPlus()` just checks the page config key (`isDisneyPlusPageConfigKey(this.pageConfigKey)`).
- **two offsets / `subtitleController.offset`:** both gone. No Disney-specific MSE offset remains, and nothing is folded into `subtitleController.offset` (it stays purely the user offset). As you suggested, I instead changed how the current timestamp is exposed: `contentTimeMs` now returns the player's true content time (`playheadPositionMs`, interpolated while playing) for Disney+, and the rest of the app follows from that. The remaining generic `_mseBaseOffsetMs` is decoupled from Disney+ and only applies to other blob/MSE sources.

I've updated the title/description to match the new approach.

extension/src/services/binding.ts
```diff
@@ -211,13 +223,29 @@ export default class Binding {
...
+        if (this._isDisneyPlus() && disneyPlusContentTimeMs !== undefined) {
+            // Disney+: video.currentTime is unreliable; use the player's true content
+            // time pushed from the page script, interpolated while playing.
+            const elapsed = this.video.paused
+                ? 0
+                : (performance.now() - (disneyPlusContentTimeAt ?? 0)) * this.video.playbackRate;
+            return disneyPlusContentTimeMs + elapsed;
+        }
+
+        return this.video.currentTime * 1000 - this._mseBaseOffsetMs;
```
killergerbah: After the most recent change, is `this._mseBaseOffsetMs` still useful for Disney Plus? If it's not useful for Disney Plus, can it be removed?

extension/src/services/binding.ts
```diff
@@ -1502,12 +1589,54 @@ export default class Binding {
...
+
+            this.video.addEventListener('play', listener);
+            this.video.addEventListener('playing', listener);
+        });
+    }
+
+    // Disney+ player-API seeks are async and do not fire 'seeked', so wait until the
+    // player's content time reaches the target before recording. Otherwise mined audio
+    // is captured from the pre-seek position.
+    private async _waitForDisneyPlusSeek(targetSec: number, timeoutMs = 3000) {
```
killergerbah: Is this needed because Disney Plus takes longer to seek than other video players?

The 'seeked' event is not actually used by asbplayer for seeking

extension/src/services/disney-plus-mse-offset.ts
```diff
@@ -0,0 +1,5 @@
+const disneyPlusPageConfigKeys = new Set(['disneyPlus', 'appsDisneyPlus']);
```
killergerbah: Does apps.disneyplus.com actually work the same way as disneyplus.com? If so the same page script would need to be configured for apps.disneyplus.com. If not then we can remove this function IMO and just do a comparison of the page key with `disneyPlus`.

extension/src/services/binding.ts
Ayase-the-Dark: Good catch — it's no longer useful for Disney+. Now that seek/sync goes through the player API, `_mseBaseOffsetMs` was only left over from the abandoned MSE-offset approach and wasn't actually serving any other source either, so I removed it entirely (along with `_detectMseBaseOffset`). `contentTimeMs` and `seek()` now fall back to plain `video.currentTime` for non-Disney+ sources, matching the pre-PR behavior.

extension/src/services/binding.ts
Ayase-the-Dark: You're right that asbplayer doesn't rely on the 'seeked' event for seeking — my comment was misleading, I've reworded it. The wait is needed because the Disney+ seek goes through the player API and is asynchronous: `timeline.info.playheadPositionMs` (which `contentTimeMs` reads) only catches up to the target a short while after the seek is requested. Without waiting, mining starts recording from the pre-seek position. For ordinary `<video>` elements the seek takes effect synchronously enough that no wait is needed, so this is Disney+-specific.

extension/src/services/disney-plus-mse-offset.ts
Ayase-the-Dark: No — apps.disneyplus.com does not work the same way. It's configured with a separate page script (`apps-disney-plus-page.ts`) that only does subtitle-track inference via MPD interception; it does not expose/drive the player API (seek/play/pause/time). Treating `appsDisneyPlus` as Disney+ would actually break seek/play/pause there since nothing handles those events. So I followed your suggestion: removed the helper (and the misnamed `disney-plus-mse-offset` file) and now just compare `pageConfigKey === 'disneyPlus'`.

killergerbah:
@Ayase-the-Dark I've tested your changes and tweaked them a bit.

I'd like to merge this PR, but can you first comment with what AI tools were used to author this change as per the [contribution guidelines](https://github.com/asbplayer/asbplayer/blob/main/CONTRIBUTING.md#ai-assisted-contributions)?

I will modify the commit message as necessary before merging.

Thank you for your work! This is a valuable feature which I would have never figured out on my own.

Ayase-the-Dark:
This change was authored with the assistance of Claude Code, using the Claude Opus 4 family — the initial investigation and MSE-based attempts with claude-opus-4-6, the pivot to the player-API approach with claude-opus-4-7, and the final implementation (the commits in this PR) with claude-opus-4-8. So the tag for the PR commits would be:

```
Assisted-by: Claude Code:claude-opus-4-8
```

Chrome DevTools MCP was used to reverse-engineer the Disney+ player API. Thanks for testing and tweaking the changes!

killergerbah:
Awesome, thanks again @Ayase-the-Dark

PR #957, opened 2026-04-14
feat: add online subtitle source support

xwxb:
### Summary

This PR adds support for importing subtitles from online sources in the Video Data Sync flow.

### What’s included

- Added a new Online subtitle sources dialog in Video Data Sync.
- Added support for:
    - jimaku.cc (API key based search and file listing)
    - subtitles.ajatt.top (directory/table parsing and file listing)
- Added entry points to open online sources from:
    - Track selector menu (Search online subtitles)
    - Dialog actions (Online sources)
- Imported online subtitle files are converted into local tracks and selected automatically for the target slot.
- Jimaku API key persistence is handled in extension storage (with localStorage fallback).
- Improved error handling for online requests:
    - safer JSON parsing for Jimaku responses
    - explicit HTTP status checks for AJATT requests
- Added/updated tests for service behaviors and failure paths.

### Screenshots
image
image

### Validation

- yarn workspace @project/extension run test extension/src/services/subtitle-sources.test.ts
- yarn workspace @project/extension run compile
- yarn workspace @project/extension run zip

All pass locally.

---

 I may still have missed details, so feedback is very welcome and I will actively revise this PR.

  I’m also not fully sure whether shipping subtitles.ajatt.top as a source has any policy/compliance concerns in this project. If
  this source should be removed, I can remove it immediately and keep the rest of the implementation.

### DIFF
--- a/common/global-state/index.ts
+++ b/common/global-state/index.ts
@@ -9,16 +9,24 @@ export enum AnnotationTutorialState {
     hasSeen = 2,
 }
 
+export interface OnlineSubtitleSourceConfig {
+    jimakuApiKey: string;
+}
+
 export const initialGlobalState: GlobalState = {
     ftueHasSeenAnkiDialogQuickSelectV2: false,
     ftueHasSeenSubtitleTrackSelector: false,
     ftueAnnotation: AnnotationTutorialState.hasNotSeen,
+    onlineSubtitleSourceConfig: {
+        jimakuApiKey: '',
+    },
 };
 
 export interface GlobalState {
     ftueHasSeenAnkiDialogQuickSelectV2: boolean;
     ftueHasSeenSubtitleTrackSelector: boolean;
     ftueAnnotation: AnnotationTutorialState;
+    onlineSubtitleSourceConfig: OnlineSubtitleSourceConfig;
 }
 
 export interface GlobalStateProvider {
--- a/extension/src/services/subtitle-sources.ts
+++ b/extension/src/services/subtitle-sources.ts
@@ -112,7 +112,6 @@ const parseOptionalFloat = (value: string | null): number | undefined => {
     return Number.isFinite(parsed) ? parsed : undefined;
 };
 
-
 export interface JimakuClientOptions {
     apiKey: string;
     baseUrl?: string;
@@ -225,4 +224,3 @@ export class JimakuClient {
         }
     }
 }
-
--- a/extension/src/services/subtitle-sources.ts
+++ b/extension/src/services/subtitle-sources.ts
@@ -1,0 +1,228 @@
+export interface JimakuEntry {
+    id: number;
+    anilist_id?: number;
+    name: string;
+    japanese_name?: string;
+    english_name?: string;
+    created_at?: string;
+    last_updated_at?: string;
+    flags?: number;
+}
+
+export interface JimakuFile {
+    id: number;
+    name: string;
+    url: string;
+    created_at?: string;
+    size?: number;
+}
+
+export interface JimakuRateLimit {
+    limit?: number;
+    remaining?: number;
+    resetAfterSeconds?: number;
+}
+
+export interface JimakuResponse<T> {
+    data: T;
+    rateLimit: JimakuRateLimit;
+}
+
+interface JimakuErrorPayload {
+    error?: string;
+    message?: string;
+}
+
+const parseJsonSafely = (text: string): unknown | undefined => {
+    if (text.length === 0) {
+        return undefined;
+    }
+
+    try {
+        return JSON.parse(text);
+    } catch {
+        return undefined;
+    }
+};
+
+const defaultJimakuBaseUrl = 'https://jimaku.cc/api';
+
+type TrustedHtmlPolicyLike = {
+    createHTML: (value: string) => string | TrustedHTML;
+};
+
+let trustedHtmlPolicy: TrustedHtmlPolicyLike | undefined;
+
+const createTrustedHtml = (html: string): string | TrustedHTML => {
+    const trustedTypesApi = (
+        globalThis as typeof globalThis & {
+            trustedTypes?: {
+                createPolicy: (
+                    name: string,
+                    policy: { createHTML: (value: string) => string }
+                ) => TrustedHtmlPolicyLike;
+                getPolicy?: (name: string) => TrustedHtmlPolicyLike | null;
+            };
+        }
+    ).trustedTypes;
+
+    if (!trustedTypesApi) {
+        return html;
+    }
+
+    if (!trustedHtmlPolicy) {
+        try {
+            trustedHtmlPolicy = trustedTypesApi.createPolicy('asbplayer-subtitle-sources', {
+                createHTML: (value) => value,
+            });
+        } catch (error) {
+            trustedHtmlPolicy = trustedTypesApi.getPolicy?.('asbplayer-subtitle-sources') ?? undefined;
+        }
+    }
+
+    return trustedHtmlPolicy ? trustedHtmlPolicy.createHTML(html) : html;
+};
+
+const parseHtmlDocument = (html: string) => {
+    const trustedHtml = createTrustedHtml(html);
+    return new DOMParser().parseFromString(trustedHtml as string, 'text/html');
+};
+
+const parseRateLimit = (headers: Headers): JimakuRateLimit => ({
+    limit: parseOptionalInt(headers.get('x-ratelimit-limit')),
+    remaining: parseOptionalInt(headers.get('x-ratelimit-remaining')),
+    resetAfterSeconds: parseOptionalFloat(headers.get('x-ratelimit-reset-after')),
+});
+
+const parseOptionalInt = (value: string | null): number | undefined => {
+    if (value === null) {
+        return undefined;
+    }
+
+    const parsed = Number.parseInt(value, 10);
+    return Number.isFinite(parsed) ? parsed : undefined;
+};
+
+const parseOptionalFloat = (value: string | null): number | undefined => {
+    if (value === null) {
+        return undefined;
+    }
+
+    const parsed = Number.parseFloat(value);
+    return Number.isFinite(parsed) ? parsed : undefined;
+};
+
+
+export interface JimakuClientOptions {
+    apiKey: string;
+    baseUrl?: string;
+    minRequestIntervalMs?: number;
+}
+
+export class JimakuClient {
+    private readonly _apiKey: string;
+    private readonly _baseUrl: string;
+    private readonly _minRequestIntervalMs: number;
+    private _lastRequestTimestampMs?: number;
+    private _lastRateLimit?: JimakuRateLimit;
+
+    constructor({ apiKey, baseUrl = defaultJimakuBaseUrl, minRequestIntervalMs = 1000 }: JimakuClientOptions) {
+        const trimmedApiKey = apiKey.trim();
+
+        if (trimmedApiKey.length === 0) {
+            throw new Error('Jimaku API key cannot be empty or whitespace-only');
+        }
+
+        this._apiKey = trimmedApiKey;
+        this._baseUrl = baseUrl;
+        this._minRequestIntervalMs = minRequestIntervalMs;
+    }
+
+    async searchEntries(query: string): Promise<JimakuResponse<JimakuEntry[]>> {
+        const searchParams = new URLSearchParams();
+        searchParams.set('query', query);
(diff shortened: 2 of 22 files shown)


xwxb:
Took a pass through the Copilot auto-review and applied the relevant fixes. For the filename and object-URL lifecycle points, the final decisions were:

- **Filename normalization for online imports**
  - We now normalize imported online subtitle names so `label` and `extension` stay consistent.
  - If the source name already has an extension, we keep it.
  - If the source name is missing an extension, we first infer it from the source URL filename; only if that is unavailable do we fall back to `srt`.
  - This avoids downstream edge cases where extension-stripping logic can produce invalid names (for example, empty basename leading to `.srt`), while still handling imperfect remote metadata.

- **Object URL lifecycle management**
  - Imported local/online subtitle tracks create blob URLs via `URL.createObjectURL`.
  - We now explicitly track URLs created by this component and revoke them when they are no longer referenced by subtitle tracks (replace/remove path).
  - We also revoke any remaining tracked URLs on component unmount as a safety net (cancel/close/navigation paths where sync never consumes them).
  - This keeps behavior unchanged for users, but prevents long-lived blob URL leaks.

In short: we kept the UX permissive for real-world source metadata, but made filename handling deterministic and tightened blob URL cleanup to avoid lifecycle leaks.

extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
```diff
@@ -0,0 +1,476 @@
...
+
+type ExtensionStorageAreaLike = {
+    get: (
+        keys?: string | string[] | Record<string, unknown> | null,
+        callback?: (items: Record<string, unknown>) => void
+    ) => Promise<Record<string, unknown>> | void;
+    set: (items: Record<string, unknown>, callback?: () => void) => Promise<void> | void;
+};
+
+const getExtensionStorage = () => {
```
killergerbah: This will always return undefined because `VideoDataSyncDialog` is inside an `about:blank` iframe manually filled with HTML inside the page.
I would suggest communicating with `video-data-sync-controller.ts` which has access to extension APIs like `browser.storage`.

I would also suggest using a `GlobalStateProvider` instead of `browser.storage` directly to store the configuration since this data should be available to the webapp as well, so that the feature can eventually be implemented for the webapp.

extension/src/services/subtitle-sources.ts
```diff
@@ -0,0 +1,346 @@
...
+        }
+
+        return {
+            data: parsedBody as T,
+            rateLimit,
+        };
+    }
+
+    private async _waitIfNeeded() {
+        if (this._lastRequestTimestampMs === undefined || this._minRequestIntervalMs <= 0) {
```
killergerbah, line 215: Should this method be using the rate limit data you're parsing out of the response?

extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
xwxb: Good point, sorry for not digging deeper into the extension architecture on the first pass, learned a lot here. I've refactored it to use the existing global-state system as suggested, so the config now persists properly across pages and the webapp can reuse it too.

Let me know if any other issues!

xwxb:
Synced with main via merge since the current commit history is a bit messy. I'll perform a final squash before merging to production.

killergerbah:
There are some minor UX quirks:
- The "entries" and then "files" tables should probably be two columns ideally, instead of two rows
- When re-opening the online sources dialog, it is briefly in a populated state then returns to the empty state

killergerbah:
Also, I'm not currently sure what the legal implications of this integration are, will need to think on it

extension/src/services/subtitle-sources.ts
xwxb: true, that was a miss. I've updated the method to use the parsed rate limit data.

xwxb:
I’ve fixed the UX quirks (two-column layout and dialog state) and regarding legal concerns, I’ve removed the AJATT source and reached out to the jimaku.cc maintainer for official permission. let me know if there's anything else

extension/src/services/subtitle-sources.ts
```diff
@@ -0,0 +1,224 @@
...
+
+const parseHtmlDocument = (html: string) => {
+    const trustedHtml = createTrustedHtml(html);
+    return new DOMParser().parseFromString(trustedHtml as string, 'text/html');
+};
+
+const parseRateLimit = (headers: Headers): JimakuRateLimit => ({
+    limit: parseOptionalInt(headers.get('x-ratelimit-limit')),
+    remaining: parseOptionalInt(headers.get('x-ratelimit-remaining')),
+    resetAfterSeconds: parseOptionalInt(headers.get('x-ratelimit-reset-after')),
```
Rapptz: This is actually a float, so it can be e.g. `1.5` for 1.5 seconds. If you parse it as an integer you get rid of the millisecond precision. That's why it's a separate header, cause otherwise `Reset-After` would have worked just fine but the new header lets me use ms.

common/global-state/index.ts
```diff
@@ -9,16 +9,24 @@ export enum AnnotationTutorialState {
...
+export interface OnlineSubtitleSourceConfig {
+    jimakuApiKey: string;
+}
+
 export const initialGlobalState: GlobalState = {
     ftueHasSeenAnkiDialogQuickSelectV2: false,
     ftueHasSeenSubtitleTrackSelector: false,
     ftueAnnotation: AnnotationTutorialState.hasNotSeen,
+    onlineSubtitleSourceConfig: {
+        jimakuApiKey: '',
```
Rapptz, line 21: I'm assuming this means that users provide their own API key, which is preferred. I'd rather not have 1 API key shared amongst every user of the service.

extension/src/services/subtitle-sources.ts
```diff
@@ -0,0 +1,224 @@
...
+    ): Promise<JimakuResponse<JimakuFile[]>> {
+        const searchParams = new URLSearchParams();
+
+        if (options?.episode !== undefined) {
+            searchParams.set('episode', `${options.episode}`);
+        }
+
+        if (options?.language) {
+            searchParams.set('language', options.language);
+        }
```
Rapptz: This query parameter does not exist in the API (nor will it ever).

Rapptz (commented):
Hello, I'm the maintainer/creator of jimaku.cc and I was asked to review the code in this PR so I gave it a go.

The implementation seems mostly fine to me, I just had a few nits/comments. From a permission perspective, it's more than fine with me if you use the API for your project since ultimately we share the same goal of helping Japanese learners. My only real requests are following the rate limit headers properly (mostly done correctly here) and ensuring that people provide their own key. I'd also prefer it if you used a user-agent to mark where the request is coming from but I understand `fetch` (in Chromium) does not allow you to do this in a browser context.

There are also some logical issues that this doesn't tackle. `jimaku` hosts a wide array of subtitle formats including `.zip`, `.idx`, `.sup`, `.sub`, `.ass` and `.srt`. As far as I'm aware and judging from other integrations, you want to have a filter to only find and apply subtitles that are supported with asbplayer, such as `.srt` and `.ass`. I might have missed where this was handled (if it was) but I couldn't find it from looking at this changeset.

Other than that, everything looks fine with me. Thanks for the work put into this, I've seen people ask for an integration in my Discord server and there have been a few 3rd party ones already.

extension/src/services/subtitle-sources.ts
xwxb: my miss, I've corrected it

common/global-state/index.ts
xwxb: yeah, current logic is that user cannot get subs until get their own apikey

extension/src/services/subtitle-sources.ts
xwxb: my bad, removed in a91c3435

xwxb:
> Hello, I'm the maintainer/creator of jimaku.cc and I was asked to review the code in this PR so I gave it a go.
> 
> The implementation seems mostly fine to me, I just had a few nits/comments. From a permission perspective, it's more than fine with me if you use the API for your project since ultimately we share the same goal of helping Japanese learners. My only real requests are following the rate limit headers properly (mostly done correctly here) and ensuring that people provide their own key. I'd also prefer it if you used a user-agent to mark where the request is coming from but I understand `fetch` (in Chromium) does not allow you to do this in a browser context.
> 
> There are also some logical issues that this doesn't tackle. `jimaku` hosts a wide array of subtitle formats including `.zip`, `.idx`, `.sup`, `.sub`, `.ass` and `.srt`. As far as I'm aware and judging from other integrations, you want to have a filter to only find and apply subtitles that are supported with asbplayer, such as `.srt` and `.ass`. I might have missed where this was handled (if it was) but I couldn't find it from looking at this changeset.
> 
> Other than that, everything looks fine with me. Thanks for the work put into this, I've seen people ask for an integration in my Discord server and there have been a few 3rd party ones already.

thanks for going above and beyond, I only expected a permission check, but your code review was incredibly helpful, helping me solve several critical issues I missed by not reading the docs carefully enough.

I've fixed all the logical issues you mentioned. I’ve now clarified and strengthened the file format restrictions as suggested. As for the API keys, the implementation is already strictly user-provided by design; there are definitely no shared keys or bypasses.

As for the User-Agent, since fetch has its limits in the browser, perhaps an X-Client header could be an alternative for identifying the source? If it's necessary,  I'll discuss this further with @killergerbah. Thanks again for your incredible support!

killergerbah:
I think this is fine to merge, I may tweak the UI/UX post-merge to satisfy my OCD. If a custom header besides user agent can be useful to @Rapptz then we should add that, possibly in a different PR.

Regarding legal implications, I was more concerned about copyright issues, but it seems that subtitle API integration isn't any worse than what asbplayer already does:

```
I guess my main question is, is it any more copyright-infringing to integrate with a copyright-infringing subtitle API vs the base functionality of providing a way to extract subtitles from streaming video sources

6:41 PM
Claudeが返答しました:
Honestly, probably not — and in some ways the API integration is cleaner.

Extracting directly from streaming sources is arguably more legally fraught because:

It likely involves circumventing DRM or access controls, which triggers the DMCA's anti-circumvention provisions (Section 1201) — a separate and stricter liability than copyright infringement itself
It interferes with the technical delivery system the studio controls
It's harder to argue transformative purpose when you're actively bypassing their infrastructure
Integrating with a third-party subtitle API just means consuming data from another source. You're not circumventing anything yourself. The infringing act (if any) already happened upstream. Your legal exposure is secondary/contributory infringement, which is a softer standard and harder to prove.

So the incremental risk of the API integration is low — you're already accepting the base copyright risk by building the tool at all. Adding an API source doesn't meaningfully change your exposure, and might actually reduce it compared to direct extraction.

The one caveat is if the API integration makes your tool more capable or scalable in ways that increase harm to rightsholders — but that's a practical concern about the product's footprint, not a meaningful legal distinction between the two approaches.

In short: the legal risk lives in the product concept, not the specific technical integration method you choose.
```

extension/src/ui/components/VideoDataSyncUi.tsx
```diff
@@ -223,6 +328,57 @@ export default function VideoDataSyncUi({ bridge }: Props) {
...
+                url: objectUrl,
+                extension,
+                localFile: true,
+            };
+
+            setSubtitles((s) => [...s, track]);
+            setSelectedSubtitleTrackIds((s) => {
+                const next = [...s];
+
+                if (onlineDialogTrackNumber !== undefined) {
```
killergerbah, line 368: This UX is better than what we have for regular files but now we have two slightly different user experiences depending on whether a regular file is imported vs an online file is imported

killergerbah (approved):

killergerbah:
@xwxb Final request: if AI tools were used to write this change, can you indicate that in the commit message like below?

```
Assisted-by: AGENT_NAME:MODEL_VERSION
```

For example:

```
Assisted-by: Claude:claude-3-opus
```

extension/src/ui/components/VideoDataSyncUi.tsx
xwxb: to be honest, I'm not entirely sure where the regular files part feels inconsistent, so I’m more than happy to let you tweak it post-merge to match your vision. If you’d prefer me to take another crack at it now, just let me know

xwxb:
@killergerbah thank for approvel, i've done the final squash with codin agent mark

xwxb:
sorry for the UT miss, any further ops should I perform?

killergerbah:
@xwxb All good I will make final tweaks then merge

PR #965, opened 2026-04-21
Add word browser

ShanaryS:
The goal is to expose the db data for user viewing and allow bulk operations (update status/states and deleting) on local tokens. Anki tokens are read-only since they are synced.
- Displays all data stored in DB with filtering and sorting
- Allows bulk deletion and updates (only status/states for local tokens)
- Multi select supports Ctrl+Click and Shift+Click sematics
  - Selection is only preserved for the visible rows, discarded if view changes (for maximum safety)
- Search handles lemma lookups (requires Yomitan) and other QOL features
  - Results ordered by exact first, regardless of sorting
  - Allows romaji input to search hiragana, all caps required for katakana lookup (in addition to the ascii text)
  - Performs some other low effort convertions for common latin and german shorthands (probably far more could be done here)
- Checkboxes are multi-tiered, allowing explict includes and excludes for each item
- Card Id and Note Id lookups are by substring, supporting multiple OR based lookups by any non-digit delimiter
- All sorting and filtering is done on the UI side to reduce coupling for better version mismatch support
  - Debouce timer on all relevant actions tuned for the best experience
  - Local tokens are considered unsuspended for filtering but omitted from UI since it doesn't really apply (track also omitted since they are global)

https://github.com/user-attachments/assets/b1ce5345-d945-460e-adfd-126993e44a23


common/util/util.ts
```diff
@@ -424,6 +426,69 @@ export function isKatakanaOnly(text: string) {
     return KATAKANA_ONLY_REGEX.test(text.normalize('NFC'));
 }
 
+// FROM: https://github.com/yomidevs/yomitan/blob/3a7c5e524d9ae4d2f7f1a12a1dd95d8172dc21ed/ext/js/language/ja/japanese-wanakana.js#L25
+export function convertToHiragana(text: string) {
```
killergerbah: I don't think we can include this code since it's GPL-3 licensed. We need to find a way to get similar behavior without including GPL code.

common/util/util.ts
```diff
@@ -695,3 +760,339 @@ export class AsyncSemaphore {
         }
     }
 }
+
+// FROM: https://github.com/yomidevs/yomitan/blob/3a7c5e524d9ae4d2f7f1a12a1dd95d8172dc21ed/ext/js/language/ja/japanese-kana-romaji-dicts.js#L20
+export const ROMAJI_TO_HIRAGANA = {
```
killergerbah: Same here, even with the attribution above including this code automatically means we need to relicense the project as GPL-3, which isn't impossible but very difficult as it requires permission from all contributors to the project.

common/util/util.ts
killergerbah: Only possibilities I can think of:
- Rewrite from brain
- Include similar behavior from an open source project without a copyleft license

killergerbah:
Great work, all the logical parts look good to me.
- I think I would just want to smooth out some of the UI/UX in the new dialog. For example, I think there's probably a standard way of displaying filter options for each column. And there might be a way to compress the "apply state updates" form into something more efficient for the one "ignored" state that currently exists.
- Need to find a suitable replacement for the GPL-3-licensed bits

ShanaryS:
> Need to find a suitable replacement for the GPL-3-licensed bits

That was my bad, I forgot to check the license and assumed it was permissive. This code is only useful for users without Yomitan connected at the time of using the `Word Browser`. Yomitan will handle the romaji conversion when getting the lemmas as the lifted code itself suggests. So we could:
1. Remove this code and rely on Yomitan, would probably tokenize as well so we can search the exact token match. I also assume Yomitan will handle these QOL things for all languages since it probably needs to.
3. Use Wanakana.js which Yomitan moved from previously
4. Lift the code from Wanakana.js since it is MIT licensed.

> I think there's probably a standard way of displaying filter options for each column.

How do you mean? Standard as in a design principles or between the filter options themselves.

> And there might be a way to compress the "apply state updates" form into something more efficient for the one "ignored" state that currently exists.

I think something like `No change`, `Add: Ignored`, and `Remove: Ignored` would work. It would also allow keeping the same localization which would allow future expansion easily.

killergerbah:
>Remove this code and rely on Yomitan, would probably tokenize as well so we can search the exact token match. I also assume Yomitan will handle these QOL things for all languages since it probably needs to.

This seems most reasonable to me, it would be hard to imagine that someone using this feature doesn't have Yomitan API running

>How do you mean? Standard as in a design principles or between the filter options themselves.

Was thinking design principles, as this seems like a kind of table that most UI frameworks should have solved

> I think something like No change, Add: Ignored, and Remove: Ignored would work. It would also allow keeping the same localization which would allow future expansion easily.

Yeah that would be simpler. Could Add/Remove be expressed via a checkbox?

ShanaryS:
I made all the changes, I also added pagination with the controls for it in the bottom left. Feel free to change as you see fit and merge when ready.

killergerbah:
@ShanaryS The updates look really good
I took another look at this, and thought it might be more natural for the word browser button to be above the Annotation track selector, closer to where all the word-DB-related stuff is. 
Was there a specific reason for the word browser to be track-specific by default? Could it instead be a filter inside the word browser?

ShanaryS:
We would need to use all the configured Yomitans when doing the search but it shouldn't be a problem. I think the design also assumes a single track.

killergerbah:
I see, then I'm thinking of making these changes, but I probably won't have time until the weekendish.
- Move word browser button above annotation section. Move track selector into the word browser itself. TBD whether it should be possible to browser words from multiple tracks at once. If we allow viewing words from multiple tracks, we would probably need to change the dictionary DB fetch method signatures.
- Refactor bulk modification UI. It serves a very specific purpose so I think it should be summoned only when necessary, or somehow take up less space up top.
- Minor rearrangements: move refresh button into the toolbar to free up some space, move the "showing.." label to the bottom

ShanaryS:
I implemented the first one, I'll leave the other two to you.

The db function now allows `undefined` for the track which returns data for all tracks. All track Yomitans are used if that track has any db entries, regardless if currently enabled. The status colors for local tokens defaults to the first track, but otherwise each track uses their configured colors.

killergerbah:
Sweet thanks

killergerbah:
@ShanaryS Made approximately the proposed changes, let me know what you think, will plan to merge soon though

ShanaryS:
Looks good!

PR #976, opened 2026-04-29
add filter for 1 Unknown and 2+ Unknown in sentence stats

### LINKED ISSUE #974: Add I+1 (sentences with one unknown word) as a selectable statistic
This is similar to the selectable `1 Uncollected` option under `Sentence statistics`.  `1 Uncollected` and `2+ Uncollected`  only looks for sentences with uncollected tokens, which as the name implies are tokens that are uncollected and the user has not added a status to the token. 

i+1, i+2, etc.. as defined by Stephen Krashen in his [Input hypothesis ](https://en.wikipedia.org/wiki/Input_hypothesis)

> Krashen called this level of input "i+1", where "i" is the learner's [interlanguage](https://en.wikipedia.org/wiki/Interlanguage) and "+1" is the next stage of language acquisition.

The `i` in this situation would be tokens that are all "known" in our case that would be `Graduated` `Young` and/or `Mature`. I propose to let the user decide what tokens would the `i` refer to since everyone marks words differently.

The `+1` `(+2..3...etc)` would refer to the "unknown" token. Which would most likely apply to `Unknown` and/or `Learning`. Once again I believe leaving it up the users choice to select which statuses would refer to +1

This feature is already present in [Migaku](https://migaku.com/faq/getting-started)  labeled as `sentences with one unknown word` and `sentences with multiple unknown words` This would probably be better names for i+1 and i+2. I only described it as such, since most of the people using ASBplayer are following input methods/guides which most times describes this as i+1. 

Image

The reason I'm requesting this is that after some recent changes `1 Uncollected` behaves at it should and no longer displays tokens that are not `uncollected`. I came across this as in this video I have no `1 uncollected` subtitle lines. But I do have a lot of sentences which have words that are marked as unknown. 

Image

I do believe in keeping `i+1` and `1 uncollected` separate as its referred as `i+1` under `Projected re watch` in the picture above. Keeping them separate would aid in mining

ShanaryS:
fixes #974

This adds a filtered view for `1 Unknown` and `2+ Unknown` for the sentence stats. These views will not contain any uncollected tokens.

Assisted-by: Copilot:GPT-5.4
- [x] I have read the [contribution guidelines](https://github.com/killergerbah/asbplayer/blob/main/CONTRIBUTING.md).

### DIFF
--- a/common/components/Statistics.tsx
+++ b/common/components/Statistics.tsx
@@ -22,7 +22,7 @@ import {
     sentenceComprehensionPointLabel,
     sentenceComprehensionXAxisLabels,
     sentenceDialogBucketData,
-    uncollectedSentenceBucketLabel,
+    statusSentenceBucketLabel,
 } from '@project/common/dictionary-statistics';
 import { AsbplayerSettings, dictionaryTrackEnabled, TokenStatus } from '@project/common/settings';
 import StatisticsSentenceDetailsDialog from '@project/common/components/StatisticsSentenceDetailsDialog';
@@ -108,6 +108,7 @@ const emptySentenceBuckets: DictionaryStatisticsSentenceBuckets = {
         entries: [],
     },
     uncollected: [],
+    unknown: [],
 };
 const distributionBarHeight = 6;
 const sentenceComprehensionGraphHeight = 120;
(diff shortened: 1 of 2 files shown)


killergerbah:
I looked at usages of `evaluateSentenceSnapshot` and couldn't find where it's possible for the unknown token count to be incremented. I guess we'd need to remove the filters in some places?

ShanaryS:
I'm not sure I fully understand. Do you mean for the projected rewatches? I assume that all `Unknown` tokens have been reviewed including the newly collected `Uncollected` i+1 tokens. I feel this better aligns with mining the content on the first pass, review, then mining again which admittedly isn't clear since a rewatch is never defined.

killergerbah:
My bad, I wasn't sure which code paths were triggering the new code paths for unknown tokens. I can see it now, and I was able to confirm it working too.

Unknown transitioning into a higher status and not being included in the projected rewatch makes sense

PR #986, opened 2026-05-08
feat: restore last local file session with File System Access API

### LINKED ISSUE #975: [Feature Request] Remember & restore last local video + subtitle file pair in webapp
### Is your feature request related to a problem?

When using the asbplayer **webapp** with local video files, I often find myself mining cards across multiple sessions. If I close the tab or browser and come back later, I have to manually re-select both the video and subtitle files from my local drive. Even though my mining history is preserved, the repetitive file-selection steps feel painful, especially when the files are buried deep in a directory tree.

### Describe the solution you'd like

A way to **restore the last-used local video + subtitle combination** with a single click (or automatically on load).

One possible implementation:

1. When the user selects local files via a new file-picker flow, persist the [`FileSystemFileHandle`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle) objects to **IndexedDB**.
2. On the next visit, show a **"Restore last session"** button on the landing page.
3. When clicked, call [`FileSystemHandle.requestPermission()`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle/requestPermission) to re-acquire user consent (required by browser security policy), then resolve the handles back to `File` objects and load them into the player.

This is conceptually similar to how some local media players remember recently-opened files.

### Describe the drawbacks / open questions

I want to be upfront about the downsides so this can be evaluated openly:

| Concern | Details |
|---------|---------|
| **Chrome-only (Chromium-only)** | The File System Access API is **not available in Firefox or Safari**. We would need to maintain two file-loading paths: the existing `<input type="file">` / drag-and-drop for universal support, plus the new handle-based path for Chromium browsers. |
| **Permission UX friction** | Even with stored handles, the browser **requires a user gesture** to re-request permission every session. A silent auto-restore is impossible; the best we can do is a one-click restore button. If the user denies permission, we need a graceful fallback. |
| **Handle invalidation** | If the user moves, renames, or deletes the file, the stored handle becomes stale. We need clear error messaging for this case. |
| **Scope creep** | asbplayer supports multiple subtitle files, audio-only files, and directory drops. Extending "restore" to all these cases adds complexity. We might need to limit the scope to "one video + one or more subtitles from the same session." |
| **Maintenance burden** | Introducing a browser-specific API means the feature cannot be uniformly tested, and the fallback path must be kept working indefinitely. |

### Describe alternatives you've considered

1. **Do nothing** — Keep the current explicit file selection. This is perfectly functional; the pain point is only about convenience.
2. **Remember file paths as strings** — Store just the file names/paths in `localStorage`. This avoids the Chrome-only API, but the browser cannot resolve a path string back to a `File` object without the user manually picking the file again, so it doesn't actually save clicks.
3. **Use a native wrapper / PWA with file handling/ Extension file api** — More powerful, automatic, but far outside the scope of the current webapp architecture.

### Additional context

- I personally use the webapp with local files frequently, so this would save me a lot of repetitive clicks. However, I completely understand if the Chrome-only limitation makes this a non-starter for the project.
- I'm opening this mainly for **public discussion** — I don't feel strongly that this *must* be implemented, but I wanted to surface the idea and its trade-offs in case others share the same pain point.
- If the consensus is positive and the design is agreed upon, I'd be happy to help implement it.

xwxb:
## Summary

- Add a File System Access-based file loading path behind the existing `browse` entry to persist local `FileSystemFileHandle`s across reloads.
- Introduce an IndexedDB repository for a single latest file session record and merge partial selections (e.g., subtitles only) into that session.
- Add a landing-page action to restore the last session, including permission re-check and fallback error handling when files are moved or deleted.
- Normalize extension parsing to lowercase, centralize file input accept extensions in the new `file-system-access` module, and add missing localized copy for restore-related errors and actions.

Closes https://github.com/asbplayer/asbplayer/issues/975

## Implementation Notes

- New module `common/file-system-access` encapsulates:
  - capability detection (`supportsFileSystemAccess`)
  - picker invocation (`showFilePicker`)
  - permission flow (`requestPermissions`)
  - handle-to-file resolution (`resolveFiles`)
  - media/subtitle extension classification helpers
- `App` keeps the previous `<input type="file">` fallback for non-FSA environments.
- Session restoration is intentionally gated by explicit user action from the landing page, with a separate visibility check on mount (`canRestoreLastSession`).
- Repository initialization is done once per mount via lazy state initialization so identity is stable for effects/callbacks.

## Discussion Topics

1. **Why a new file loading path is necessary**
   - Standard `<input type="file">` and drag-and-drop can load files immediately, but they do not provide durable `FileSystemFileHandle`s.
   - Without handles, there is nothing to persist across reloads. Only the File System Access API (plus IndexedDB for handle serialization) makes one-click local session restoration possible.
   - This is the core reason we introduce a new loading path rather than extending the existing input flow.

2. **Extension classification moved to `file-system-access`**
   - The previous `INPUT_ACCEPT_FILE_EXTENSIONS` constant and the extension-to-type mapping were hard-coded in `App.tsx`.
   - With the addition of FSA-specific picker types, the extensions are now centralized in `common/file-system-access/file-system-access.ts`.
   - This means `App.tsx` imports `inputAcceptFileExtensions` rather than defining its own list. If we expect these lists to diverge (e.g., picker types vs. `<input accept>`), we may want to split them again.

3. **Restore action placement / UX weight**
   - Current placement is in the lower action area as an outlined button as suggested in https://github.com/asbplayer/asbplayer/issues/975#issuecomment-4348157600.
   - Alternative: move this trigger closer to the main CTA text (`Drag and drop... or browse`) and make it a lighter inline link-style action for reduced visual weight.

4. **`handleFileSelector` is now async / event handler returns `Promise<void>`**
   - `handleFileSelector` was changed from `() => void` to `async () => Promise<void>` to support the FSA picker flow.
   - React event handlers do not await async functions; if the component unmounts while the picker is open, a subsequent `setState` (e.g., `saveFileSession` → `setCanRestoreLastSession`) could warn about updates on an unmounted component.
   - Options: add an `isMounted` guard, move the async work into a `useEffect`-like pattern, or accept the low-probability warning given the current landing-page architecture.

## Optional Design Notes (For Deeper Review)

- **Merge design for partial selections**
  - `IndexedDBFileSessionRepository.merge` preserves the previous video handle when the user picks only subtitles, and vice versa, so the saved session always reflects the latest complete set.

- **Repository lifetime (`useState` lazy init vs. `useMemo`)**
  - The repository is created once per mount with lazy `useState` initialization instead of `useMemo`, making the one-per-mount lifecycle explicit and the reference stable for effects and callbacks.

### DIFF
--- a/common/app/components/App.tsx
+++ b/common/app/components/App.tsx
@@ -51,6 +51,7 @@ import { useTranslation } from 'react-i18next';
 import { LocalizedError } from './localized-error';
 import { DisplaySubtitleModel } from './SubtitlePlayer';
 import { useCopyHistory } from '../hooks/use-copy-history';
+import { useFileSession } from '../hooks/use-file-session';
 import { useI18n } from '../hooks/use-i18n';
 import { useAppKeyBinder } from '../hooks/use-app-key-binder';
 import { useAnki } from '../hooks/use-anki';
@@ -60,6 +61,7 @@ import { useAppWebSocketClient } from '../hooks/use-app-web-socket-client';
 import { LoadSubtitlesCommand } from '../../web-socket-client';
 import { ExtensionBridgedCopyHistoryRepository } from '../services/extension-bridged-copy-history-repository';
 import { IndexedDBCopyHistoryRepository } from '../../copy-history';
+import { supportsFileSystemAccess, showFilePicker, requestPermissions, resolveFiles } from '../../file-system-access';
 import { isMobile } from 'react-device-detect';
 import { GlobalState } from '../../global-state';
 import mp3WorkerFactory from '../../audio-clip/mp3-encoder-worker.ts?worker';
@@ -75,9 +77,6 @@ import StatisticsOverlay, { StatisticsOverlayProps } from '../../components/Stat
 const latestExtensionVersion = '1.15.0';
 const extensionUrl =
     'https://chromewebstore.google.com/detail/asbplayer-language-learni/hkledmpjpaehamkiehglnbelcpdflcab';
-
-const INPUT_ACCEPT_FILE_EXTENSIONS =
-    '.srt,.ass,.vtt,.sup,.mp3,.m4a,.aac,.flac,.ogg,.wav,.opus,.mkv,.mp4,.avi,.m4v,.webm';
 
 const useContentStyles = makeStyles<Theme, ContentProps>((theme) => ({
     content: {
@@ -97,6 +96,67 @@ const useContentStyles = makeStyles<Theme, ContentProps>((theme) => ({
     }),
 }));
 
+const videoExtensions = ['.mkv', '.mp4', '.m4v', '.avi', '.webm'] as const;
+const audioExtensions = ['.mp3', '.m4a', '.aac', '.flac', '.ogg', '.wav', '.opus', '.m4b'] as const;
+const subtitleExtensions = [
+    '.srt',
+    '.ass',
+    '.vtt',
+    '.sup',
+    '.nfvtt',
+    '.ytxml',
+    '.ytsrv3',
+    '.dfxp',
+    '.ttml2',
+    '.bbjson',
+] as const;
+const inputAcceptFileExtensions = [...subtitleExtensions, ...audioExtensions, ...videoExtensions].join(',');
+
+const VIDEO_EXT_SET = new Set<string>(videoExtensions);
+const AUDIO_EXT_SET = new Set<string>(audioExtensions);
+const SUBTITLE_EXT_SET = new Set<string>(subtitleExtensions);
+
+const getExtension = (fileName: string) => {
+    const index = fileName.lastIndexOf('.');
+    return index === -1 ? '' : fileName.substring(index).toLowerCase();
+};
+
+async function extractDropFileHandles(items: DataTransferItemList): Promise<FileSystemFileHandle[] | undefined> {
+    const handlePromises: Promise<any>[] = [];
+
+    for (let i = 0; i < items.length; ++i) {
+        const item = items[i];
+        if (item.kind !== 'file') {
+            continue;
+        }
+
+        // Persist dropped handles only when the browser exposes getAsFileSystemHandle.
+        const getAsFileSystemHandle = (item as any).getAsFileSystemHandle as (() => Promise<any>) | undefined;
+        if (typeof getAsFileSystemHandle !== 'function') {
+            return undefined;
+        }
+
+        // Capture handle promises synchronously before DataTransfer gets cleared.
+        handlePromises.push(getAsFileSystemHandle.call(item));
+    }
+
+    const handles: FileSystemFileHandle[] = [];
+    for (const handlePromise of handlePromises) {
+        try {
+            const handle = await handlePromise;
+            if (handle?.kind === 'file') {
+                handles.push(handle as FileSystemFileHandle);
+            }
+        } catch (e) {
+            // Best-effort only; if handle access fails, keep loading dropped files normally.
+            console.warn('Failed to read dropped file handle:', e);
+            return undefined;
+        }
+    }
+
+    return handles.length > 0 ? handles : undefined;
+}
+
 function extractSources(files: FileList | File[]): MediaSources {
     let subtitleFiles: File[] = [];
     let audioFile: File | undefined = undefined;
@@ -104,51 +164,28 @@ function extractSources(files: FileList | File[]): MediaSources {
 
     for (let i = 0; i < files.length; ++i) {
         const f = files[i];
-        const extensionStartIndex = f.name.lastIndexOf('.');
+        const extension = getExtension(f.name);
 
-        if (extensionStartIndex === -1) {
+        if (extension === '') {
             throw new LocalizedError('error.unknownExtension', { fileName: f.name });
         }
 
-        const extension = f.name.substring(extensionStartIndex + 1, f.name.length);
-        switch (extension) {
-            case 'ass':
-            case 'srt':
-            case 'vtt':
-            case 'nfvtt':
-            case 'sup':
-            case 'ytxml':
-            case 'ytsrv3':
-            case 'dfxp':
-            case 'ttml2':
-            case 'bbjson':
-                subtitleFiles.push(f);
-                break;
-            case 'mkv':
-            case 'mp4':
-            case 'm4v':
-            case 'avi':
-            case 'webm':
-                if (videoFile) {
-                    throw new LocalizedError('error.onlyOneVideoFile');
-                }
-                videoFile = f;
-                break;
-            case 'mp3':
-            case 'm4a':
-            case 'aac':
-            case 'flac':
-            case 'ogg':
-            case 'wav':
-            case 'opus':
-            case 'm4b':
-                if (videoFile) {
-                    throw new LocalizedError('error.onlyOneAudioFile');
-                }
-                videoFile = f;
-                break;
-            default:
-                throw new LocalizedError('error.unsupportedExtension', { extension });
+        if (SUBTITLE_EXT_SET.has(extension)) {
+            subtitleFiles.push(f);
+        } else if (VIDEO_EXT_SET.has(extension)) {
+            if (videoFile) {
+                throw new LocalizedError('error.onlyOneVideoFile');
+            }
+            videoFile = f;
+        } else if (AUDIO_EXT_SET.has(extension)) {
+            if (videoFile) {
+                throw new LocalizedError('error.onlyOneAudioFile');
+            }
+            videoFile = f;
+        } else {
+            throw new LocalizedError('error.unsupportedExtension', {
+                extension: extension.startsWith('.') ? extension.substring(1) : extension,
+            });
         }
     }
 
@@ -350,6 +387,7 @@ function App({
     const [isSidePanelOpen, setIsSidePanelOpen] = useState<boolean>(false);
     const [statisticsOverlayOpen, setStatisticsOverlayOpen] = useState<boolean>(false);
     const [statisticsOverlayDismissed, setStatisticsOverlayDismissed] = useState<boolean>(false);
+    const { canRestoreLastSession, saveSession: saveFileSession, fetchSession, clearSession } = useFileSession();
     const [lastError, setLastError] = useState<any>();
     const fileInputRef = useRef<HTMLInputElement>(null);
     const { subtitleFiles } = sources;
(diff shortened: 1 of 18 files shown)


common/app/components/App.tsx
```diff
@@ -958,14 +959,82 @@ function App({
...
                 console.error(e);
                 handleError(e);
+                return false;
             }
         },
         [handleError]
     );
 
+    // === File System Access: incrementally merge handles into the saved session ===
+    const saveFileSession = useCallback(
```
killergerbah: The new callbacks and effects look like great candidates to be extracted into a single hook that provides these as utility functions and state, similar to what is being done in `useCopyHistory`. This would help keep `App`'s line count down.

xwxb: refactored in 4f4f444c, mostly use `useFileSession` hook to encapsulate, and kept permission requests, handleFiles coordination, and error handling in App.tsx to prevent the hook from being overly coupled with app-specific logic.

common/app/hooks/use-file-session.ts
```diff
@@ -0,0 +1,70 @@
+import { useCallback, useEffect, useState } from 'react';
+import {
+    IndexedDBFileSessionRepository,
+    isMediaExtension,
+    isSubtitleExtension,
+    supportsFileSystemAccess,
+} from '../../file-system-access';
+
+export const useFileSession = () => {
+    const [fileSessionRepository] = useState<IndexedDBFileSessionRepository | undefined>(() =>
```
killergerbah: This doesn't change so you can store it outside of React state in this file. This way the initialization function won't run every time the app re-renders.

common/file-system-access/file-system-access.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * File System Access API helpers (Chrome-only).
+ * Provides showOpenFilePicker-based file selection that returns FileSystemFileHandle objects,
+ * and utilities to re-acquire permissions and resolve handles back to File objects on revisit.
+ */
+
+const videoExtensions = ['.mkv', '.mp4', '.m4v', '.avi', '.webm'] as const;
```
killergerbah: This is a second copy of supported extensions in addition to the switch-cases we already have in `App.tsx`. I would suggest removing all extension-related logic here and having the app specify which files are subtitles, videos, or audio instead, so that the logic for determining file types remains in one place.

This would mean that the `saveSession` signature needs to change so that the caller specifies the subtitle and video files from outside.

common/file-system-access/file-system-access.ts
```diff
@@ -0,0 +1,118 @@
...
+            files.push(await handle.getFile());
+        } catch {
+            errors.push(handle);
+        }
+    }
+
+    return { files, errors };
+}
+
+export async function showFilePicker(): Promise<FileSystemFileHandle[] | undefined> {
```
killergerbah: As consequence of the comment above, this function can change so that the caller specifies which extensions are audio/video/text.

common/app/components/App.tsx
```diff
@@ -1244,7 +1259,24 @@ function App({
...
-    const handleFileSelector = useCallback(() => fileInputRef.current?.click(), []);
+    const handleFileSelector = useCallback(async () => {
+        if (supportsFileSystemAccess()) {
+            try {
+                const handles = await showFilePicker();
+                if (!handles || handles.length === 0) return;
+                const { files } = await resolveFiles(handles);
+                if (files.length === 0) return;
+                if (handleFiles({ files })) {
+                    await saveFileSession(handles);
```
killergerbah: If it's possible you should add handling for drag-and-drop as well. Also, moving `saveFileSession` into `handleFiles` should result in the least amount of change. File-reading paths can simply give `FileSystemHandles` to `handleFiles` if they are available. I would also suggest re-using the merge logic in `handleFiles`, to avoid any possibility that the custom `merge` in `FileSessionRepository` results in different behavior.

common/app/hooks/use-file-session.ts
xwxb: Fixed in 35d010bf. Although I used a lazy initializer in useState to ensure it only runs once, I agree that using React state (or useRef) for a constant singleton isn't the most idiomatic approach. I've refactored it to a module-level singleton. Thanks!

common/file-system-access/file-system-access.ts
xwxb: I missed the `switch` statement in `App.tsx` earlier and mistakenly thought consolidating extensions in the lower layer would be cleaner. However, I agree that maintaining this logic at the application level is more appropriate for decoupling.

I've refactored this in b3f7dca by uplifting the extension configuration to `App.tsx` and making the underlying file picker utility generic. The implementation might still be a bit aggressive—please let me know what you think. I'll address the drag-and-drop persistence in a separate update when I'm available next.

killergerbah: The refactor looks good, thanks

common/app/components/App.tsx
xwxb: Addressed in b3f7dca1.

Agreed that drag-and-drop should persist file handles — implemented that and passed local test. However, I spent a good amount of time today carefully evaluating the suggestions around merging `saveFileSession` into `handleFiles` and unifying the merge logic.

I actually wrote a version that did exactly that (adding `fileHandles` to `handleFiles`'s signature, running `saveFileSession` inside it), and it immediately broke drag-and-drop: the `handleDrop` callback became `async` to `await` handle extraction, which caused `DataTransfer.files` to be empty after the async boundary — files silently failed to load with no errors. Beyond the bug, the enlarged signature and mixed sync/async concerns made `handleFiles` noticeably harder to reason about.

After that experience I settled on the current approach:
- `handleFiles` stays focused on `File` loading only (sync, no handle awareness).
- A new `persistFileSessionHandles` callback handles the classification and best-effort persistence, called by each entry point (`handleFileSelector`, `handleDrop`) after a successful `handleFiles` call.
- `extractDropFileHandles` acquires handles from `DataTransferItem` on Chromium and silently falls back to `undefined` on unsupported browsers.

Regarding the merge logic: I decided to keep it in `FileSessionRepository` rather than pulling it up. The two "merges" are fundamentally different — `handleFiles` merges in-memory UI state (what's currently loaded for playback), while the repository merges persisted handles (what to restore on next visit). They operate on different data types at different lifecycle points, so sharing an abstraction would couple them without real benefit.

That said, this is where I landed today — if you see anything I might have missed or have further suggestions, I'd appreciate hearing them.

common/app/components/App.tsx
```diff
@@ -1226,13 +1298,25 @@ function App({
...
             }
 
-            if (e.dataTransfer.items && e.dataTransfer.items.length > 0 && allDirectories(e.dataTransfer.items)) {
-                handleDirectory(e.dataTransfer.items);
-            } else if (e.dataTransfer.files && e.dataTransfer.files.length > 0) {
-                handleFiles({ files: e.dataTransfer.files });
+            if (dataTransfer.items && dataTransfer.items.length > 0 && allDirectories(dataTransfer.items)) {
+                handleDirectory(dataTransfer.items);
+            } else if (dataTransfer.files && dataTransfer.files.length > 0) {
+                // Snapshot file objects before any async work; drag DataTransfer can be cleared after event unwinds.
```
killergerbah: This comment is either out-of-date or maybe just not very clear. I guess it's referring to the files disappearing from the data transfer object if they're not handled in the same callstack? Should the comment be reworded or moved?

common/app/components/App.tsx
```diff
@@ -958,14 +968,75 @@ function App({
...
+                } else if (SUBTITLE_EXT_SET.has(extension)) {
+                    subtitleHandles.push(handle);
+                }
+            }
+
+            if (!videoHandle && subtitleHandles.length === 0) {
+                return;
+            }
+
+            // Saving file handles should not break the active load flow; merge remains in repository.
```
killergerbah: This comment as well, I don't think it'll be useful to any developers reading it unless they were closely involved in the refactor that's in this PR

common/app/components/App.tsx
```diff
@@ -97,58 +96,90 @@ const useContentStyles = makeStyles<Theme, ContentProps>((theme) => ({
...
+            return undefined;
+        }
+
+        try {
+            const handle = await getAsFileSystemHandle.call(item);
+            if (handle?.kind === 'file') {
+                handles.push(handle as FileSystemFileHandle);
+            }
+        } catch (e) {
+            // Handle acquisition is best-effort; never block the normal drop flow if this fails.
```
killergerbah: What's a normal drop flow? Drag-and-drop?

common/app/components/App.tsx
```diff
@@ -97,58 +96,90 @@ const useContentStyles = makeStyles<Theme, ContentProps>((theme) => ({
...
+        }
+
+        // Chromium exposes getAsFileSystemHandle; other browsers should quietly fall back to file-only loading.
+        const getAsFileSystemHandle = (item as any).getAsFileSystemHandle as (() => Promise<any>) | undefined;
+        if (typeof getAsFileSystemHandle !== 'function') {
+            return undefined;
+        }
+
+        try {
+            const handle = await getAsFileSystemHandle.call(item);
```
killergerbah: There seems to be a bug here. When dragging-and-dropping two subtitle files, only the first handle is obtained. I think this is due to the `await` which as you pointed out, causes the `DataTransferItemList` to get cleared out.

common/app/components/App.tsx
xwxb: the old wording was too vague. I rephrased the comment to explicitly say we copy `DataTransfer.files` synchronously because the drag payload may be cleared after the drop handler returns.

common/app/components/App.tsx
xwxb: Agree. “merge remains in repository” was too implementation-specific for future readers, so I simplified the comment to focus on behavior: session persistence runs in background and does not block file loading.

common/app/components/App.tsx
xwxb: Makes sense, I rewrote it to explicitly say we keep dropped-file loading working even if handle acquisition fails.

common/app/components/App.tsx
xwxb: Good catch. I used to not note that dir load and multi-sub load is supported. indeed awaiting `getAsFileSystemHandle` inside the loop can lose later items when `DataTransferItemList` gets cleared. I changed it in 0dea46b to capture all handle promises synchronously first, then await them, so multi-file drops can persist all available handles. And test locally to ensure mult-sub handle is stored in indexdb.

killergerbah (approved):
Awesome, thank you for the great idea and implementation

xwxb:
@killergerbah 
Thank you for the patient and thorough review — I learned a lot from the feedbacks. Squashed and force-pushed. Ready to merge whenever you see fit.

PR #990, opened 2026-05-09
Fix UR Play subtitle detection, share HLS helper across sites

jprostko:
### Summary

UR Play's subtitle detection broke after a recent site change where the Next.js Page Router that was previously utilized was changed to the Next.js App Router with RSC. Instead of adapting to the Next.js change, this pull request replaces its entrypoint with the XHR-intercept-and-parse-HLS pattern that already powers Yle Areena, and promotes that helper to `pages/m3u8-util.ts` so it can be shared. Also fixes two extension-extraction bugs that UR Play's JWT-signed URLs surfaced.

### Refactor

- Move `inferTracksFromInterceptedM3u8` from `entrypoints/areena-yle-page.ts` to `pages/m3u8-util.ts`. Yle Areena and UR Play import from there, and both entrypoints collapse to a regex and a `defineUnlistedScript` call.
- Generalize `subtitleTrackSegmentsFromM3U8` to iterate every SUBTITLES media group, not just `sub-main`. Same generalization 824d2f9e applied inline to Yle Areena. Lifting it lets `inferTracksFromInterceptedM3u8` call into it instead of carrying its own copy. Disney+ unaffected if its manifests only expose `sub-main`.
- `subtitleTrackSegmentsFromM3U8` returns `[]` instead of rejecting when SUBTITLES are absent. Disney+ never caught the rejection, so the prior path surfaced as an unhandled rejection in `inferTracks.onRequest`.
- Upgrade `baseUrlForUrl` to URL-parse before locating the filename slash, so query strings containing `/` aren't mistakenly treated as part of the path.

### Bugfix

UR Play's CDN signs segment URLs with JWTs in the query string. JWTs are three dot-separated base64 parts, which broke two spots deriving file extensions via `url.lastIndexOf('.')`:

- `extractExtension` in `pages/util.ts`: found the last `.` inside the JWT signature instead of before the file extension. Rewritten to strip the query string and fragment first. Output unchanged for URLs without dots in their query or fragment.
- The array branch of `_subtitlesForUrl` in `controllers/video-data-sync-controller.ts`: had its own inline `lastIndexOf('.')`, ignoring the caller's `extension` parameter. Switched to `extractExtension(firstUri, extension)` so the caller's extension is the fallback.

### Cleanup

- `m3u8-util.ts` had two imports of `m3u8-parser` (`Parser` and `Parser as m3U8Parser`), present since 1e1eecc7. Removed the duplicate.
- `areena-yle-page.ts` had a dead `subtitleTrackSegmentsFromM3U8` import that 824d2f9e left behind. Removed.
- `areena-yle-page.ts` had a dead exported `Playlist` interface that nothing imports. Removed.

### Tested

Helper consumers:

- [x] UR Play: cold load `/serie/`, cold load `/program/`, SPA navigation between episodes, single-language and multi-language programs. Engelska and Svenska tracks load and display. Site works correctly again.
- [x] Disney+: works identically as v1.16.0
- [ ] Yle Areena: unable to test due to strict Finland residency requirement.

Spot-checks for the `extractExtension` / controller changes (sites whose URLs may or may not contain dots in query strings):

- [x] Hulu: works identically as v1.16.0
- [x] Prime Video: works fine with #991 applied

Other sites such as Netflix, YouTube, NRK TV, SVT Play, Plex, etc. should be unaffected by this changeset.

### Additional notes

`yarn run verify` ran successfully with no errors after all of the changes made in this pull request.  Both the Chrome and Firefox extensions were tested with the various sites mentioned above.

### Contributor guideline acceptance


killergerbah:
Excellent work, thank you. I tested on Yle Areena and it's working with this change.

killergerbah (approved):

PR #991, opened 2026-05-09
Fix Amazon Prime subtitle detection after fetch migration

### LINKED ISSUE #988: Amazon Prime subtitle detection not working
From persist3nt:
>So, after the latest update, I can't seem to capture the video IDs of shows on Amazon.co.jp's Prime Video anymore. My app version is 40cb184 and my extension version is 1.16.0. Chrome, Windows 11, all latest patches. Any ideas? 

From jprostko:
> I suspect the site changed.  I quickly tried one of the 1.15 snapshots and it exhibited the same behavior.  I suspect it isn't a code regression, but I admittedly didn't dig into it at all,  since it is getting late here.  But yeah, that entrypoint hasn't been changed in almost a year.

jprostko:
### Summary

Amazon Prime Video's player migrated its catalog and playback-resource API calls from `XMLHttpRequest` to `fetch()`. The page script only monkey-patched `XMLHttpRequest.prototype.open`/`send`, so those calls bypassed the interceptor entirely. `lastEntityId` was never set, the metadata poll timed out, and asbplayer reported "Could not capture metadata" on every Prime title.

### Changes

- Extracted the URL-match logic into a `captureRequestUrl()` helper shared by the XHR and fetch interceptors.
- Added a `window.fetch` patch that mirrors the existing XHR patch. When the body is a string in `init.body` (current Prime behavior), capture it directly. When the body lives on a `Request` object, clone the request and read the body asynchronously.
- Set `Content-Type: application/json` on the re-fired POST in `tracksFromUrl`. The new `/playback/prs/GetVodPlaybackResources` endpoint rejects the default `text/plain;charset=UTF-8` body and replies with an error rather than the subtitle list.
- Detect when the response indicates a stale session (`timedTextUrls.error` set instead of `timedTextUrls.result`) and throw, routing through the existing catch path. This previously produced a silent empty picker.
- Updated the user-facing error messages to give an actionable fix ("refresh this page and start playback") with a parenthetical reason for debugging.

### Tested

Testing was done on Amazon Prime Video (in the US via amazon.com) with various movies/series and they work as desired.  The ViaPlay Prime Video add-on subscription was also tested and works after this pull request.  Could not test with other Amazon regional domains such as amazon.co.jp or amazon.se.

### Known limitation

When a user resumes a video on a page that has been open for a while, Amazon's player resumes from cached MediaSource state and does not refetch playback resources. The page script then has nothing to capture, so subtitle detection fails with one of the new error variants. This is likely a longstanding limitation that was not introduced by this PR. The error message now tells the user to refresh the page, which restores normal operation.

### Additional notes

`yarn run verify` ran successfully with no errors after all of the changes made in this pull request.  Both the Chrome and Firefox extensions were tested.

Fixes #988.

### Contributor guideline acceptance

### DIFF
--- a/extension/src/entrypoints/amazon-prime-page.ts
+++ b/extension/src/entrypoints/amazon-prime-page.ts
@@ -22,32 +22,44 @@ export default defineUnlistedScript(() => {
 
     let lastEntityId: string | undefined;
 
+    // Returns the captured titleId if the URL is a VOD playback resources request. The
+    // caller pairs the body with the same id when the request is actually sent.
+    const captureRequestUrl = (url: string): string | undefined => {
+        if (url.includes('GetVodPlaybackResources')) {
+            const titleId = urlParam(url, 'titleId');
+
+            if (titleId) {
+                const urls = metadataUrls[titleId] ?? {};
+                urls.vodPlaybackResourcesUrl = url;
+                metadataUrls[titleId] = urls;
+                lastEntityId = titleId;
+                return titleId;
+            }
+        }
+
+        if (url.includes('playerChromeResources') && url.includes('catalogMetadataV2')) {
+            const entityId = urlParam(url, 'entityId');
+
+            if (entityId) {
+                const urls = metadataUrls[entityId] ?? {};
+                urls.playerChromeResourcesUrl = url;
+                metadataUrls[entityId] = urls;
+                lastEntityId = entityId;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


killergerbah:
Thank you again for investigating this. I appreciate the very thorough contribution. I tested this and it works with some Amazon videos I bought a while back.

killergerbah (approved):

PR #997, opened 2026-05-11
add WaniKani as token source

ShanaryS:
This adds `WaniKani` as a source for tokens similar to how Anki used. Token priority is now `Local > Anki word / WaniKani > Anki sentence`.

I also refactored `dictionary-db.ts` as adding `WaniKani` made it become too large and hard to follow. There is now `dictionary-db-anki.ts` and `dictionary-db-wanikani.ts` which are functions only that take in the instance from `dictionary-db.ts` to run their respective pipelines. This better scales as well for future sources. I added our first migration for the database migration as well. Each future source should probably be a migration in a similar vain.

* Implemented `WaniKani` following their [best practices](https://docs.api.wanikani.com/20170710/#best-practices) and rate limiting
  * The endpoints we use are:
    * `/user`: Only to check if the token is valid on user action
    * `/subjects`: This is the content the user is studying. We only care about two types, `vocabulary` and `kana_vocabulary`.
    * `/assigments`: This operates on the subjects which does things such as schedule their reviews and SRS.
    * `/spaced_repetition_systems`: The SRS stages for `WaniKani` go from `0-9` which covers all our `TokenStatus`. This is customizable though per subject (in the API maybe not to the user) but we link the SRS system to the specific subject so that we handle it regardless.
    * `/resets`: Used to detect if the user has done a reset since last build. This completely destroys the cached responses for `/subjects`, `/assignments`, and `/spaced_repetition_systems`
  * Hidden subjects or assigments are ignored, I believe this means that `WaniKani` itself has deprecated these.
* The `WaniKani` db pipeline is largely similar to `Anki` however it's far more efficient since their API is designed to for heavy processing unlike `AnkiConnect` with tons of filters, mainly `updated_after` for our use case
  * DB Migration:
    *  Updates the `meta` store to support concurrent builds (between `Anki` and `WaniKani`)
    * Adds `subjectIds` to the `token` store, this works the exact same as `cardIds`
    * Adds two new stores `waniKaniSubjects` and `waniKaniAssigments` which are the caches of the `/subjects` and `/assigments` API endpoints per their best practice
  * Helper functions for the db pipelines are prefixed with `_` to show that they are private, as they need to be exported. The only public functions are those on the `DictionaryDB` class.
* Add build button for WaniKani in Annotation settings along with a WaniKani section to input API token
  * Token only needs read-only permissions, will only auto query on input change unlike page load since requests aren't local
  * Building without any token input will clear the db of WaniKani records similar to no field for Anki
  * Build has state changes similar to anki, however build states are per track since tokens are (anki is global so the db pipeline is slightly different)
* Annotation settings and WordBrowser is gated by `supportsDictionaryWaniKani()`
  * Outdated extensions work fine however outdated App instances will show the WaniKani records in the browser without the source (not worth addressing).
  * Updated the columns of `Note IDs` and `Card IDs` to `Note IDs / Subject IDs` and `Card IDs / Assignment IDs`
* `Anki word` and `WaniKani` tokens are considered equal
  * `getBulk` and `getAllTokens` chooses the highest status of the two (inline with other similar tiebreaks, these are more specific than `dictionaryTokenMatchStrategyPriority`) in order to keep the existing contract of an object of tokens
  * `getByLemmaBulk` chooses the highest reported status for a lemma between `Anki word` and `Wanikani`.
* The statistics also displays the current media tokens in your upcoming review similar to Anki
  * This is controlled by a button to switch between `Anki` and `WaniKani` similar to the tracks
  * If only one of them has data it will be defaulted to, otherwise it's `Anki`
* Now recording the time that build state messages reaches the UI so that ETA and duration calculations don't drift after completion

Assisted-by: Copilot:GPT-5.5


common/dictionary-db/dictionary-db-anki.ts
```diff
@@ -0,0 +1,962 @@
...
+        suspended: boolean;
+    }
+>;
+
+type AnkiTrackStatesForDB = Map<number, TrackStateForDB>;
+
+export async function buildAnkiCachePipeline(
+    db: _DictionaryDatabase,
+    profile: string,
+    settings: AsbplayerSettings,
```
killergerbah, line 70: Would it be possible to constrain this type to only what is required? It feels a bit scary to send the whole blob, most of it is unnecessary and is also potentially going across app/extension boundaries which means it would be a little more difficult to change this interface later if we wanted to. Not that it will practically make a huge difference, but as a matter of principle.

ShanaryS: I removed it completely, we can just pass `SettingsProvider` to Local and Extension DB class. There should be no issues with App/Extension version differences since the DB class itself is now creating the settings and would just ignore any message with it. Since it would communicate through dictionary handler, the order wouldn't matter since we are just now ignoring they settings object on the message.

For statistics though I wanted to the same but it's not possible since here the extension can communicate with the app, rather than the app passing in settings only. So older App instances (which shouldn't be common) would be expecting settings in the snapshot but it wouldn't be present causing it to not work. But for this I just stripped the settings down to dictionaryTracks since that's all that's needed.

Edit:
Actually I don't think we can just remove it. Cause outdated extensions will expect it. So I'll just strip it only in both places to keep it simple.

ShanaryS: I can actually keep most of the benefits going forward. But the App still needs to send the settings in case it's communicating with an old extension. So realistically settings won't be used going forward. This simplifies a couple functions since we know it will always be available now.

I opted not to do the same for statistics since publishing that can hit many different consumers so it basically impossible to remove without breaking extension v1.18.0+ but App < v1.18.0. I believe you said previously that's not a big concern, only people hosting the website locally could be affected if they aren't also building the extension.

killergerbah: Yeah for the app we should just assume that latest is always deployed. Ppl running it locally will need to pull.

common/wanikani/wanikani.ts
```diff
@@ -0,0 +1,230 @@
...
+            }
+            return response.json();
+        }
+    }
+
+    private async _waitForRateLimitReset(response: Response) {
+        const resetHeader = response.headers.get('RateLimit-Reset');
+        const resetValue = resetHeader === null ? Number.NaN : Number(resetHeader);
+        const delay = Number.isFinite(resetValue)
+            ? Math.max(1000, resetValue * 1000 - Date.now() + 1000)
```
killergerbah, line 226: Should `Math.max(1000, resetValue * 1000 - Date.now() + 1000)` be `Math.max(DEFAULT_RATE_LIMIT_RETRY_MS, resetValue * 1000 - Date.now() + 1000)` instead?

common/dictionary-db/dictionary-db.ts
```diff
@@ -67,6 +75,10 @@ export interface DictionaryTokenRecord {
     lemmas: string[];
     states: TokenState[];
     cardIds: number[];
+    subjectIds?: number[]; // Guaranteed to exist by db, only here to enforce type safety for the App communicating with an outdated extension
```
killergerbah: The token record can only have one source, so can `cardIds` serve the purpose of `subjectIds` instead? It would save you a column

common/dictionary-db/dictionary-db-anki.ts
```diff
@@ -0,0 +1,962 @@
...
+    TokenStatus,
+} from '@project/common/settings';
+import { HAS_LETTER_REGEX, inBatches, mapAsync } from '@project/common/util';
+import { Yomitan } from '@project/common/yomitan/yomitan';
+import { v4 as uuidv4 } from 'uuid';
+import {
+    _DictionaryDatabase,
+    BUILD_MIN_EXPIRATION_MS,
+    _buildIdHealthCheck,
+    _clearBuildIds,
```
killergerbah, line 27: Would it make sense to put these underscore-prefixed functions in their own module e.g. `dictionary-db/private`? That would obviate the need for an underscore and possibly increase readability

common/wanikani/wanikani.ts
ShanaryS: The idea is to always wait at least 1s even if it says to retry <1s. I'm not really concerned about WaniKani having a time above 60s, we would never hit the rate limit anyways.

common/dictionary-db/dictionary-db.ts
ShanaryS: I'm not sure, I would if I could rename it which is fine from a migration point. But the older messages expect cardIds so the db would be named differently. If we were to add 5 more sources, it would be annoying to have a column for each of them when the rest are empty, so I think it's probably best to just use cardIds.

common/dictionary-db/dictionary-db-anki.ts
ShanaryS: I wanted to minimize the amount of files the logic is spread across so that it's easier to follow and reason about.

common/components/DictionarySettingsTab.tsx
```diff
@@ -1254,6 +1570,79 @@ const DictionarySettingsTab: React.FC<Props> = ({
...
+                        <SettingsSection>{t('settings.dictionaryWaniKaniSection')}</SettingsSection>
+                        {waniKaniUserInfo && (
+                            <Typography variant="body2" color="textSecondary">
+                                {`${waniKaniUserInfo.data.username}: ${waniKaniUserInfo.data.level}/${
+                                    waniKaniUserInfo.data.subscription.max_level_granted ?? '?'
+                                }`}
+                            </Typography>
+                        )}
+                        <SettingsTextField
+                            type={showDictionaryWaniKaniApiToken ? 'text' : 'password'}
```
killergerbah: Just now I got a dialog like the following:

Image

Is there a way to prevent something like this from appearing?

common/dictionary-db/dictionary-db.ts
killergerbah: Just to make sure we're on the same page - I was suggesting that we don't add a column `subjectIds` and just use `cardIds` for both Anki and Wanikani. I think that would prevent this situation:

> If we were to add 5 more sources, it would be annoying to have a column for each of them when the rest are empty

common/components/DictionarySettingsTab.tsx
ShanaryS: It's probably the password field on the WaniKani token. I wanted to keep it hidden by default.

killergerbah: Can CSS be used instead of the password input type? As this dialog is just confusing and I don't think there's a benefit to the browser saving the API key, since asbplayer is already saving it.

ShanaryS: Yeah the pop-up is unintentional.

common/dictionary-db/dictionary-db.ts
ShanaryS: Now we are only using `cardIds` in the token store. Most likely future external sources would mimic the same design but they also could need a separate column if they need to track two things. Or maybe would would use a the primary key from another store we create as a lookup for all the necessary ids we need to match the token with the external source.

You'll need to delete your db and rebuild to fix the cache.

ShanaryS:
BTW the commit [group stats by lemma if enabled by settings](https://github.com/asbplayer/asbplayer/pull/997/commits/8136c66640a25a7b97633d96b9be74a4acca60c7) is a feature/bugfix that's outside this PR. Previously stats would only consider the surface form since using lemmas to count the stats would cause the sentence filtered view to have the wrong status annotation. This fixes it by allowing the use of lemma for only for certain stats while always using the surface form for stats that will be rendered as richText.

This should mean the stats now group by lemma when the `ANY_FORM_COLLECTED` or `LEMMA_FORM_COLLECTED` is being used. It's also valid to only group by lemma or only by surface but I think I'll tie it to the strategy selected for now.

common/dictionary-db/dictionary-db.ts
killergerbah: Makes sense, yeah we can add more columns as necessary

killergerbah (approved):
This has been working really well with my old WK account. Stuff that's hard to change later like DB schema and extension/app API looks good to me. Nothing else jumps out.

PR #1004, opened 2026-05-16
fix: search jimaku dramas entries alongside anime entries

### LINKED ISSUE #1010: Make online subtitle source support grabbing subs from both the Anime and Live Action sections of jimaku.cc
As of now, the "Search online subtitles" feature for grabbing subs from jimaku.cc will only return anime entries. This is because there is a [boolean](https://jimaku.cc/api/docs#tag/entries/GET/api/entries/search) (anime) you need to set when calling the jimaku API which will determine if you either get search results from the anime or live action section. Ideally, you should be able to search only for anime or live action, since adaptations will likely have identical titles in either section.

Dr-TNineS:
Users reported this on QQ: searching live-action/drama subtitles (e.g. 逃げるは恥だが役に立つ) returned no results 
because subtitles under https://jimaku.cc/dramas were not queried.

Online subtitle search only queried the default anime category on jimaku.cc, missing subtitles listed under the dramas/live-action section. This changes searchEntries to accept an optional flags parameter and updates the search UI to fetch both anime and drama entries in parallel, deduplicating by entry id.

Assisted-by: Codex:GPT-5.3-Codex

### DIFF
--- a/extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
+++ b/extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
@@ -130,9 +130,23 @@ export default function OnlineSubtitleSourceDialog({
 
         try {
             const client = new JimakuClient({ apiKey: jimakuApiKey });
-            const entries = (await client.searchEntries(query)).data;
+            const [animeResult, dramaResult] = await Promise.all([
+                client.searchEntries(query),
+                client.searchEntries(query, false),
+            ]);
+
+            // Merge and deduplicate by entry id
+            const seen = new Set<number>();
+            const mergedEntries = [...animeResult.data, ...dramaResult.data].filter((entry) => {
+                if (seen.has(entry.id)) {
+                    return false;
+                }
+                seen.add(entry.id);
+                return true;
+            });
+
             setLastQuery(query);
-            setJimakuEntries(entries.map((entry) => ({ id: entry.id, name: entry.name })));
+            setJimakuEntries(mergedEntries.map((entry) => ({ id: entry.id, name: entry.name })));
             setJimakuSelectedEntry(undefined);
             setJimakuFiles(undefined);
         } catch (e) {
--- a/extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
+++ b/extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
@@ -16,7 +16,7 @@ import ToggleButtonGroup from '@mui/material/ToggleButtonGroup';
 import Stack from '@mui/material/Stack';
 import TextField from '@mui/material/TextField';
 import Typography from '@mui/material/Typography';
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { JimakuClient, JimakuEntry } from '@/services/subtitle-sources';
 import IconButton from '@mui/material/IconButton';
@@ -103,6 +103,7 @@ export default function OnlineSubtitleSourceDialog({
     const [jimakuEntries, setJimakuEntries] = useState<{ id: number; name: string }[]>([]);
     const [jimakuSelectedEntry, setJimakuSelectedEntry] = useState<JimakuEntry>();
     const [jimakuFiles, setJimakuFiles] = useState<OnlineSubtitleImportCandidate[]>();
+    const resultsCache = useRef<Map<string, { anime: JimakuEntry[]; drama: JimakuEntry[] }>>(new Map());
 
     const normalizedDetectedTitleHint = useMemo(
         () => normalizeDetectedTitleHint(detectedTitleHint),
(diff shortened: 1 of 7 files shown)


xwxb:
Thanks for the PR! I didn't realize when I first implemented this that Jimaku.cc defaults to searching only anime (`anime=true`).

However, did you verify the behavior of the `flags` parameter? I just did some testing with the API, and it seems that `&flags=2` doesn't actually filter for dramas or return any different results. According to the documentation and my tests, the correct way to search for live-action/dramas is to explicitly set `anime=false` in the query parameters.
https://jimaku.cc/api/docs#tag/entries/GET/api/entries/search

For example, searching for '逃げるは恥だが役に立つ' with `anime=false` returns the expected results, but searching with `flags=2` (or the default) returns an empty list.

Dr-TNineS:
Thanks for catching that! You're right — the correct query parameter is anime=false, not flags=2. I've updated searchEntries to accept an anime?: boolean parameter and the UI now searches with anime=false for dramas. Tests and compile pass. PTAL!

xwxb:
Looks good! I've double-checked the API behavior, and it makes sense now. Thanks for the fix!

Dr-TNineS:
Thanks for the review and for catching the anime vs flags issue! Glad it's sorted out. 🎉

xwxb:
That said, for the long term, it might be better to let users select the search category to avoid unnecessary API requests and be more respectful of the upstream provider. We can treat this as a temporary fix for now, and I might look into adding a toggle for it later.

Dr-TNineS:
Good point, a category toggle would definitely be more API-friendly long-term. Happy to help if you decide to implement that later. For now I'm just glad we got the dramas working! 🎉

Dr-TNineS:
Added a category toggle (Anime / Drama) so users can limit searches to the relevant category and reduce unnecessary API calls. Previously `searchEntries` always queried both anime and drama in parallel, which doubled requests even when the user only needed one.

While working on this I also fixed a pre-existing issue: the search button becomes disabled after the first search because `lastQuery === query` prevents re-searching the same term. This meant if you accidentally searched with the wrong category, switching to the correct one still wouldn't return any results — neither clicking the button nor pressing Enter would trigger a new search. Users might wrongly assume the subtitle they're looking for simply isn't in the database.

Fixed by tracking `lastSearchCategory` alongside `lastQuery`, so the button only disables when both the query and category are unchanged. Also added `jimakuSearchCategory` to the `useCallback` dependency array to avoid stale closures.

I discussed this with xwxb, and he'd like to hear your thoughts as well — @killergerbah, what's your take?

extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
```diff
@@ -130,17 +137,35 @@ export default function OnlineSubtitleSourceDialog({
 
         try {
             const client = new JimakuClient({ apiKey: jimakuApiKey });
-            const entries = (await client.searchEntries(query)).data;
+            const requests: Promise<{ data: { id: number; name: string }[] }>[] = [];
```
killergerbah: Looks like `requests` no longer needs to be an array

killergerbah:
> I discussed this with xwxb, and he'd like to hear your thoughts as well — @killergerbah, what's your take?

I agree we should not double the requests if we can help it. Users are not likely to be searching for both anime and drama at the same time. Would just want to make sure that the UI/UX is reasonably efficient. I haven't tested your PR but may make some comments when I do.

Dr-TNineS:
Thanks for taking a look! Based on your feedback and discussion with xwxb, I've made a few changes:

- Removed the "Both" option — the toggle now only has **Anime** and **Drama**, with Anime as the default. Only one API request per search, no more doubling.
- Simplified `handleSearchJimaku` — removed the array / `Promise.all` / `flatMap` / dedup logic, replaced with a plain ternary since there's only ever one request now.
- Fixed the re-search debounce — tracked `lastSearchCategory` so switching categories with the same query triggers a new search instead of silently doing nothing.

Let me know if you spot anything else when you test it.

extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
killergerbah, line 146: This should trigger a query automatically if the user switches categories. Also, query results should be cached until the search term is changed.

extension/src/ui/components/OnlineSubtitleSourceDialog.tsx
```diff
@@ -235,6 +247,20 @@ export default function OnlineSubtitleSourceDialog({
                                 },
                             }}
                         />
+                    <ToggleButtonGroup
```
killergerbah: This looks really out of place

image

killergerbah (commented):
Some minor comments. Feel free to give them a try if not I will solve them post merge.

Dr-TNineS:
Thanks for the review! I gave both comments a try.

Category switching now auto-searches when applicable, and results are cached until the search term changes. I also adjusted the toggle button group styling so it aligns with the search field.

Let me know if you’d prefer any changes to the behavior or layout.

killergerbah (approved):
Great work, thanks
I'll probably adjust the UI further post-merge

PR #1003, opened 2026-05-14
Add note for frequency-based dictionaries

Roka20012:
- [x] I have read the [contribution guidelines](https://github.com/asbplayer/asbplayer/blob/main/CONTRIBUTING.md).

Hello, as a newbie to the ASB player, it's not obvious how to make the frequency feature work correctly.

I had a problem with configuring the frequency number for subtitles, and with the help of the Discord community, I managed to fix it 

So I added a note about Yomitan frequency-based dictionaries and configuration.

Let me know if it's a needed note or no?
Thanks for the great extension!

PR #1005, opened 2026-05-16
Infer if rank-based to allow undeclared frequency dictionaries

ShanaryS:
Related to: #1003

Some frequency dictionaries in Yomitan may not have the `frequencyMode` which explicitly says if it's `rank-based` or `occurrence-based`. In these cases, we will now infer the type by leveraging the fact that common words are common. The test is simple, take the 10 most and least occurrent words in the current subtitle then compare these two groups against each other to see if the most occurrent words have the lowest frequency. To be considered `rank-based`, at least 7 of these comparisons need to win. Here is an example of a dictionary that would be previously ignored:

```json
{"mostOccurringWords": [
	{"token":"の","frequency":1,"occurrences":169},
	{"token":"に","frequency":3,"occurrences":122},
	{"token":"は","frequency":7,"occurrences":105},
	{"token":"が","frequency":4,"occurrences":101},
	{"token":"を","frequency":5,"occurrences":89},
	{"token":"だ","frequency":6,"occurrences":81},
	{"token":"で","frequency":12,"occurrences":73},
	{"token":"な","frequency":89,"occurrences":69},
	{"token":"も","frequency":13,"occurrences":64},
	{"token":"と","frequency":10,"occurrences":61}
],
"leastOccurringWords": [
	{"token":"眈々","frequency":62783,"occurrences":1},
	{"token":"チンピ","frequency":66671,"occurrences":1},
	{"token":"血生臭い","frequency":78451,"occurrences":1},
	{"token":"たぶん","frequency":81587,"occurrences":1},
	{"token":"澱んだ","frequency":85568,"occurrences":1},
	{"token":"呼び立てて","frequency":89612,"occurrences":1},
	{"token":"喰らいついて","frequency":91401,"occurrences":1},
	{"token":"弩","frequency":100182,"occurrences":1},
	{"token":"うえ","frequency":102895,"occurrences":1},
	{"token":"かかり","frequency":140158,"occurrences":1}
]}
```

Due to [Zipf's law](https://en.wikipedia.org/wiki/Zipf%27s_law), we can expect this to hold for pretty much any text that's natural language. This is even accurate with only 10 subtitle events with 50 unique words total. So this is a very strong signal and extremely low risk for full subtitles. The main shortcomings would be if a name or a place was used frequently but it's still unlikely to beat out the top words. At least we shouldn't expect 4 or more of the top 10 most occurrent words to not be of the most frequent in the dictionaries.

If we do need to make it stronger, we can increase the number of wins needed up from 7 or even change the group size for comparison to a higher number. There is a tradeoff though as we move into median words we could start seeing more random wins/losses due to the content word variance.

### DIFF
--- a/common/yomitan/yomitan.ts
+++ b/common/yomitan/yomitan.ts
@@ -6,6 +6,8 @@ import { coerce, lt, gte } from 'semver';
 const TOKENIZE_BATCH_SIZE = 100; // 1k can cause 1.5GB memory on Yomitan for subtitles, Anki cards may be larger too
 const TERM_ENTRIES_BATCH_SIZE = 10; // 100 is only 10% faster (17s vs 19s for a 23min subtitle)
 const TERM_ENTRIES_DEBOUNCE_MS = 10; // Prevents using too much resources
+const FREQUENCY_MODE_INFERENCE_GROUP_SIZE = 10;
+const FREQUENCY_MODE_INFERENCE_RANK_BASED_MATCHES = 7;
 
 const YEAR_MONTH_REGEX = /(?<year>20\d{2})(?<month>[01]\d)/;
 
@@ -38,6 +40,7 @@ interface TermSource {
     isPrimary: boolean;
 }
 
+type FrequencyMode = 'rank-based' | 'occurrence-based';
 interface TermFrequency {
     index: number;
     headwordIndex: number;
@@ -45,7 +48,7 @@ interface TermFrequency {
     dictionaryIndex: number;
     dictionaryAlias: string;
     hasReading: boolean;
-    frequencyMode?: 'occurrence-based' | 'rank-based' | null;
+    frequencyMode?: FrequencyMode | null;
     frequency: number;
     displayValue: string | null;
     displayValueParsed: boolean;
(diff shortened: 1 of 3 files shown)


common/yomitan/yomitan.ts
```diff
@@ -541,6 +550,76 @@ export class Yomitan {
         }
     }
 
+    private frequencyMode(token: string, f: TermFrequency): FrequencyMode | undefined {
+        if (f.frequencyMode) return f.frequencyMode;
+
+        let frequencies = this.frequencyModeInferenceData.get(f.dictionary);
```
killergerbah, line 556: Nitpick - the method name does not feel descriptive of what it's doing, which is getting the frequency mode and also populating `this.frequencyModeInferenceData`

The method could be renamed or the behavior split into different methods

killergerbah (approved):

killergerbah:
Do you think the algorithm will still be reasonably accurate for texts with really common words only? Either way feel free to merge when ready

ShanaryS:
There are definitely ways this can be tripped up but it would need to be artificial and short. I don't think it's possible outside of maybe media with very little dialog (<10 sentences) or subtitles for YouTube videos that aren't natural language. But these probably won't be used for mining at least not frequently.

As well what we use frequency for is purely visual though it may be used for automation in the future. In those cases I don't think people would trigger the automation for things that would trip it up. I also think most frequency dictionaries are rank-based, so it may be unlikely to really matter.

common/yomitan/yomitan.ts
ShanaryS: I renamed it to `resolveFrequencyMode`. I prefer a single function as it needs to do both jobs around the same time.

PR #1008, opened 2026-05-17
option to match words across language scripts

ShanaryS:
Multiple lemmas are used to allow for matching through different scripts, only Kanji/Kana is explicitly supported but this could apply to more languages in the future.

Currently `ANY_FOR_COLLECTED` and `LEMMA_FORM_COLLECTED` will match Kanji/Kana subtitles with Kanji/Kana collection. A new option `dictionaryMatchAcrossScripts` when set to `false` will only allow `Kanji <-> Kanji` and `Kana <-> Kana` matches. This is a very strict mode which will require the user to collect the Kana form even when they already have the Kanji.

This new option will default to `true` which matches the behavior of the current logic implicitly but now we have one change. We will no longer match Kanji in the subtitle through Kana lemmas or collection. This will prevent Kanji homophones being marked when the user only has one of them collected. We will still match Kana subtitles if you only have the Kanji collected as this is a common case for Japanese. In summary:

```
if dictionaryMatchAcrossScripts:
- Kana subtitles can match kanji in collection, could be homophones but text processing can't handle it so we allow it.
- Kanji subtitles only match with kanji in collection, prevents kana collected matches all kanji homophones.

if not dictionaryMatchAcrossScripts:
- Never match across scripts, downside is if kanji is collected kana will need to be collected too.
- Essentially a strict mode where the user needs to collect all script forms of a word.
```


common/settings/settings.ts
```diff
@@ -227,6 +227,7 @@ export interface DictionaryTrack {
     readonly dictionaryColorizeOnHoverOnly: boolean; // Currently applies to both colorization and reading annotations, named in case we want to separate later
     readonly dictionaryHighlightOnHover: boolean;
     readonly dictionaryTokenMatchStrategy: TokenMatchStrategy;
+    readonly dictionaryMatchAcrossScripts: boolean;
     readonly dictionaryTokenMatchStrategyPriority: TokenMatchStrategyPriority;
     readonly dictionaryYomitanUrl: string;
```
killergerbah, line 232: Should the new setting trigger a DB rebuild?

killergerbah (commented):
Do you think ultimately the user should be able to control which lemmas they know in the word DB? I can see this making it complicated to rebuild the DB since user edits would be mixed in.

common/settings/settings.ts
ShanaryS: No, it already parses kana along with kanji when storing lemmas. Right now lemmas being an array and this option only applies to Japanese. This new option is only a restriction in the matching from before, even when set to true.

If we were to explicitly support multiple scripts for other languages, then we would need to rebuild. But even in that case it would be very easy to do by just adding a new field onto the `meta` table like `version` where we reparse with Yomitan during playback (when we know it has to be available).

ShanaryS:
> Do you think ultimately the user should be able to control which lemmas they know in the word DB?

The issue is that it needs to be consistent. So if Yomitan doesn't parse it the same way it might break the strategies that relies on lemma. Even if there are cases where the current capabilities don't suffice, I rather accept the edge cases than potentially open up to such problems. So far I haven't really seen any complaints so I have to assume the parsing is acceptable to most users.

If such a need arises, I prefer to write code that handles it and trigger a recalculation on all the tokens in the db. I'm honestly surprised we haven't needed the user to select a target language by now to handle more edge cases.

> I can see this making it complicated to rebuild the DB since user edits would be mixed in.

Yeah that's another can of worms. Right now Anki and WaniKani are easy since worst case we can just wipe the db and reprocesses.

common/locales/de.json
```diff
@@ -492,6 +492,7 @@
         "dictionaryTokenMatchStrategyLemmaOrExactFormCollected": "Lemma or exact form collected",
         "dictionaryTokenMatchStrategyLemmaFormCollected": "Lemma form collected",
         "dictionaryTokenMatchStrategyExactFormCollected": "Exact form collected",
+        "dictionaryMatchAcrossScripts": "Match across language scripts",
```
killergerbah, line 495: I wonder if there's a way to make this setting's description more comprehensible, without making any references to Japanese

killergerbah (approved):
The logic loosely makes sense to me, I don't see any obvious errors
Feel free to merge when ready

killergerbah:
> The issue is that it needs to be consistent. So if Yomitan doesn't parse it the same way it might break the strategies that relies on lemma. Even if there are cases where the current capabilities don't suffice, I rather accept the edge cases than potentially open up to such problems. So far I haven't really seen any complaints so I have to assume the parsing is acceptable to most users.

Maybe the fact that no one complains means that people are generally fine with the lemmatization being approximate. Which make sense, no one should be relying 100% on the tool to understand their own known vocabulary.

common/locales/de.json
ShanaryS: I think it's very difficult. I can't think of any normal terms that would get it across without using examples. I think the current description makes it sound like an extra feature so people will probably keep it enabled. Disabling it is valid too so it's not a big deal if they do.

PR #1011, opened 2026-05-18
Feature/skip re-recording media when mining same subtitle

Roka20012:
Should cache the last recorded subtitle sentence

- Add MediaCache to reuse audio and screenshot from a previous mine when the same subtitle is mined again. 
- Add a "Clear cached media" toolbar button in the Anki dialog that clears the cache and immediately re-records the full subtitle from scratch.

Should fix
https://github.com/asbplayer/asbplayer/issues/914
https://github.com/asbplayer/asbplayer/issues/842

Demo:

Video is low quality (GitHub restrictions for file size so converted to gif)

But you can see here

1. Media for a subtitle is cached
2. When using re-recording media with custom timeline is also cached
3. When pressing clear cached media, the whole subtitle is re-recorded with initial timelines(not adjusted manually)

I think maybe it's better to move the clear cached media button somewhere else, and change the icon, but so far it's just a POC, open to suggestions

ezgif-3359b6390a6a90e3

I am new for asbplayer code, so this feature is 
Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/extension/src/services/media-cache.ts
+++ b/extension/src/services/media-cache.ts
@@ -1,0 +1,24 @@
+import { AudioModel, ImageModel } from '@project/common';
+
+export interface CachedMediaRecord {
+    subtitleStart: number;
+    subtitleEnd: number;
+    audioModel?: AudioModel;
+    imageModel?: ImageModel;
+}
+
+export default class MediaCache {
+    private readonly _byTab = new Map<string, CachedMediaRecord>();
+
+    key(tabId: number, src: string) {
+        return `${tabId}:${src}`;
+    }
+
+    get(key: string): CachedMediaRecord | undefined {
+        return this._byTab.get(key);
+    }
+
+    set(key: string, record: CachedMediaRecord) {
+        this._byTab.set(key, record);
+    }
+}
--- a/extension/src/ui/components/AnkiUi.tsx
+++ b/extension/src/ui/components/AnkiUi.tsx
@@ -234,6 +234,27 @@ export default function AnkiUi({ bridge }: Props) {
         bridge.sendMessageFromServer(message);
     }, [bridge, savedState]);
 
+    const handleClearMediaCache = useCallback(() => {
+        setOpen(false);
+        const state = savedState();
+        const fullInterval = [state.subtitle.start, state.subtitle.end];
+        const resetState = {
+            ...state,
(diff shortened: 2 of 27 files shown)


extension/src/services/media-cache.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface CachedMediaRecord {
+    subtitleStart: number;
+    subtitleEnd: number;
+    audioModel?: AudioModel;
+    imageModel?: ImageModel;
+}
+
+export default class MediaCache {
+    private readonly _byTab = new Map<string, CachedMediaRecord>();
```
killergerbah, line 11: The service workers running the background script can go away, when that happens this cache will be lost. I wonder if IndexedDB can be used instead.

extension/src/services/media-cache.ts
```diff
@@ -0,0 +1,24 @@
+import { AudioModel, ImageModel } from '@project/common';
+
+export interface CachedMediaRecord {
+    subtitleStart: number;
```
killergerbah, line 4: I think the subtitle text should be included as well, to make the cache resistant to similar timestamps existing across different subtitle files. Which is not likely, but it is also not much effort to include the text here and in the comparison that determines whether to use the cached audio.

extension/src/services/media-cache.ts
```diff
@@ -0,0 +1,24 @@
...
+    subtitleEnd: number;
+    audioModel?: AudioModel;
+    imageModel?: ImageModel;
```
killergerbah, line 7: What is the value of caching the image as well?

extension/src/ui/components/AnkiUi.tsx
```diff
@@ -234,6 +234,27 @@ export default function AnkiUi({ bridge }: Props) {
         bridge.sendMessageFromServer(message);
     }, [bridge, savedState]);
 
+    const handleClearMediaCache = useCallback(() => {
```
killergerbah, line 237: I would be hesitant to include this in the Anki UI unless absolutely necessary. Most users won't care for implementation details such as the cache being used here, so putting it front and center doesn't feel like a good use of UI space.
Also, why would there be any need to clear the cache if the cached media is accurate?

killergerbah (commented):
Thanks, this is a highly requested and valuable feature

PR #1012, opened 2026-05-19
project Anki and WaniKani stats on past/future material

ShanaryS:
_Anyone currently using the `WaniKani` features will need to [rebuild their Anki DB](https://docs.asbplayer.dev/docs/guides/annotation#clear-anki-word-database). The previous migration was updated to force a rebuild which will be fine for all users except those who used the unreleased WaniKani asbplayer._

---

The projected section can now project knowledge for `Anki` and `WaniKani`. This allows the user to select the number of `new` cards and their WaniKani level that asbplayer will use to project their stats for. This allows a user to see what their stats would be after completing their next `X new cards` (sorted by the order they will be introducted) or reaching a WaniKani level.

The projected rewatch now also promotes each status by one level in addition to collecting the i+1 sentences. Even though this feature makes a lot of assumptions, I think this one makes sense to do as well.

Optimized the runtime calls needed for statistics by storing more info in the db. The main api calls are `/user` to WaniKani so we can get the max level the user has and for the anki due dates since it's impossible to determine through `AnkiConnect`.

Finally some changes were made to how we group by lemmas for stats. If the user is using a strategy that groups by lemma, all stats will now follow suit except for the sentence filters. Those need to use the surface form otherwise the token styling will disagree with the filter.

Assisted-by: Copilot:GPT-5.5

### DIFF
--- a/common/components/Statistics.tsx
+++ b/common/components/Statistics.tsx
@@ -17,6 +17,9 @@ import {
     percent,
     percentDisplay,
     DictionaryStatisticsSnapshot,
+    DictionaryStatisticsRewatchProjection,
+    DictionaryStatisticsRewatchProjectionsByTrack,
+    DictionaryStatisticsRewatchSnapshot,
     processDictionaryStatisticsAnkiTrackSnapshot,
     processDictionaryStatisticsSnapshot,
     processDictionaryStatisticsWaniKaniTrackSnapshot,
@@ -38,6 +41,7 @@ import IconButton from '@mui/material/IconButton';
 import LinearProgress from '@mui/material/LinearProgress';
 import MenuItem from '@mui/material/MenuItem';
 import Paper from '@mui/material/Paper';
+import Switch from '@mui/material/Switch';
 import TextField from '@mui/material/TextField';
 import Typography, { type TypographyProps } from '@mui/material/Typography';
 import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';
@@ -50,6 +54,8 @@ import BarChartIcon from '@mui/icons-material/BarChart';
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import PictureInPictureAltIcon from '@mui/icons-material/PictureInPictureAlt';
 import Link from '@mui/material/Link';
+import { WaniKani, type WaniKaniUser } from '@project/common/wanikani';
+import SwitchLabelWithHoverEffect from './SwitchLabelWithHoverEffect';
 
 export interface MediaInfo {
     sourceString: string;
@@ -467,6 +473,8 @@ interface SentenceStatsPanelProps {
     globalKnownCount: number;
     onOpenSentenceBucketDetails: (bucket: DictionaryStatisticsSentenceDialogBucket) => void;
     headerAction?: ReactNode;
+    extraStats?: ReactNode;
+    hideStats?: boolean;
     sentenceFiltersPosition: 'top' | 'bottom';
     emptyMessage?: string;
 }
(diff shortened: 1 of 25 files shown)


common/components/Statistics.tsx
```diff
@@ -527,8 +538,254 @@ function StatBoxes({ keyPrefix, stats }: StatBoxesProps) {
...
+    const [userInfo, setUserInfo] = useState<WaniKaniUser>();
+    const [error, setError] = useState<string>();
+
+    useEffect(() => {
+        const trimmedApiToken = apiToken.trim();
+        setUserInfo(undefined);
+        setError(undefined);
+        if (!trimmedApiToken) return;
+
+        let canceled = false;
```
killergerbah: I looked into this and looks like this pattern isn't necessary: https://github.com/reactwg/react-18/discussions/82

So `canceled` can be removed, it makes the code more readable as well

common/components/Statistics.tsx
```diff
@@ -527,8 +538,254 @@ function StatBoxes({ keyPrefix, stats }: StatBoxesProps) {
...
+                            }
+                        />
+                    }
+                    label={t('statistics.anki.projectReviewedCards')}
+                    labelPlacement="start"
+                    sx={{ width: '100%', mt: -0.25 }}
+                />
+            )}
+            {(ankiAvailable || waniKaniAvailable) && (
+                <Typography variant="caption" color="text.secondary">
```
killergerbah: Is this supposed to be a form control?

common/components/Statistics.tsx
```diff
@@ -630,45 +927,22 @@ function SentenceStatsPanel({
         >
             <Box sx={{ display: 'flex', flexDirection: 'column', gap: 1.5, height: '100%' }}>
                 <Box sx={{ display: 'flex', justifyContent: 'space-between', gap: 1, alignItems: 'flex-start' }}>
-                    <StatisticsSectionSubHeading>{title}</StatisticsSectionSubHeading>
+                    <Box sx={{ display: 'inline-flex', alignItems: 'center', gap: 0.5 }}>
+                        <StatisticsSectionSubHeading>{title}</StatisticsSectionSubHeading>
+                        {infoLines !== undefined && <StatisticsInfoTooltip label={title} lines={infoLines} />}
```
killergerbah, line 1025: This looks misaligned on my computer:
image

common/components/Statistics.tsx
```diff
@@ -527,8 +538,254 @@ function StatBoxes({ keyPrefix, stats }: StatBoxesProps) {
...
+                    <Typography variant="subtitle2" sx={{ mb: 0.5 }}>
+                        {t('statistics.waniKani.projectedWords')}
+                    </Typography>
+                    <StatRow
+                        label={
+                            snapshot.waniKaniStats.level === undefined
+                                ? t('statistics.waniKani.currentKnowledge')
+                                : t('statistics.waniKani.projectedLevel', { level: snapshot.waniKaniStats.level })
+                        }
+                        value={`${snapshot.waniKaniStats.knownWords} / ${snapshot.waniKaniStats.totalWords}`}
```
killergerbah, line 616: I have a really old WK deck at level 60 (haven't reviewed for years) and every level shows up as 1/1, is that expected?

ShanaryS: Have you rebuilt the WaniKani cache since the last PR? If not trying clearing the token, build, then rebuild with the token.

killergerbah: Oh that worked, thanks
I should learn to read

killergerbah (commented):
Looks good, nothing jumps out to me.
Before merging I would like to see if I can make the UI more intuitive. When I first tried using the feature, it wasn't totally obvious to me how to use it (possibly because my brain is small) / what the new projection inputs meant, but I think that can be improved.

- Maybe introduce Anki/WK logos to make it obvious what is Anki-related and what is WaniKani-related
- See if the wording can be updated to make it obvious that the Anki input refers to # new cards reviewed etc.
- Maybe rearrange things so that Anki and WK info is closer together. Currently both Anki and WK stats are spread out throughout the UI, including the section at the bottom.

Also, just a general thought and I'm not sure if there's a good answer. Does it make sense for all 3 projection inputs to be independently adjustable? Rewatching the same media 10 times vs getting to WK level 60 vs reviewing the next 10 Anki cards seem like distinct timelines.

Regarding Anki specifically, does it make sense to track projected stats per-deck? Wondering if the stats lose much value if we aggregated them under a single "X new cards reviewed" input. Just looking for ways to simplify the UI if possible.

common/components/Statistics.tsx
ShanaryS: I don't think so, it's just for displaying the title for the section of inputs for the Anki/WaniKani projection. But I'm not sure I fully understand how FormControl affects everything. If you think it should it probably should.

ShanaryS:
> Also, just a general thought and I'm not sure if there's a good answer. Does it make sense for all 3 projection inputs to be independently adjustable? Rewatching the same media 10 times vs getting to WK level 60 vs reviewing the next 10 Anki cards seem like distinct timelines.

I'd like them to be independently adjustable into the same stats, even though they don't really relate to each other. The main reason is that a lot of things change by the projection and it would take up too much space if they were separate. But I also think it's only a value add, most people would only use a single one at a time but if they want to use them all it doesn't hurt.

> Regarding Anki specifically, does it make sense to track projected stats per-deck? Wondering if the stats lose much value if we aggregated them under a single "X new cards reviewed" input. Just looking for ways to simplify the UI if possible.

I had it like that for the MVP and it's fine, but similar to above I though this gave more control. The thing is that if the user has a premade deck and a mining deck, they could have stopped the pre-made one when they started mining. But the new cards only completely exhaust the pre-made deck before it starts marking the mining as new since it was likely added first to Anki. So make it it per deck avoids that complication and is more useful as users can see how finishing either deck would affect their comprehension.

> Before merging I would like to see if I can make the UI more intuitive.

I'll think of ways to simplify the design.

common/components/Statistics.tsx
killergerbah: I wasn't clear, what I meant was due to that phrase appearing directly under "Treat all reviewed cards as mature" it looks like it also should have been a radio switch:

image

But now I also see that "Treat upcoming as mature" can be interpreted as treating all the upcoming X sentences and X levels as mature.

ShanaryS:
I changed the projected section to move the Anki and WaniKani options into popover buttons so it's clear what refers to what. I also updated `Treat upcoming as mature` to `Treat upcoming material as mature` for the deck new cards and level controls.

The projected stats are meant to be compared with the current watch so I want them close to each other. I considered putting the projected controls for Anki and WaniKani in their general statistics section but it felt too far from projected and confusing as to what it applies to (it doesn't change those stats). Especially if you were to have like 5 decks, I think using a popover is good so it doesn't eat up too much space.

I also considered combining the current watch stats and the projected stats into the same view with like a diff style change on only the numbers but that would be a large refactor. Maybe I'll do that in the future as we have the flexibility to change the UI without any concern to version mismatch.

---

Another note on all 3 stats affecting the same projection:
- Rewatches controls how promoting Uncollected (through i+1 sentences) improves knowledge
- Anki (treat new cards as mature) and WaniKani (level above current) shows what knowledge would be by promoting Unknown
- Anki (treat reviewed as mature) and WaniKani (level to current or below) shows what knowledge would be if all Learning+ was Mature

So these options while independent can come together by complimenting each other well.

killergerbah (approved):
It looks really good! It feels a lot more intuitive to use now, since it feels like the UI is directing the user as to what to do by starting out with just the form inputs only.

Feel free to merge when ready

killergerbah:
Btw regarding this

>Maybe I'll do that in the future as we have the flexibility to change the UI without any concern to version mismatch.

Don't we have that flexibility already? The latest UI will always receive stats from the latest database code

ShanaryS:
> Don't we have that flexibility already? The latest UI will always receive stats from the latest database code

Yeah since the stats use the side panel but that could change in the future. Then an old extension could be different from App statistics.

But I don't think we should be worried for stats anyways since it's purely for display. If I do a refactor in the future then I would probably remove a lot of the fallbacks and simplify the code.

PR #1014, opened 2026-05-21
Reduce Yomitan batch when native messaging limits are exceeded

ShanaryS:
The main goal is to reduce the batch size used when the responses from Yomitan are too large for the browsers native messaging. The `yomitan-api` supports Chrome, Firefox, and Edge. For which the limits are:

- Chrome:
  - Input: `1MB`
  - Output: `64MB`
- Firefox/Edge:
  - Input: `1MB`
  - Output: `4GB`

As you can see the limits will only ever hit for Chrome. Input is never a concern as we at most send an array of 100 tokens. The main idea behind the reducing strategy is that it's random what set of tokens can trigger a large response and that it's relatively infrequent. So we require a good deal of evidence to permanently reduce the batch size (for that Yomitan instance). This ensures we prefer speed when possible while maintaining reliability.

I also added better retry delays for Yomitan failures during annotations.

### DIFF
--- a/common/components/WordBrowserDialog.tsx
+++ b/common/components/WordBrowserDialog.tsx
@@ -754,7 +754,7 @@ const BulkUpdateDialog: React.FC<{
                             count: selectedRows.length,
                         })}
                     </Typography>
-                    {bulkStatus === TokenStatus.UNCOLLECTED && (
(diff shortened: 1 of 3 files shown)


killergerbah (approved):
I remember hitting this and adding the option for gzip compression to my local copy of the Yomitan API. This seemed to fix the issue as well. 

Feel free to merge when ready

PR #1015, opened 2026-05-21
Feature/download-audio - add ability to download recorded audio on device

Roka20012:
Have added the ability to download recorded audio on the device

Useful when need audio for mining sentences

image

Assisted with: Claude Sonnet 4.6

### DIFF
--- a/common/components/AudioField.tsx
+++ b/common/components/AudioField.tsx
@@ -7,6 +7,7 @@ import IconButton from '@mui/material/IconButton';
 import FiberManualRecordIcon from '@mui/icons-material/FiberManualRecord';
 import PlayArrowIcon from '@mui/icons-material/PlayArrow';
 import PauseIcon from '@mui/icons-material/Pause';
+import DownloadIcon from '@mui/icons-material/Download';
 import { AudioClip } from '../audio-clip';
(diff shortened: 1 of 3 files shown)


common/components/AudioField.tsx
```diff
@@ -97,6 +98,20 @@ export default function AudioField({
                     </span>
                 </Tooltip>
             )}
+            <Tooltip title={t('ankiDialog.downloadAudio')!}>
```
killergerbah: Nitpick, can this be moved to be the first button? If the button is the last, the recording button becomes the middle. The recording button is probably the most-used button and it would be harder to aim at if it's in the middle.

killergerbah: Can you use `action.downloadAudio` instead? It's a little costly to add new loc strings when suitable ones already exist. The strings remain untranslated until a human does it, and also makes the loc file (which is loaded by every UI dialog) a little bigger.

Roka20012: thx, for feedback fixed

Roka20012: done, thx

killergerbah (approved):

killergerbah:
Thanks @Roka20012

PR #1019, opened 2026-05-24
Fix sub detection on osnplus

killergerbah:
This feels like a somewhat risky change and so would like another pair of eyes on it

- Main bug fix was a regex change for osnplus manifest file. This is easy.
- The riskier change was providing a default src string for video elements that don't have them. osnplus and twitch are examples of sites that don't assign a src string to the video element. The main reason for doing this was to ensure that video elements without src strings can be identified and distinguished from other video elements, and to avoid any possible bugs that might be caused by code assuming src strings should be nonempty.

### DIFF
--- a/extension/src/services/binding.ts
+++ b/extension/src/services/binding.ts
@@ -127,6 +127,8 @@ const startAudioRecordingErrorResponse: (e: any) => StartRecordingResponse = (e:
 };
 
 export default class Binding {
+    private readonly _fallbackVideoSrc = uuidv4();
+
     subscribed: boolean = false;
 
     ankiUiSavedState?: AnkiUiSavedState;
@@ -213,7 +215,7 @@ export default class Binding {
 
     constructor(video: HTMLMediaElement, hasPageScript: boolean, frameId?: string) {
         this.video = video;
-        this._registeredVideoSrc = video.src || uuidv4();
+        this._registeredVideoSrc = video.src || this._fallbackVideoSrc;
         this.hasPageScript = hasPageScript;
         this.dictionary = new DictionaryProvider(new ExtensionDictionaryStorage());
         this.settings = new SettingsProvider(new ExtensionSettingsStorage());
@@ -676,9 +678,7 @@ export default class Binding {
 
         if (this.hasPageScript) {
             this.videoChangeListener = () => {
-                if (this.video.src) {
-                    this._updateRegisteredVideoSrc(this.video.src);
-                }
+                this._updateRegisteredVideoSrc(this.video.src || this._fallbackVideoSrc);
                 this.videoDataSyncController.requestSubtitles();
                 this._resetSubtitles();
             };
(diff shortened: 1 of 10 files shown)


extension/src/services/binding.ts
```diff
@@ -673,17 +676,19 @@ export default class Binding {
...
             this.video.addEventListener('loadedmetadata', this.videoChangeListener);
         }
 
         this.heartbeatInterval = setInterval(() => {
-            const src = this.video.src || undefined;
-            this._updateRegisteredVideoSrc(src);
-            if (!src) return;
+            if (this.video.src) {
+                this._updateRegisteredVideoSrc(this.video.src);
+            }
```
ShanaryS: By only updating if there is a `video.src`, won't changes to a video without a `src` not work properly? I would think that we would also be using `uuidv4()` somewhere else in this class to cover the case of a video change to one without a `src`.

Unless we are assuming that any website without a `src` will never have one and so will use the previous `uuidv4()` value. Even so we will never call `_notifyVideoDisappeared()` if the new video has no source and `unbind()` is never called. This seems fine as `video-disappeared` is only used to prune `VideoElements`.

I think all of this is fine we are a fine with the `src` not changing with a new video if the new one doesn't have it. Especially if we expect sites to either always provide it or never do.

extension/src/services/binding.ts
```diff
@@ -214,10 +213,11 @@ export default class Binding {
 
     constructor(video: HTMLMediaElement, hasPageScript: boolean, frameId?: string) {
         this.video = video;
+        this._registeredVideoSrc = video.src || uuidv4();
```
ShanaryS: Since the video src is now always defined, does this have any negative effects on the codebase checking for `src === undefined`? I assume it's fine since those just cover when there isn't a video at all but I'm not confident in that.

extension/src/services/binding.ts
killergerbah: Yeah my brain had completely excluded the possibility that could go from empty to nonempty to empty `src`. But technically it's possible. To accommodate this I'm now generating a single UUID that will always be used as the fallback in case `src` isn't defined. That way the fallback string is stable whenever this happens.

extension/src/services/binding.ts
killergerbah: Yeah should be fine, there has never been any really special treatment of videos with undefined src strings, besides excluding them from video detection by default

ShanaryS (approved):
I think it's good now, as long as we accept that we won't send disappeared messages until the binding instance itself is unbound. So changing videos with both using the fallback is undetectable (unless through other means).

killergerbah:
> as long as we accept that we won't send disappeared messages until the binding instance itself is unbound. So changing videos with both using the fallback is undetectable (unless through other means).

I think this is probably fine, it's _possible_ that extension-generated UUID would work for everything since we (probably) just need them to be distinct and serve as a way to target a specific binding instance.

PR #1022, opened 2026-05-25
Ignore capitalization when annotating words

ShanaryS:
Annotations will now ignore capitalization when matching tokens. This requires no change to the data stored, only using `anyOfIngoreCase()` instead of `anyOf()` for the db functions and normalizing the tokens for when using in `SubtitleAnnotations`. This required also reestablishing the db invariants around source priority withing `SubtitleAnnotations`.

Yomitan always returns the lower case version so normalization was already handle for the `ANY_FORM_COLLECTED` strategy. Now this allows the same flexibility for `EXACT_FORM_COLLECTED` and `LEMMA_FORM_COLLECTED`.

I think this makes sense to add without an option, the normalization only applies to English letters. There may be languages where this could be a legitimate difference in meaning but I don't think it's worth worrying about for now. We likely would need specific rules for those as we come across them if they exist.

### DIFF
--- a/common/subtitle-annotations/subtitle-annotations.ts
+++ b/common/subtitle-annotations/subtitle-annotations.ts
@@ -35,6 +35,7 @@ import {
     TokenState,
     TokenStatus,
     TokenStyling,
+    isWordSource,
 } from '@project/common/settings';
 import { TokenStatusInfo, DictionaryProvider, LemmaResults, TokenResults } from '@project/common/dictionary-db';
 import {
@@ -340,6 +341,28 @@ class TokenCollection extends TokenCollectionBase<TokenStatusResult> {
         );
         this.ts.updateTokenStates(normalizedKey, states);
     }
+
+    private resolve(
+        normalizedTokens: string[],
+        sourceMatches: (source: DictionaryTokenSource) => boolean
+    ): TokenStatusResult[] {
+        const statusResults: TokenStatusResult[] = [];
+        for (const normalizedToken of normalizedTokens) {
+            const statusResult = this.collection.get(normalizedToken);
+            if (statusResult && sourceMatches(statusResult.source)) statusResults.push(statusResult);
+        }
+        return statusResults;
+    }
+
+    resolveForWord(normalizedTokens: string[]): TokenStatusResult[] {
+        if (!this.wordEnabled) return [];
+        return this.resolve(normalizedTokens, (source) => isWordSource(source));
+    }
+
+    resolveForSentence(normalizedTokens: string[]): TokenStatusResult[] {
+        if (!this.sentenceEnabled) return [];
+        return this.resolve(normalizedTokens, (source) => !isWordSource(source));
+    }
 }
 
 class TokenCollectionArray extends TokenCollectionBase<TokenStatusResult[]> {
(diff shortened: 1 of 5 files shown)


common/subtitle-annotations/subtitle-annotations.ts
```diff
@@ -162,6 +172,82 @@ function getAnyFormStatusResults(
...
+    return {
+        ...(compareTokenStatusResults(left, right) >= 0 ? left : right),
+        externalCandidateStatuses: dedupeTokenStatusInfos([
+            ...(left.externalCandidateStatuses ?? []),
+            ...(right.externalCandidateStatuses ?? []),
+        ]),
+    };
+}
+
+function addCollectedAnyFormStatus(
```
killergerbah: Would it be more readable for `collection: Map<string, TokenStatusResult[]>` to be wrapped by a class with state mutating methods for `setTokenStates`, `addTokenIndex` etc? It might improve readability and separation of concerns.

common/subtitle-annotations/subtitle-annotations.ts
```diff
@@ -162,6 +172,82 @@ function getAnyFormStatusResults(
...
+    if (sourcePriority !== 0) return sourcePriority;
+    const statusPriority = left.status - right.status;
+    if (statusPriority !== 0) return statusPriority;
+    if (isExternalWordSource(left.source) && isExternalWordSource(right.source)) {
+        return externalWordSourcePriority(left.source) - externalWordSourcePriority(right.source);
+    }
+    return 0;
+}
+
+function mergeTokenStatusResults(left: TokenStatusResult, right: TokenStatusResult): TokenStatusResult {
```
killergerbah: I only understand the code in this file at a very surface level. What's the reason for adding this merge logic?

ShanaryS: The main complication by ignoring case is that now multiple tokens need to resolve to the same normalized key so that `Run` and `run` both gets combined. This means that `SubtitleAnnotations` needs to be aware of the db invariants about source priority and tie-breaks which is what `compareTokenStatusResults()` does. So that local tokens are still preferred, words over sentence tokens, and finally by highest status.

We also need to merge `externalCandidateStatuses` regardless of whatever token ends up winning. The projected stats PR fixed this bug in the original implementation where stats would only show if the Anki token ended up winning, when it should show regardless if a local token ended up being chosen (or now a WaniKani token). This merging here preserves the the invariant that "If the token in this media matches with Anki or WaniKani data then stats needs to be aware of it regardless if it's being used for annotations".

common/subtitle-annotations/subtitle-annotations.ts
ShanaryS: Yep, this was well overdue. As new features were added there became too many stray functions with unclear usage. So now I created two classes, one for `TokenState` so we can group the functions that operate on the track data. Then another for the tokens cached from the db, which also allows grouping those functions together.

I also finally got around to simplifying the logic around the strategy and priority. There has always been a lot of duplicated code there that's now been simplified with the help of the classes..

killergerbah (approved):
After the refactor the code looks a lot cleaner

Feel free to merge when ready

PR #1026, opened 2026-05-27
Stop subtitle picker resetting while user is interacting with it

jprostko:
### Summary

The subtitle picker can be reset and the video auto-resumed while a user is actively interacting with the picker, on any site whose `<video>` element fires `loadedmetadata` outside an actual video change, or whose player explicitly calls `video.play()` on an internal timer.

Two mechanisms feed the same symptom:

**1. `loadedmetadata` firings re-fetch synced data, wipe the user's in-progress picker selection, and clear loaded subtitles.** The binding's `videoChangeListener` fires on `loadedmetadata` and calls `videoDataSyncController.requestSubtitles()` (which clears `_syncedData` and `_autoSyncAttempted`) plus `_resetSubtitles()` (which clears the loaded subtitles from the player). When the resulting fresh `asbplayer-synced-data` event arrives at `_setSyncedData`, the auto-sync evaluation regenerates the picker's contents from the new data and pushes the update into the UI, overwriting any selection the user had made but not yet confirmed.

**2. Internal `setTimeout` calls `video.play()` while the picker is open**, un-pausing the video even though `_prepareShow` had paused it via `this._context.pause()`. asbplayer's pause is set once and never re-enforced.

Hulu exhibits both behaviors. Disney+ does not (verified by probe that `loadedmetadata` never fires while paused). Amazon Prime also does not exhibit the symptom in testing. Verified Hulu's `play()` call by probing `video.play` and seeing `app.js` invoke it via `setTimeout` (fires roughly 10 seconds after pause). Direct instrumentation in a follow-up testing session showed that the `setTimeout` callback also rotates the MediaSource blob URL, which is what fires `loadedmetadata`. The rotation was observed only for the first two pauses of a fresh session; subsequent pauses (including pauses of 117+ seconds) did not trigger it. The exact gating condition is not fully characterized but appears tied to early-session DRM license or manifest state. The synced-state guard short-circuits the bug regardless of how narrow the natural trigger turns out to be in practice.

A related correctness issue: if the user has the picker open during a true SPA soft-navigation (e.g. Hulu autoplay or "Next Episode" while the picker is up), the open-picker guard would suppress the legitimate refresh for the new video and leave the picker stale on the previous episode's tracks. The fix gates the suppression on the href matching the one captured when the picker opened, so spurious `loadedmetadata` firings on the same video are still suppressed, but soft-navigations correctly dismiss and refresh. A second guard separately suppresses refresh when subtitles are already synced for the current href, so spurious firings during normal watching (picker closed) also do not clear loaded subtitles.

### Blast radius

The *symptom* is only observed on Hulu so far, but the *code paths* this PR modifies are shared by every page script asbplayer binds to:

- `binding.ts` `videoChangeListener` (fires on every bound video element's `loadedmetadata`), plus synced-state bookkeeping in `_updateSubtitles`, `_resetSubtitles`, and `unbind`
- `video-data-sync-controller.ts` `requestSubtitles`, `_setSyncedData`, `_prepareShow`, `_hideAndResume`, `unbind`, and the new explicit `cancel` handler

The new guards should be no-ops for sites that do not fire media-change events while the picker is open, and the play-blocker only has an effect if the page or player tries to resume playback while the picker is open. Because the code paths are shared, regression testing should sample across the streaming ecosystem rather than only the obviously-affected sites.

### Changes

**`extension/src/services/binding.ts`**

The `videoChangeListener` body now has two guards. The first skips refresh when the picker is already open at `window.location.href` (captured when the picker opened). The second skips refresh when subtitles are already synced for `window.location.href` (tracked via `_lastSyncedLocation`, set in `_updateSubtitles` and cleared in `_resetSubtitles` and `unbind`). While either guard holds, `_updateRegisteredVideoSrc` still runs (internal src tracking) but `requestSubtitles` and `_resetSubtitles` are skipped, so subs the user already had loaded stay loaded across spurious `loadedmetadata` firings. On a true soft-navigation (different href), both guards fall through and the listener refreshes for the new video.

**`extension/src/controllers/video-data-sync-controller.ts`**

New public `pickerVisible` and `openedLocation` getters expose picker state (visibility, and the href captured when the picker opened) to the binding without poking at private fields.

`requestSubtitles` returns early when the picker is open on the same href, so spurious `loadedmetadata` firings can no longer clear `_syncedData` or re-dispatch `asbplayer-get-synced-data` while the user is interacting. When called with the picker open on a different href (a soft-navigation), it dismisses the stale picker via `_hideAndResume` and proceeds with the fresh fetch.

`_setSyncedData` gates the auto-sync apply (`_syncData(autoSelectedTracks)`), the failure prompt (`show(...)`), and the trailing model `updateState` on `!this.pickerVisible`. One exception preserves the loading-to-loaded transition: if the picker was opened before subtitle data had arrived (`_syncedData.subtitles === undefined`), the first `_setSyncedData` event still pushes a model update into the visible picker so it can move from loading to populated. `wasLoading` is captured from `_syncedData?.subtitles` being undefined before the new data is assigned.

`_prepareShow` captures `window.location.href` and installs a `play` event listener on the video that immediately re-pauses via `this._context.pause()`. Both are cleaned up from `_hideAndResume`: the location is cleared, and the listener is removed via a new `_cleanupPlayBlocker` helper.

`unbind` now also calls `_cleanupPlayBlocker` and clears the captured location, so neither the play listener nor the picker-session marker leaks if the controller is torn down with the picker still open.

A new explicit `cancel` command handler now calls `_hideAndResume` instead of relying on cancel falling through the default `dataWasSynced = true` path implicitly.

### Tested

The below were tested successfully (sites not tested are noted).

Direct verification of the fix:

- Hulu: video playing, open picker, change track without confirming, wait 30+ seconds. Hulu's setTimeout fires play() during that window. Verify the video does not visibly resume (the play blocker re-pauses immediately). Dropdown stays where you put it. Cancel restores playback with the previously-applied track. Confirm applies the new one.
- Hulu soft-nav: open picker on Episode 1 (do not confirm). Trigger Next Episode (autoplay or click). The picker should dismiss and refresh for Episode 2's tracks instead of staying on Episode 1's.

Regression smoke across the ecosystem (sampling sites that bind to asbplayer):

- Direct inferTracks callers (Disney+, SVT Play): open subtitle picker. Tracks load. Cancel and Confirm behave as before. Auto-sync on initial page load (picker hidden) still applies the saved preference correctly. (Did not test Bandai Channel, NRK TV, TVer, Bilibili, but ideally they would receive testing as well.)
- Helper-routed inferTracks users (HBO Max, UR Play): smoke test that subtitle picker still loads on a representative video. (Did not test Disney+ Apps, iWantTFC, OSN Plus, Viki, Yle Areena, but ideally they would receive testing as well.)
- Page-scripted sites that don't use inferTracks (Amazon Prime, Netflix): subtitle picker still opens and existing subtitle workflows behave as before. (Did not test Emby/Jellyfin, Plex, Stremio, but ideally they would receive testing as well.)
- Binding-only sites without page scripts (Twitch, Archive.org): subtitle picker still reachable via file open / drag, no regression in behavior.

Cold-load edge case:

- Open a watch URL on any `inferTracks`-using site and immediately invoke the subtitle picker shortcut. Picker should transition from loading state to populated once tracks arrive (not stay stuck in loading).

### Known limitations

If the picker opens while subtitle data is still loading and the user edits the Video Name field before tracks arrive, the loading-to-loaded refresh may overwrite their edit (inferred from how `updateState` rebuilds the model, not verified empirically against the React UI). This is narrow (there is little useful metadata to edit before tracks appear anyway), and the alternative (leaving the picker stuck in loading state forever until reopened) is worse.

Separately, the href-based soft-navigation detection assumes the SPA updates `window.location.href` before or at the same time as the new video's `loadedmetadata` fires. On sites where the URL update lags the player event (Hulu's `/v6/playlist` was observed firing 343-620ms before the URL settled, for example), a fast next-episode transition could fire `loadedmetadata` ahead of the URL update, in which case the picker would not refresh. The fix is still a strict improvement over the prior behavior (which never refreshed on soft-nav), but a fully bulletproof solution would need a signal independent of `window.location.href` such as a page-script-supplied generation counter, which is out of scope here.

### Ordering against the Hulu subtitle rewrite

This PR has no code dependency on the Hulu subtitle rewrite (PR #1027). The changes here work on their own. But the bug it fixes was latent in asbplayer until the Hulu rewrite made synced-data responses reliable on repeated requests, which is what now exposes this picker-state behavior.

**Ideally, PR #1027 should land after or alongside this one** to avoid a window where Hulu users get the working subtitle detection fix but also start hitting the newly-exposed picker bug.

### Unrelated observation during testing

While running the test plan on Disney+, hit a "Data Sync failed: Failed to fetch" error when selecting Swedish subtitles in asbplayer after Disney+'s own native UI had already loaded the Swedish track. Other languages worked, only Swedish failed, and a page refresh fixed it. The error originates from `_syncDataArray` when one of the segmented subtitle fetches rejects. Most plausible cause is that Disney+'s segmented subtitle URLs are tokenized or single-use, so once Disney+'s own player has consumed them they cannot be re-fetched by the extension.

This code path is not modified by this PR and the behavior is pre-existing. Flagging as an observation, not something this PR addresses or claims to fix.

### Additional notes

### DIFF
--- a/extension/src/controllers/video-data-sync-controller.ts
+++ b/extension/src/controllers/video-data-sync-controller.ts
@@ -78,7 +78,7 @@ export default class VideoDataSyncController {
     private _syncedData?: VideoData;
     private _wasPaused?: boolean;
     private _playBlocker?: () => void;
-    private _openedPathname?: string;
+    private _openedLocation?: string;
     private _fullscreenElement?: Element;
     private _activeElement?: Element;
     private _autoSyncAttempted: boolean = false;
@@ -118,7 +118,7 @@ export default class VideoDataSyncController {
         this._dataReceivedListener = undefined;
         this._syncedData = undefined;
         this._cleanupPlayBlocker();
-        this._openedPathname = undefined;
+        this._openedLocation = undefined;
     }
 
     updateSettings({ streamingAutoSync, streamingLastLanguagesSynced }: AsbplayerSettings) {
@@ -146,8 +146,8 @@ export default class VideoDataSyncController {
         return !this._frame.hidden;
     }
 
-    get openedPathname(): string | undefined {
-        return this._openedPathname;
+    get openedLocation(): string | undefined {
+        return this._openedLocation;
     }
 
     async requestSubtitles() {
@@ -155,11 +155,11 @@ export default class VideoDataSyncController {
             return;
         }
 
-        // While the picker is open on the same path, skip refresh so cyclic
+        // While the picker is open on the same location, skip refresh so
         // player events do not clobber an in-progress user selection. On a
         // true soft-navigation, dismiss the stale picker and continue.
         if (this.pickerVisible) {
-            if (this.openedPathname !== undefined && window.location.pathname !== this.openedPathname) {
+            if (this.openedLocation !== undefined && window.location.href !== this.openedLocation) {
                 this._hideAndResume();
             } else {
                 return;
@@ -461,7 +461,7 @@ export default class VideoDataSyncController {
     }
 
     private _prepareShow() {
-        this._openedPathname = window.location.pathname;
+        this._openedLocation = window.location.href;
         this._wasPaused = this._wasPaused ?? this._context.video.paused;
         this._context.pause();
 
@@ -498,7 +498,7 @@ export default class VideoDataSyncController {
 
     private _hideAndResume() {
         this._cleanupPlayBlocker();
-        this._openedPathname = undefined;
+        this._openedLocation = undefined;
         this._context.keyBindings.bind(this._context);
         this._context.subtitleController.forceHideSubtitles = false;
         this._context.mobileVideoOverlayController.forceHide = false;
--- a/extension/src/controllers/video-data-sync-controller.ts
+++ b/extension/src/controllers/video-data-sync-controller.ts
@@ -159,7 +159,7 @@ export default class VideoDataSyncController {
         // player events do not clobber an in-progress user selection. On a
         // true soft-navigation, dismiss the stale picker and continue.
         if (this.pickerVisible) {
-            if (this._openedPathname !== undefined && window.location.pathname !== this._openedPathname) {
+            if (this.openedPathname !== undefined && window.location.pathname !== this.openedPathname) {
                 this._hideAndResume();
             } else {
                 return;
@@ -327,22 +327,22 @@ export default class VideoDataSyncController {
                 this._autoSyncAttempted = true;
                 const subs = this._matchLastSyncedWithAvailableTracks();
 
-                if (subs.completeMatch && this._frame.hidden) {
+                if (subs.completeMatch && !this.pickerVisible) {
                     const autoSelectedTracks: VideoDataSubtitleTrack[] = subs.autoSelectedTracks;
                     await this._syncData(autoSelectedTracks);
-                } else if (!subs.completeMatch && this._frame.hidden) {
+                } else if (!subs.completeMatch && !this.pickerVisible) {
                     const shouldPrompt = await this._settings.getSingle('streamingAutoSyncPromptOnFailure');
 
                     if (shouldPrompt) {
                         await this.show({ reason: VideoDataUiOpenReason.failedToAutoLoadPreferredTrack });
                     }
-                } else if (this._frame.clientIfLoaded !== undefined && wasLoading) {
+                } else if (wasLoading) {
                     // Picker is open in loading state. Populate it now that tracks have arrived.
-                    this._frame.clientIfLoaded.updateState(await this._buildModel({}));
+                    this._frame.clientIfLoaded?.updateState(await this._buildModel({}));
                 }
             }
-        } else if (this._frame.clientIfLoaded !== undefined && (this._frame.hidden || wasLoading)) {
-            this._frame.clientIfLoaded.updateState(await this._buildModel({}));
+        } else if (!this.pickerVisible || wasLoading) {
+            this._frame.clientIfLoaded?.updateState(await this._buildModel({}));
         }
     }
 
@@ -470,7 +470,7 @@ export default class VideoDataSyncController {
         // picker is dismissed.
         if (!this._playBlocker) {
             this._playBlocker = () => {
-                this._context.video.pause();
+                this._context.pause();
             };
             this._context.video.addEventListener('play', this._playBlocker);
         }
--- a/extension/src/controllers/video-data-sync-controller.ts
+++ b/extension/src/controllers/video-data-sync-controller.ts
@@ -77,6 +77,8 @@ export default class VideoDataSyncController {
     private _emptySubtitle: VideoDataSubtitleTrack;
     private _syncedData?: VideoData;
     private _wasPaused?: boolean;
+    private _playBlocker?: () => void;
+    private _openedPathname?: string;
     private _fullscreenElement?: Element;
     private _activeElement?: Element;
     private _autoSyncAttempted: boolean = false;
@@ -115,6 +117,8 @@ export default class VideoDataSyncController {
 
         this._dataReceivedListener = undefined;
         this._syncedData = undefined;
+        this._cleanupPlayBlocker();
+        this._openedPathname = undefined;
     }
 
     updateSettings({ streamingAutoSync, streamingLastLanguagesSynced }: AsbplayerSettings) {
@@ -138,9 +142,28 @@ export default class VideoDataSyncController {
         }
     }
 
+    get pickerVisible(): boolean {
+        return !this._frame.hidden;
+    }
+
+    get openedPathname(): string | undefined {
+        return this._openedPathname;
+    }
+
     async requestSubtitles() {
         if (!this._context.hasPageScript) {
             return;
+        }
+
+        // While the picker is open on the same path, skip refresh so cyclic
+        // player events do not clobber an in-progress user selection. On a
+        // true soft-navigation, dismiss the stale picker and continue.
+        if (this.pickerVisible) {
+            if (this._openedPathname !== undefined && window.location.pathname !== this._openedPathname) {
+                this._hideAndResume();
+            } else {
+                return;
+            }
         }
 
         const pageDelegate = await currentPageDelegate();
@@ -296,6 +319,7 @@ export default class VideoDataSyncController {
     }
 
     private async _setSyncedData(data: VideoData) {
+        const wasLoading = this._syncedData?.subtitles === undefined;
         this._syncedData = data;
 
         if (this._syncedData?.subtitles !== undefined && (await this._canAutoSync())) {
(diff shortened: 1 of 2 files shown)


extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -296,29 +319,29 @@ export default class VideoDataSyncController {
...
                     if (shouldPrompt) {
                         await this.show({ reason: VideoDataUiOpenReason.failedToAutoLoadPreferredTrack });
                     }
+                } else if (this._frame.clientIfLoaded !== undefined && wasLoading) {
+                    // Picker is open in loading state. Populate it now that tracks have arrived.
+                    this._frame.clientIfLoaded.updateState(await this._buildModel({}));
                 }
             }
-        } else if (this._frame.clientIfLoaded !== undefined) {
+        } else if (this._frame.clientIfLoaded !== undefined && (this._frame.hidden || wasLoading)) {
```
killergerbah: Since both of these `this._frame.clientIfLoaded !== undefined` checks are in the last `else` block anyway, would it be more readable to to omit that check and write `this._frame.clientIfLoaded?.updateState(await this._buildModel({}));` instead?

extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -296,29 +319,29 @@ export default class VideoDataSyncController {
...
+        const wasLoading = this._syncedData?.subtitles === undefined;
         this._syncedData = data;
 
         if (this._syncedData?.subtitles !== undefined && (await this._canAutoSync())) {
             if (!this._autoSyncAttempted) {
                 this._autoSyncAttempted = true;
                 const subs = this._matchLastSyncedWithAvailableTracks();
 
-                if (subs.completeMatch) {
+                if (subs.completeMatch && this._frame.hidden) {
```
killergerbah: I'm a little nervous about blocking logical state changes with a visual state check. You might have explained it already in the PR description, but what was the reason for blocking subtitle auto-sync with `this._frame.hidden`?

The other `this._frame.hidden` guards make more sense to me because they are blocking visual state changes only

extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -433,9 +461,20 @@ export default class VideoDataSyncController {
...
+        this._openedPathname = window.location.pathname;
         this._wasPaused = this._wasPaused ?? this._context.video.paused;
         this._context.pause();
 
+        // Some players (e.g. Hulu) call video.play() on an internal timer that
+        // ignores the picker being open. Re-pause on any play event until the
+        // picker is dismissed.
+        if (!this._playBlocker) {
+            this._playBlocker = () => {
+                this._context.video.pause();
```
killergerbah: `this._context.pause()` feels safer, since it accounts for site-specific pause behavior, and will probably continue to do so for future sites besides Netflix.

extension/src/controllers/video-data-sync-controller.ts
jprostko: Applied, thanks. Dropped the explicit `clientIfLoaded !== undefined` checks from both branches. The outer branch's condition line also includes the change from `_frame.hidden` to `!this.pickerVisible` (part of the rename I made across all three `_setSyncedData` checks, which will be explained shortly in the upcoming response to your auto-sync visibility comment below), since both edits land on the same line.

extension/src/controllers/video-data-sync-controller.ts
jprostko: The reasoning for this is that the guard prevents auto-sync from mutating what is loaded into the subtitle controller while the picker is open. `_syncData` fetches the saved preference's subtitle files and calls `_syncSubtitles`, which changes the loaded tracks. Without the gate, any re-entry into `_setSyncedData` while the picker is open would silently load the saved preference's tracks. The fetch is wasted, and if the user hits Cancel rather than confirming a different choice, those tracks can become visible on the player after the picker closes, even though the user did not pick them during that picker session.

The gate is effectively a no-op for sites that don't fire `_setSyncedData` while the picker is open. Other sites' auto-sync still runs on initial page load, and for sites that don't have Hulu's cyclic-event pattern, the gate condition simply doesn't trigger. The gate is shaped to be additive, not invasive.

Also, during this latest push (eda1874), I changed `_frame.hidden` to `!this.pickerVisible` at the three `_setSyncedData` checks for naming consistency with the binding-side `videoChangeListener` guard. (I also swapped `_openedPathname` to `this.openedPathname` at its one read site in `requestSubtitles` for the same reason. `binding.ts` already consumes that getter too, so this makes the controller's internal reads go through the same getters consistently.)

Since `pickerVisible` is just `!_frame.hidden` under the hood, it's only a naming improvement though, not actual decoupling, obviously. If you'd prefer real decoupling, I can introduce a separate logical state like `_pickerSessionActive` toggled in `_prepareShow`/`_hideAndResume` and gate these checks on it instead. The flag would be set when the picker opens and cleared when it closes, so its value always matches visibility. Behavior would be unchanged, and the gate would read as "is the user in a picker session?" rather than "is the iframe hidden?".

Happy to make that change if you'd prefer. It shouldn't take much work to implement, and is the better way to go since the gate would actually decouple from visibility rather than just renaming around it.

extension/src/controllers/video-data-sync-controller.ts
jprostko: Good call. I applied it as you suggested, switching `_playBlocker` to call `this._context.pause()`.

extension/src/controllers/video-data-sync-controller.ts
killergerbah: Understood, thanks for the detailed explanation. I agree it makes sense for the picker session to block subtitle auto-loading. At that point, the user is actually trying to pick subtitles and likely does not want subtitles to auto-load, or the picker to close from under them.

Regarding the new logical state, I think there are advantages in both directions. If tracking new state then it doesn't depend directly on `this.frame.hidden` but if not tracking new state, then there is less state to track and less chance of inconsistency. I don't have a strong opinion so I'll leave it up to you if you which way you want to go.

jprostko: Thanks. I'll keep it on `pickerVisible` for now. The separate flag would decouple the check from frame visibility, but since both would flip at the same picker open/close boundaries, adding the flag would just duplicate state without capturing anything different. If the picker lifecycle ever diverges from frame visibility, that would be the right time to introduce a dedicated session flag.

jprostko:
Converted this back to a draft due to me finding another Hulu bug which also made me find a bug that I inadvertently introduced that would negatively impact Emby and Jellyfin. The fixes can all live in this PR though once I get it all fixed.

jprostko:
Follow-up commit adds a synced-state guard so the bug is suppressed during normal watching too (not just while the picker is open), and switches the same-video identity checks from `pathname` to `href` across all sites, so hash-based video identity (e.g. Emby/Jellyfin) is handled correctly. PR description updated with the full rationale and refines the earlier characterization of Hulu's `loadedmetadata` behavior based on direct instrumentation.

killergerbah (approved):
The change makes sense to me. It should be noted that some sites don't change `href` at all, but can change the video (like Apple TV) so we might need to make site-specific exceptions to this behavior. But I think this change should cover the vast majority of cases.

jprostko:
> The change makes sense to me. It should be noted that some sites don't change `href` at all, but can change the video (like Apple TV) so we might need to make site-specific exceptions to this behavior. But I think this change should cover the vast majority of cases.

Yes, that is true. I kind of hinted at something along the lines of that at the end of the **Known limitations** section, but yes, it is something we will definitely need to keep in mind.

PR #1027, opened 2026-05-27
Fix Hulu silently dropping non-first subtitle tracks

jprostko:
### Summary

The Hulu page script's subtitle dedup filter was unconditionally dropping every track after the first whenever Hulu's webvtt object keys were already lowercase, which is the common case. Symptom: only English subtitles ever appeared, and Spanish (or any other language) was silently missing from the picker.

The old `extractSubtitleTracks` dedup check compared each existing subtitle's own `label` to its own `language`. But the loop builds entries with `label: language, language: language.toLowerCase()`, so for an already-lowercase key the two are equal, the filter matches the just-added track, and the next iteration's push is skipped. Verified by capturing a real `/v6/playlist` response in a live tab. The webvtt keys came back as `["en", "es"]`.

This PR refactors `hulu-page.ts` to use `inferTracks` (the shared helper used directly or through page utilities by other streaming scripts) plus a `window.fetch` wrapper that observes Hulu's own `/v6/playlist` response. Basename comes from a DOM scrape of `[data-testid="player-metadata"]`. This eliminates the previous architecture's duplicate POST to `/v6/playlist` (whose response carried session tokens like `sauron_id` and `sauron_token` and risked registering a second Hulu playback session against the user's account) and the `/upnext` GET.

Cached tracks are keyed by entityId extracted from the response's `content_eab_id` field, not by `window.location.pathname`. Hulu's in-player next-video button issues the new video's `/v6/playlist` request several hundred milliseconds before it updates the URL via `replaceState`. Keying by pathname at fetch time would file the new video's tracks under the previous URL. Keying by the `entityId` from the response is the reliable identifier for this flow.

The PR also adds an `ignoreVideoElements` config in `pages.json` for Hulu's `VideoPreviewPlayer` element (the up-next preview player). It has a valid blob `src` and would otherwise become a second binding, triggering asbplayer's multi-video element picker every time the subtitle picker is invoked.

### Tested

Testing was done on hulu.com in the US with various movies/series, and they work as desired.  Testing done included:

- Verified Hulu episode with English + Spanish subtitles shows both tracks.
- Verified in-player next-video navigation shows the new video's tracks and basename.
- Verified subtitle picker opens without the multi-video element prompt.
- Spot-checked episode and movie basename extraction.

### Known limitation

The new flow passively observes Hulu's own `/v6/playlist` fetch rather than actively re-POSTing as the previous implementation did. If the page script injects after Hulu has already fetched the playlist, such as when the extension is enabled late or asbplayer is toggled mid-session, there is nothing to observe and subtitles will be unavailable until the next `/v6/playlist` fires, typically on episode navigation or page reload. This is a deliberate tradeoff to avoid the duplicate POST and second-session-token risk in the previous active re-fetch approach.

### Ordering against the subtitle picker fix

This PR has no code dependency on the subtitle picker stability fix (PR #1026). The changes here work on their own. But the Hulu rewrite makes synced-data responses reliable on repeated requests, which exposes a latent picker-state bug on Hulu that the picker stability PR fixes.

**Ideally, PR #1026 should land before or alongside this one** to avoid a window where Hulu users get working subtitle detection but also start hitting the newly-exposed picker bug.

### Additional notes

### DIFF
--- a/extension/src/entrypoints/hulu-page.ts
+++ b/extension/src/entrypoints/hulu-page.ts
@@ -1,151 +1,129 @@
-import { VideoDataSubtitleTrack } from '@project/common';
-import { extractExtension, trackFromDef } from '@/pages/util';
+import { VideoDataSubtitleTrackDef } from '@project/common';
+import { extractExtension, inferTracks, poll } from '@/pages/util';
 
 export default defineUnlistedScript(() => {
-    setTimeout(() => {
-        function isObject(val: any) {
-            return typeof val === 'object' && !Array.isArray(val) && val !== null;
+    const tracksByEntityId = new Map<string, VideoDataSubtitleTrackDef[]>();
+
+    function entityIdFromEabId(eabId: unknown): string | undefined {
+        if (typeof eabId !== 'string') {
+            return undefined;
         }
+        const m = eabId.match(/^EAB::([0-9a-fA-F-]+)/);
+        return m ? m[1] : undefined;
+    }
 
-        function extractSubtitleTracks(value: any) {
-            const subtitles = [];
-            if (isObject(value.transcripts_urls?.webvtt)) {
-                const urls = value.transcripts_urls.webvtt;
+    function entityIdFromPathname(pathname: string): string | undefined {
+        const m = pathname.match(/^\/watch\/([0-9a-fA-F-]+)/);
+        return m ? m[1] : undefined;
+    }
 
-                for (const language of Object.keys(urls)) {
-                    const url = urls[language];
-
-                    if (typeof url === 'string') {
-                        if (subtitles.find((s) => s.label === s.language) === undefined) {
-                            subtitles.push(
-                                trackFromDef({
+    const originalFetch = window.fetch;
+    window.fetch = function (...args) {
+        // @ts-ignore
+        const promise = originalFetch.apply(this, args);
+        const input = args[0];
+        const url =
+            typeof input === 'string'
+                ? input
+                : input instanceof Request
+                  ? input.url
+                  : input instanceof URL
+                    ? input.href
+                    : '';
+        if (url.includes('play.hulu.com/v6/playlist')) {
+            promise
+                .then((response) => response.clone().json())
+                .then((json) => {
+                    // Key by entityId from the response itself. Hulu sometimes fires
+                    // the next video's playlist before updating window.location, so
+                    // the URL is not a reliable identifier at fetch time.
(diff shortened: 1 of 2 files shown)


extension/src/entrypoints/hulu-page.ts
```diff
@@ -1,151 +1,129 @@
...
+            typeof input === 'string'
+                ? input
+                : input instanceof Request
+                  ? input.url
+                  : input instanceof URL
+                    ? input.href
+                    : '';
+        if (url.includes('play.hulu.com/v6/playlist')) {
+            promise
+                .then((response) => response.clone().json())
```
killergerbah, line 35: When I wrote the original script I didn't know about this method of intercepting the JSON from a `fetch` request, which is probably why I decided to re-fetch instead. This is much better.

killergerbah (approved):
I don't have a Hulu subscription right now so I can't test this but the code looks good and I trust that it works. Thanks as always.

extension/src/entrypoints/hulu-page.ts
```diff
@@ -1,151 +1,129 @@
...
+            const series = pm.querySelector('span')?.textContent?.trim();
+            if (!series) {
+                continue;
+            }
 
-        let upnextController: AbortController | undefined;
+            const divs = Array.from(pm.querySelectorAll('div')).map((d) => (d.textContent || '').trim());
+            const seasonEpIdx = divs.findIndex((t) => /^S\d+\s+E\d+$/.test(t));
+            if (seasonEpIdx === -1) {
+                return series;
```
killergerbah, line 81: I suppose we would get here if viewing a movie which doesn't have an episode number?

extension/src/entrypoints/hulu-page.ts
jprostko: Thanks. The motivation was avoiding the duplicate POST and the session-token risk the active re-fetch carried along with it. Passive observation of Hulu's own response felt like the cleaner path once that was the goal.

extension/src/entrypoints/hulu-page.ts
jprostko: Yes, that's the movie path.

For movies, the player-metadata DOM lacks an `S# E#` element, so `seasonEpIdx === -1` and we return the first span's textContent. That text is the series name for episodes (combined with `seasonEp` and the episode title in the other branches) and the movie title for movies, so returning it alone produces the right basename in both cases. I verified this on several movies during testing.

PR #1030, opened 2026-05-27
Fix: seek-timestamp command active tab restriction

mcgrizzz:
- [x] Ran `yarn verify` - All passed

Current Issue:

- The seek-timestamp command will seek on all video tabs instead the active/focused tab. 

The fix:

- Only send the seek command to the active tab, using the same browser tabs query
  pattern already used in `client.onMineSubtitle`.


killergerbah:
Looks good, thanks

PR #1031, opened 2026-05-28
retain newlines from mecab tokenize

ShanaryS:
I can't believe I missed this during the original implementation but the `MeCab` parser drops newlines at the end of tokens unlike the `scanning-parser`. This means subtitles with multiple lines were being reduced to a single line.

The solution is to record the positions of the newlines before sending to be to tokenize so that we can re-add them at the correct positions after.

### DIFF
--- a/common/yomitan/yomitan.ts
+++ b/common/yomitan/yomitan.ts
@@ -359,6 +359,14 @@ export class Yomitan {
         }
     }
 
+    verifyTokenizeResult(originalText: string, tokenizeRes: TokenPart[][]): void {
+        const originalTextFromTokenize = tokenizeRes.map((t) => t.map((p) => p.text).join('')).join('');
+        if (originalTextFromTokenize === originalText) return;
+        throw new Error(
+            `Tokenize result does not match the original text:\n${originalText}\n--->\n${originalTextFromTokenize}`
+        );
+    }
+
     private extractLemmaFromMecab(token: string, tokenPart: TokenPartResult): void {
         if (!this.getSupportsMecabLemma()) return;
         const lemmas: string[] = [];
--- a/common/yomitan/yomitan.ts
+++ b/common/yomitan/yomitan.ts
@@ -1,6 +1,14 @@
 import { Fetcher, HttpFetcher, Progress } from '@project/common';
 import { DictionaryTrack } from '@project/common/settings';
-import { AsyncSemaphore, fromBatches, HAS_LETTER_REGEX, inBatches, isKanaOnly } from '@project/common/util';
+import {
+    AsyncSemaphore,
+    fromBatches,
+    HAS_LETTER_REGEX,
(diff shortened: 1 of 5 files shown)


common/yomitan/yomitan.ts
```diff
@@ -309,15 +324,26 @@ export class Yomitan {
...
+    private cacheFromTokenize(
+        tokenizeResult: TokenizeResult,
+        tokensForText: TokenPartResult[][],
+        newlines: { text: string; index: number }[]
+    ): void {
+        let currIndex = 0;
         for (const tokenParts of tokenizeResult.content) {
-            tokensForText.push(tokenParts);
             const tokenPart = tokenParts[0];
-            if (!tokenPart) return;
```
killergerbah, line 316: If `tokenPart` is empty here does that mean it's empty for the rest of the loop? If it's not empty for the rest of the loop then the code will fall throug to the lines below more than before

common/yomitan/yomitan.ts
```diff
@@ -309,15 +324,26 @@ export class Yomitan {
...
-                .join('')
-                .trim();
+            if (!tokenPart) continue;
+
+            const tokenText = tokenParts.map((p) => p.text).join('');
+            currIndex += tokenText.length;
+            while (newlines.length && newlines[0].index < currIndex) {
+                const { text } = newlines.shift()!;
+                if (tokenText.includes(text)) continue; // scanning-parser includes newlines
+                tokensForText.push([{ text, reading: '' }]);
```
killergerbah, line 342: Since this line appears before `tokensForText.push(tokenParts`, does this mean that `tokensForText` is receiving the newline characters before receiving `tokenParts`?

ShanaryS: It's due to how the index boundary is being checked. When we enter this while loop we are checking if we need a newline before this current token. For `scanning-parser` the newline is present as the current token so we do nothing.

I originally had `tokensForText.push(tokenParts)` before the while loop thinking the same as you but testing showed it was an off-by-one error.

common/yomitan/yomitan.ts
ShanaryS: For `scanning-parser` it will always be present since no empty arrays are pushed in the Yomitan code. But mecab parsing isn't as defensive so it's technically possible to have empty arrays pushed but I'm not sure if it will ever happen, it would probably be some parsing error. But I think continue would be better than return as it means skipping a word rather than the rest of the sentence.

In any case, I now added a check against the tokenize result to confirm if it matches what was present in the subtitle before. I confirmed it also catches this newline issue which would have tipped me off earlier. Now if tokens are missing/changed, we will get the strikethrough error with logs of the expected and new instead of silent failures.

killergerbah (approved):
Looks good to me

Feel free to merge when ready

PR #1033, opened 2026-05-30
Feat/get-bound-media command

mcgrizzz:

This PR adds a new command `get-bound-media` as per part of the discussion here: https://github.com/asbplayer/asbplayer/issues/840

The command returns a list of currently bound media by asbplayer. Each has the following shape:

```ts
export interface BoundMedia {
    tabId: number; 
    src: string; 
    title?: string;
    faviconUrl?: string;
    subscribed: boolean;
    synced: boolean;
    loadedSubtitles: boolean;
    syncedTimestamp?: number;
    
    // Added ontop of the base VideoElement type
    //Is the tab active
    active: boolean; 
    //Is this the focused tab (you can have multiple windows with active tabs, but only one focused window)
    focused: boolean; 
    //Chrome/mozilla will unload tabs that are inactive, in theory this should be set in those cases
    discarded: boolean; 
}
```

### A note about the `browser.tabs.onReplaced` listener 

I noticed while testing that duplicate entries could appear for what seemed to be the same tab, even though there was only one instance of that tab/video open. This appears to happen in Chrome when a discarded/unloaded tab is restored: `browser.tabs.onRemoved` is not called, but `browser.tabs.onReplaced` is. This PR includes a fix for that case.

### DIFF
--- a/common/web-socket-client/web-socket-client.ts
+++ b/common/web-socket-client/web-socket-client.ts
@@ -40,6 +40,30 @@ export interface SeekTimestampCommand {
     };
 }
 
+export interface GetBoundMediaCommand {
+    command: 'get-bound-media';
+    messageId: string;
+    body: {};
+}
+
+export interface BoundMedia {
+    tabId: number;
+    src: string;
+    title?: string;
+    faviconUrl?: string;
+    subscribed: boolean;
+    synced: boolean;
+    loadedSubtitles: boolean;
+    syncedTimestamp?: number;
+    active: boolean;
+    focused: boolean;
+    discarded: boolean;
+}
+
+interface GetBoundMediaResponseBody {
+    media: BoundMedia[];
+}
+
 export class WebSocketClient {
     private _socket?: WebSocket;
     private _pingInterval?: NodeJS.Timeout;
@@ -50,6 +74,7 @@ export class WebSocketClient {
     onMineSubtitle?: (command: MineSubtitleCommand) => Promise<boolean>;
     onLoadSubtitles?: (command: LoadSubtitlesCommand) => Promise<void>;
     onSeekTimestamp?: (command: SeekTimestampCommand) => Promise<void>;
+    onGetBoundMedia?: () => Promise<BoundMedia[]>;
 
     get socket() {
         return this._socket;
@@ -130,6 +155,15 @@ export class WebSocketClient {
                             body: {},
                         };
                         this._socket?.send(JSON.stringify(response));
+                    } else if (payload.command === 'get-bound-media') {
+                        const messageId = payload.messageId;
+                        const media = (await this.onGetBoundMedia?.()) ?? [];
+                        const response: Response<GetBoundMediaResponseBody> = {
+                            command: 'response',
+                            messageId,
+                            body: { media },
+                        };
+                        this._socket?.send(JSON.stringify(response));
                     }
                 }
             };
@@ -199,5 +233,6 @@ export class WebSocketClient {
         this.onMineSubtitle = undefined;
         this.onSeekTimestamp = undefined;
         this.onLoadSubtitles = undefined;
+        this.onGetBoundMedia = undefined;
     }
 }
--- a/common/web-socket-client/web-socket-client.ts
+++ b/common/web-socket-client/web-socket-client.ts
@@ -1,3 +1,7 @@
+import type { SubtitleTrack } from '../src/model';
+
+export type { SubtitleTrack };
+
 export interface MineSubtitleCommand {
     command: 'mine-subtitle';
     messageId: string;
@@ -47,17 +51,12 @@ export interface GetBoundMediaCommand {
 }
 
 export interface BoundMedia {
-    tabId: number;
-    src: string;
+    id: string; // Derived from a hash of `streaming:<tabId>:<src>` or `local:<asbplayerId>`.
+    type: 'streaming' | 'local';
     title?: string;
     faviconUrl?: string;
-    subscribed: boolean;
-    synced: boolean;
-    loadedSubtitles: boolean;
-    syncedTimestamp?: number;
+    loadedSubtitles: SubtitleTrack[];
     active: boolean;
-    focused: boolean;
-    discarded: boolean;
 }
 
 interface GetBoundMediaResponseBody {
@@ -156,14 +155,16 @@ export class WebSocketClient {
                         };
                         this._socket?.send(JSON.stringify(response));
                     } else if (payload.command === 'get-bound-media') {
-                        const messageId = payload.messageId;
-                        const media = (await this.onGetBoundMedia?.()) ?? [];
-                        const response: Response<GetBoundMediaResponseBody> = {
-                            command: 'response',
-                            messageId,
-                            body: { media },
-                        };
-                        this._socket?.send(JSON.stringify(response));
+                        if (this.onGetBoundMedia !== undefined) {
+                            const messageId = payload.messageId;
+                            const media = await this.onGetBoundMedia();
+                            const response: Response<GetBoundMediaResponseBody> = {
+                                command: 'response',
+                                messageId,
+                                body: { media },
+                            };
+                            this._socket?.send(JSON.stringify(response));
+                        }
                     }
                 }
             };
--- a/docs/docs/reference/external-api.md
+++ b/docs/docs/reference/external-api.md
@@ -103,12 +103,64 @@ asbplayer, as a WebSocket client, responds to the following commands from a WebS
 }
 ```
 
+### `get-bound-media`
+
+Returns every `<video>` element asbplayer is currently tracking, keyed by tab and `src`.
+
+#### Request
+
+```javascript
+{
+    "command": "get-bound-media",
+    // Message ID to correlate with asbplayer's response
+    "messageId": "9f1c2b3a-4d5e-6f70-8190-a1b2c3d4e5f6",
+    "body": {}
+}
+```
+
+#### Response
+
+```javascript
+{
+    "command": "response",
+    // Same message ID received in request
+    "messageId": "9f1c2b3a-4d5e-6f70-8190-a1b2c3d4e5f6",
+    "body": {
+        "media": [{
+            // ID of the tab the video element lives in
+            "tabId": 123,
+            // The video element's src; serves as the media's identifier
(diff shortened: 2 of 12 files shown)


common/web-socket-client/web-socket-client.ts
```diff
@@ -40,6 +40,30 @@ export interface SeekTimestampCommand {
     };
 }
 
+export interface GetBoundMediaCommand {
+    command: 'get-bound-media';
+    messageId: string;
+    body: {};
+}
+
+export interface BoundMedia {
```
killergerbah, line 53: Since it's possible for other software to start relying on this structure I think it makes sense to start out conservative with what it includes:

- A lot of the fields don't have obvious use-cases (at least to me).
- Some of them are also implementation details that happen to have an internal use case for the extension to function, and exposing them would make them harder to change.

Unless you have use-cases that require other fields in mind I would suggest something a little cleaner like below.

```
export interface SubtitleTrack {
  name: string
  empty: boolean
}

export interface BoundMedia {
  id: string;
  type: "streaming" | "local";
  // This should probably be informed by how the `get-subtitles` API is designed
  loadedSubtitles: SubtitleTrack[];
  focused: boolean;
}
```

extension/src/services/web-socket-client-binding.ts
```diff
@@ -145,6 +146,37 @@ export const bindWebSocketClient = async (settings: SettingsProvider, tabRegistr
...
+        for (const tab of allTabs) {
+            if (tab.id !== undefined) {
+                tabStateById.set(tab.id, { active: tab.active ?? false, discarded: tab.discarded ?? false });
+            }
+        }
+
+        const focusedTabs = await browser.tabs.query({ active: true, lastFocusedWindow: true });
+        const focusedTabId = focusedTabs[0]?.id;
+
+        return videoElements.map((videoElement) => {
```
killergerbah: This only includes bound video elements but if we simplify what the API provides then we can include media loaded into asbplayer webapp instances as well.

common/web-socket-client/web-socket-client.ts
mcgrizzz: Sure, we definitely can clean it up.

### Title and faviconUrl
The only specific use I'm thinking of is the ability to provide a user-facing selector in whatever client software. In that case `title` and maybe `faviconUrl` are very helpful to have. Otherwise, there's no human-readable way to discern the different bound media.

### `active` vs `focused`
I noticed you kept `focused`, but not `active` or `discarded`. If we only keep one of these, I wonder if `active` is a better choice.

Sometimes the `focused` API via the query API was a bit unpredictable. `focused` seems to apply to a focused window only when there is more than one window. It reports false when there's only one window, at least in my testing.

### Bound media ID
For `id`, I do not see an existing ID that works cleanly across both streaming and local media. So we'd probably have to generate this, maybe something along the lines of:

```ts
hash("streaming:" + tabId + ":" + src)
hash("local:" + asbplayerId)
```

### Loaded subtitles shape
As for `loadedSubtitles: SubtitleTrack[];`,

Maybe this is just a map of the loaded tracks and then the `get-subtitles` endpoint can just request exactly what it wants via `bound media ID` and `track number`. In this case, empty array response means no subtitles loaded. 

```json
[
  {
    "trackNumber": 1,
    "fileName": "ja.srt"
  },
  {
    "trackNumber": 3,
    "fileName": "en.srt"
  }
]
```

### Overall shape
So final structure would look something like this:

```ts
export interface SubtitleTrack {
  trackNumber: number;
  fileName: string;
}

export interface BoundMedia {
  id: string;
  type: "streaming" | "local";
  title?: string;
  faviconUrl?: string;
  loadedSubtitles: SubtitleTrack[];
  active: boolean;
}
```

extension/src/services/web-socket-client-binding.ts
mcgrizzz: Oversight on my part. I can definitely update this with webapp instances too once the simpler response shape is finalized.

common/web-socket-client/web-socket-client.ts
killergerbah: > The only specific use I'm thinking of is the ability to provide a user-facing selector in whatever client software. In that case title and maybe faviconUrl are very helpful to have. Otherwise, there's no human-readable way to discern the different bound media.

Sounds good

> I noticed you kept focused, but not active or discarded. If we only keep one of these, I wonder if active is a better choice.
Sometimes the focused API via the query API was a bit unpredictable. focused seems to apply to a focused window only when there is more than one window. It reports false when there's only one window, at least in my testing.

Also sounds good

> For id, I do not see an existing ID that works cleanly across both streaming and local media. So we'd probably have to generate this, maybe something along the lines of:
```
hash("streaming:" + tabId + ":" + src)
hash("local:" + asbplayerId)
```

Yes that would make unique IDs but I don't see what problem that solves other than making a globally unique single ID.  Any API behavior that uses the ID could also use a [src/asbplayer ID, type] pair. Also, within asbplayer code we are already using `mediaId` which refers to the asbplayer instance ID for local video or the video `src` string for streaming video.

mcgrizzz: > Yes that would make unique IDs but I don't see what problem that solves other than making a globally unique single ID. Any API behavior that uses the ID could also use a [src/asbplayer ID, type] pair. Also, within asbplayer code we are already using mediaId which refers to the asbplayer instance ID for local video or the video src string for streaming video.

Ah gotcha, I didn't realize `mediaId` was already serving that role internally.

I'll drop the hashed ID and use the existing `mediaId` instead. 

So the final shape I proposed should be the same. It's just that the id will be coming from `mediaId`. If you want id to explicitly be called 'mediaId', I can do that as well.

```ts
export interface SubtitleTrack {
  trackNumber: number;
  fileName: string;
}

export interface BoundMedia {
  id: string;
  type: "streaming" | "local";
  title?: string;
  faviconUrl?: string;
  loadedSubtitles: SubtitleTrack[];
  active: boolean;
}
```

killergerbah: Hey sorry so I thought about it and the original solution you proposed is better because only the single unique ID would be required to reference the same media in other APIs. 

So this looks good to me:

```
export interface SubtitleTrack {
  trackNumber: number;
  fileName: string;
}

export interface BoundMedia {
  // hash("streaming:" + tabId + ":" + src)
  // hash("local:" + asbplayerId)
  id: string;
  type: "streaming" | "local";
  title?: string;
  faviconUrl?: string;
  loadedSubtitles: SubtitleTrack[];
  active: boolean;
}
```

or whatever you can think of to make the ID unique. That way when `get-subtitles` or similar is implemented it can just be something like `get-subtitles?id=xyz&trackNumber=abc`

mcgrizzz:
A couple of notes about this implementation...

#### Modified asbplayer interface

A list of subtitle tracks is now included within `_updateAsbplayers` and the `Asbplayer` interface within `tab-registry.ts`. They were also added to `VideoElement`. They're transferred via heartbeats. This makes them available to `onGetBoundMedia`.

#### WebSocket client asymmetry

As of right now, only the extension WebSocket client implements `onGetBoundMedia` since it has access to all the bound media including apps. I also added an `if (this.onGetBoundMedia !== undefined) {` check within the common client so clients that do not implement this handler do not respond.

The main issue is that `get-bound-media` is broadcast to connected WebSocket clients, but not every client has authoritative state. When I had both the extension and app open and connected, the app could respond first with an empty response before the extension had a chance to respond with the actual bound media list.

One option would be to implement `onGetBoundMedia` on the app side and guard it with `extension.installed`, so the app only responds when there is no extension. However, that still leaves an ambiguity when multiple app pages are open, since more than one app client could respond.
 
The other option would be to allow multiple responses and advise WebSocket servers(other apps) to combine however many responses they receive. I don't think that is a great 'solution'and just pushes the aggregation problem onto every consumer.

#### cyrb53 hashing

For the IDs, I decided on `cyrb53`. It's supposed to be fast and collision resistant. I also wrapped it in a `boundMediaId()` wrapper function so the underlying hash function can be swapped without refactoring anything else. Right now this is in `web-socket-client-binding` because that's the only place it's being used, but if you have a better place for it, let me know.

#### `get-subtitles` and other implementation

With this groundwork set, `get-subtitles` should be pretty trivial to get implemented. Then we can also update the `seek-timestamp`, `load-subtitles`, and even `mine-subtitle` endpoints to allow optional targets.

docs/docs/reference/external-api.md
```diff
@@ -103,12 +103,64 @@ asbplayer, as a WebSocket client, responds to the following commands from a WebS
...
+
+#### Response
+
+```javascript
+{
+    "command": "response",
+    // Same message ID received in request
+    "messageId": "9f1c2b3a-4d5e-6f70-8190-a1b2c3d4e5f6",
+    "body": {
+        "media": [{
```
killergerbah, line 129: Looks like you'll need to catch up the docs

killergerbah:
Thanks this looks pretty good. Just one comment to catch up the docs.

>The main issue is that get-bound-media is broadcast to connected WebSocket clients, but not every client has authoritative state. When I had both the extension and app open and connected, the app could respond first with an empty response before the extension had a chance to respond with the actual bound media list.

Yeah system design is a little weird and maybe ill-thought-out. For most consumers of the API asbplayer acts like a server but it's really a WebSocket client. The extension provides a convenient way to designate a "master" client even if there are several app clients connected at the same time. Because of this, I don't think any of the existing WebSocket messages behave well with multiple webapp clients.

>One option would be to implement onGetBoundMedia on the app side and guard it with extension.installed, so the app only responds when there is no extension. However, that still leaves an ambiguity when multiple app pages are open, since more than one app client could respond.

Yeah that is actually a reasonable solution. There may be better ones though so we can defer this for now. One thing that came to mind is using session storage as a way to coordinate or collect multiple app instance states.

>The other option would be to allow multiple responses and advise WebSocket servers(other apps) to combine however many responses they receive. I don't think that is a great 'solution'and just pushes the aggregation problem onto every consumer.

Agreed not a great solution but it would probably work most of the time and I don't foresee a reason for anyone to build new WebSocket server implementations aside from the Anki plugin.

mcgrizzz:
Sounds good, I'll defer that one issue for now and catch up the docs.

killergerbah (approved):
Awesome, thanks, appreciate all the work and thought you're putting into the API

PR #1034, opened 2026-05-31
add pitch accent and granular annotation triggers

ShanaryS:
# Support Pitch Accent
Currently, Yomitan doesn't expose anything for tones or gender so this is the only accent annotation we can support for now. I think accent annotation is the least important of any annotation so it's not a big deal.

image

* There are lots of pitch accent styles, but for subtitles the only one that makes sense is to draw around each mora of the subtitles.
	* This is completely handled by css classes and their definitions in `subtitles.css` and `video.css`
* This feature is pretty coupled with `reading`, they even use the same options for the settings. The pitch accent will on the fuirigana (so it must be enabled for that word, e.g. status) or directly on the kana tokens.
	* I don't think there is much need to separate these since you cannot ask for pitch accent without also getting the reading
	* If the reading is being applied to the furigana, then we will remove the per tokenPart aspect of the reading and apply it to the entire token, including with the kana get their reading as well. e.g. 飛び切り readings would be `と き ` so make it contiguous as `とびきり` so connecting and reading pitch is easier (as you can see in the image above)
* When pitch accent is being displayed, the lines will take on the status color or black when its not enabled (see image above)
  * We will not color the token using the usual styling in this case to prevent clashing with other styles and reducing clutter.
* We will use the explicit pitch for attachment particles if provided or generally maintain the previous pitch if not explicit
  * We determine a attached particle if there was pitch accent data for the previous token and the current is a single kana token.
* The pitch accent data is only available via `/termEntries`, I would need a PR on Yomitan to add it to the `/tokenize` results like the `frequency`.
  * This only affects `scanning-parser`, which shouldn't really be used for Japanese as `MeCab` is far superior. So it may not be worth going for. We still get it through `bulkTermEntries` but `scanning-parser` is twice as slow with pitch accent enabled since we need to replicate the work done in Yomitan with the call.

# Granular Annotation Hover Settings
image

* I believe pitch accent should not be displayed all the time, it makes the subtitles too cluttered. So it was time to make the annotation hover more granular to enable/disable each annotation.
  * We now have support for hover semantics on the `subtitlePlayer`. Since the hover settings weren't granular before, it only applied to the video as that's where it matters most and users still had access to full annotations without hovering.
  * Since we are now making it granular, this includes separate toggles for `video` (extension/VideoPlayer) and for `subtitlePlayer` (sidePanel/App/Stats). This is because users may want full annotations for `subtitlePlayer` but only show a few on hover for `video` or any other combination.
  * We default to pitch accent being hover only for `video` and `subtitlePlayer`. For users with the current global annotation hover toggle, it will be migrated to just `color`, `reading`, and `frequency` for `video` only
* This means that each context render their own subtitles, which is not very different from the current behavior but it's now very explicit
* All the hover settings (including highlight on hover) has now been moved to the Styling section (renamed to `Styling and Hover`)
  * Since these settings are granular, they are affected by the previews for annotations. Every display setting is now represented in the previews.

# Full Annotation Customization by Status & States
image

* Rather than restricting the status combination to trigger annotation, we will now allow arbitrary selection of the status (which has been requested)
* Previously only readings had a setting to control how ignored states are handled but now it's available to all annotations.
* Contiguous statuses of 3 or more will be concatenated using `→` to save space

# Add Customizable Annotation Sizes
image

* Since we now have the structure in place to target `video` and `subtitlePlayer` annotation contexts, it makes sense to expose these settings to the user (which has been requested).
* I refactored these settings to now be more explicit with a toggle between the targets which also affects the previews explicitly rather than the unintuitive previous behavior.
* Only the necessary settings that is reasonable to differ between targets are independent, so `Highlight words on hover` and the behavior of annotation (status/states triggers, strategies, etc) should always be global.

Assisted-by: Copilot:GPT-5.5


common/app/components/Player.tsx
```diff
@@ -542,16 +560,17 @@ const Player = React.memo(function Player({
             subtitleCollectionOptions,
             mediaId,
             (updatedSubtitles, dictionaryTracks) => {
-                renderRichTextOntoSubtitles(updatedSubtitles, dictionaryTracks);
-                channel?.subtitlesUpdated(updatedSubtitles);
+                const playerSubtitles = subtitlesForPlayer(updatedSubtitles, dictionaryTracks);
+                if (channel) channel.subtitlesUpdated(updatedSubtitles);
```
killergerbah, line 564: Are we sending the subtitles without rich text over the channel because we prefer for the rich text to be recalculated on the other side?

common/app/components/subtitles.css
```diff
@@ -17,17 +17,67 @@
 }
 
 .asb-reading rt {
-    font-size: 0.5em !important;
+    font-size: var(--asb-reading-size) !important;
```
killergerbah, line 20: I see that `tokenAnnotationStyleValues` returns an empty object in case it's not defined, could this result in an undesirable `font-size`?

common/settings/settings.ts
```diff
@@ -245,37 +396,91 @@ export enum TokenFrequencyAnnotation {
...
-        dt.dictionaryColorizeSubtitles ||
-        dt.dictionaryAutoGenerateStatistics ||
-        dt.dictionaryTokenReadingAnnotation === TokenReadingAnnotation.LEARNING_OR_BELOW ||
-        dt.dictionaryTokenReadingAnnotation === TokenReadingAnnotation.UNKNOWN_OR_BELOW ||
-        dt.dictionaryTokenFrequencyAnnotation === TokenFrequencyAnnotation.UNCOLLECTED_ONLY
-    );
+export function dictionaryStatusCollectionEnabled(dt: DictionaryTrack, options: { includeStates: boolean }): boolean {
+    const { includeStates } = options;
+    if (dt.dictionaryTokenAnnotationConfig.colorizeEnabled || dt.dictionaryAutoGenerateStatistics) return true;
+    for (const annotation of ['reading', 'frequency', 'pitchAccent'] as (keyof TokenAnnotationTriggerOptions)[]) {
```
killergerbah: Would it make sense to declare this array closer to the `TokenAnnotationTriggerOptions` type so that it's easier to keep it consistent if you add another option?

common/subtitle-annotations/subtitle-annotations.ts
```diff
@@ -1719,18 +1746,23 @@ export class SubtitleAnnotations extends SubtitleCollection<RichSubtitleModel> {
...
             token.frequency = await ts.yt.frequency(trimmedToken);
         } else {
             this.refreshCache.add(index);
         }
     }
 
+    private async _updatePitchAccent(token: Token, trimmedToken: string, index: number, ts: TrackState): Promise<void> {
+        if (!ts.yt) throw new Error('Yomitan uninitialized - cannot update token pitch accent');
+        if (token.status == null || !shouldUseAnnotation('pitchAccent', token.status, token.states, ts.dt)) return;
+        if ((this.initialized && !this.generateStatisticsRequested) || ts.yt.getSupportsBulkPitchAccent()) {
```
killergerbah, line 1759: Does this mean that if stats have been requested, pitch accent updates are delayed?

common/app/components/Player.tsx
killergerbah: Nvm I see why this is required now, due to the video/subtitle list specific settings

common/app/components/subtitles.css
ShanaryS: richText cannot exist if the dictionaryTrack isn't available, so it's not currently possible for this to happen. But there is no harm to having a default defined.

common/settings/settings.ts
ShanaryS: We need to update both `dictionaryTrackEnabled()` and `dictionaryTrackStatusCollection()` enabled with any new types so it's unlikely to be missed. But I think we can share an array for both of them with a slight refactor. Adding new annotations will always require nuance throughout the code though.

common/subtitle-annotations/subtitle-annotations.ts
ShanaryS: Yes. Tokenize and lemmatize are required to do anything. Readings come along with tokenize so it's never a concern.

Frequency was the first optional annotation and so I added code to lazy load it without blocking the annotations. But with stats we need frequency so we have to wait for it for stats to be considered complete. This is not a big deal though since Yomitan now has frequency from tokenize and bulk `/termEntries` in the case or mecab. So this comes for free from tokenize as well.

Pitch accent is completely visual and Yomitan has not merged my changes to include it in tokenize so non-mecab users will have annotations twice as slow (mecab already needs `/termEntries` which has everything). I opted to lazy load it like how frequency is (again no longer relevant) but also defer it if stats are loading since the stats overlay is far more important.

This check is irrelevant for mecab users as the supports check returns true for it. But we have to see if Yomitan accepts my PR (probably will, just exposing the data with a single line change) for scanning-parser support.

common/components/DictionarySettingsTab.tsx
```diff
@@ -915,201 +1177,266 @@ const DictionarySettingsTab: React.FC<Props> = ({
...
-                    <FormLabel component="legend">{t('settings.dictionaryTokenReadingAnnotation')}</FormLabel>
-                    <RadioGroup row={false}>
-                        <LabelWithHoverEffect
-                            control={
-                                <Radio
-                                    checked={
-                                        selectedDictionary.dictionaryTokenReadingAnnotation ===
-                                        TokenReadingAnnotation.ALWAYS
+                {supportsDictionaryTokenAnnotationConfig &&
+                    tokenAnnotationTriggerOptions.map(({ annotation, labelKey }) => {
```
killergerbah, line 1223: Now that there are a lot more display related settings I wonder if they can be combined in an elegant way with the previews at the bottom. For these options in particular, there's a new toggle per token status. Then the remaining problem is how to deal with ignored tokens.

common/components/DictionarySettingsTab.tsx
```diff
@@ -1827,6 +2154,98 @@ const DictionarySettingsTab: React.FC<Props> = ({
...
+                                labelPlacement="start"
+                            />
+                        </Stack>
+                        <RadioGroup
+                            row
+                            value={tokenAnnotationTarget}
+                            onChange={(e) => setTokenAnnotationTarget(e.target.value as TokenAnnotationConfigTarget)}
+                            sx={{ justifyContent: 'center' }}
+                        >
+                            {tokenAnnotationTargets.map(({ target, labelKey }) => (
```
killergerbah, line 2228: I'm assuming you just threw something together as a placeholder, but mentioning anyway that we'll want to reorganize this UI before release

ShanaryS: Yeah, as long as it's an explicit toggle between video and subtitlePlayer it should be fine. I think this makes it the best for displaying previews. This also makes it tricky to simplify it with the settings that enable the annotations per status/state since those are global.

common/components/DictionarySettingsTab.tsx
ShanaryS: The previews will reflect what annotation is enabled for each status. Enabling for status doesn't really need a preview imo, it's more so about what it means to show annotation for each status rather than how it will look. Also, those settings under that heading are all that controls whether annotations are enabled and so are at the top.

So I don't think there is a need to combine them as users won't be tinkering with the preview based on enabled status unlike the colors, hover, and sizes. I'll add a separate preview for ignored behavior though as it makes sense to do.

docs/docs/guides/annotation.md
```diff
@@ -10,6 +10,7 @@ asbplayer can annotate subtitles to better assist with language learning. Annota
     - known status can be sourced from and synced with Anki, WaniKani, and/or tracked locally in asbplayer (includes import/export features to help seed known words)
 - reading annotation (readings displayed above each word or based on status)
 - frequency annotation (rank-based frequency displayed below each word or based on status)
+- pitch accent annotation (accent patterns displayed on furigana or kana words)
```
killergerbah: Is there any additional setup required for pitch accent to display? I guess it should be obvious that users need to install a pitch accent dictionary but it wouldn't hurt to make it explicit IMO

killergerbah:
I started testing this on my computer and the mora classes don't seem to be applied correctly:

image

Same issue happens on streaming video:

image

It looks like the low and high class definitions aren't being picked up by my browser:

image


Also, on an 80 minute subtitle file the tab is getting out of memory killed. Attached the file in case you can reproduce it.

[Oshi.no.Ko.S01E01.1080p.NF.WEB-DL.DDP2.0.H.264-ZigZag.txt](https://github.com/user-attachments/files/28926007/Oshi.no.Ko.S01E01.1080p.NF.WEB-DL.DDP2.0.H.264-ZigZag.txt)

ShanaryS:
The weird display is likely because the pitch accent size is too large. I made a mistake setting the default to 1 instead of 0.1.

killergerbah:
Nice, can confirm the .1em fixed things for me

General comment on the feature - I know you're intentionally giving the pitch accent styles the same colors to reduce clutter but I wonder if there's a way to make the pitch accent styles more distinguishable from the token status styles. The main reason being that learners who are likely to benefit from pitch accent annotations also the most likely to confuse token status underlines for a lower pitch. In any case, I think the default underline style for token status is confusing when combined with pitch accent. Maybe at least the default settings could make them more distinguishable, after that it's up to the user if they want to use lines everywhere.

ShanaryS:
> I think the default underline style for token status is confusing when combined with pitch accent

Would adding a warning below the pitch accent dropdown and at the status styling be sufficient when underline/overline is being used with pitch accent? I think underline is the best default and a lot of people probably won't use pitch accent (especially since it's only for Japanese). Also since pitch accent defaults to on hover it should mitigate this somewhat, but I agree it's an issue for kana.

> Also, on an 80 minute subtitle file the tab is getting out of memory killed. Attached the file in case you can reproduce it.

I can't reproduce the crash but I do see the issues since it's struggling to complete. I did a bunch of testing and memory profiling and the issue solely lies with the rendering of the rich text, the runtime cache for annotations or yomitan is negligible. This can be confirmed by putting a return at the top `renderRichTextOntoSubtitles` which hopefully allows it to load for you.

The next steps is to figure out how to best handle rendering. We probably would have needed to do this eventually. But something like only rendering for subtitles in the next 5s and clearing the past ones should fix it for video. For SubtitlePlayer perhaps we can render based on currently visible subtitles. I know there is some caching being done with `OffscreenDomCache` so that's probably where this would live. I'll spend some time working on it but I think any ideas you have around this would be helpful.

common/src/model.ts
```diff
@@ -1,6 +1,7 @@
 import type { AnkiSettings, TokenState, TokenStatus } from '../settings/settings';
 import type { OnlineSubtitleSourceConfig } from '../global-state';
 import type { TokenStatusInfo } from '../dictionary-db';
+import { PitchAccentPosition } from '../yomitan/yomitan';
```
killergerbah: Can this be a type import?

killergerbah:
> I can't reproduce the crash but I do see the issues since it's struggling to complete. I did a bunch of testing and memory profiling and the issue solely lies with the rendering of the rich text, the runtime cache for annotations or yomitan is negligible. This can be confirmed by putting a return at the top renderRichTextOntoSubtitles which hopefully allows it to load for you.

> The next steps is to figure out how to best handle rendering. We probably would have needed to do this eventually. But something like only rendering for subtitles in the next 5s and clearing the past ones should fix it for video. For SubtitlePlayer perhaps we can render based on currently visible subtitles. I know there is some caching being done with OffscreenDomCache so that's probably where this would live. I'll spend some time working on it but I think any ideas you have around this would be helpful.

My first reaction was that maybe it had to do with the amount of token data being processed but I have observed in the past that DOM by itself takes up a lot of memory as well, so your explanation makes sense to me.

Unless your measurements indicate otherwise I would be surprised if streaming video is having similar issues because all of the subtitles are rendered offscreen, so I'll just comment on SubtitlePlayer for now.

The main problem I've encountered with partial rendering/virtual table is scrollbar calculation when each row in the list can have variable height. The contents of each row is essentially arbitrary - it can contain images or text with arbitrary styling so I think the only reliable way to measure the height of each row is to actually render it. Perhaps this is something that the dom cache can be used for.

Also, if DOM is the main issue then maybe I agree this can be deferred to a different diff, since it's not a specific problem with this change. I also agree we should try to solve the problem sooner than later since annotations seem to be making the subtitle DOM significantly bigger.

I am extra busy lately but I might be able to put in some hours on the weekend to help out as well. My first thought is to re-attempt integrating a virtual table library.

killergerbah:
>Would adding a warning below the pitch accent dropdown and at the status styling be sufficient when underline/overline is being used with pitch accent? I think underline is the best default and a lot of people probably won't use pitch accent (especially since it's only for Japanese). Also since pitch accent defaults to on hover it should mitigate this somewhat, but I agree it's an issue for kana.

I forgot to reply to this, yeah i agree a warning is sufficient, as long as the issue can be conveyed in a few words

ShanaryS:
> Unless your measurements indicate otherwise I would be surprised if streaming video is having similar issues because all of the subtitles are rendered offscreen, so I'll just comment on SubtitlePlayer for now.

I didn't test it for video only the SubtitlePlayer so it may not be an issue. But video should be relatively trivial to at least know it's not visible. But we can take it one at a time.

> My first thought is to re-attempt integrating a virtual table library.

I was messing around with reducing the preemptive caching for the rich text and it seems to improve the performance. It has a bunch of bugs but virtualization may not be necessary.

killergerbah (commented):
I just did another skim through and nothing jumped out
I'll approve after doing another pass tomorrow morning

killergerbah (approved):
Feel free to merge when ready
As mentioned before we should clean up the settings UI before release

PR #1038, opened 2026-06-04
Support ankiConnect API Key

### LINKED ISSUE #1032: Anki Connect authentication support
Anki Connect has optional [authentication support](https://git.sr.ht/~foosoft/anki-connect#authentication), which I'd like to use for improved security, but asbplayer doesn't seem to support this.

Image

I'd like to be able to set the Anki Connect API key.

Perhaps adding another field adds too much clutter to the UI (it might be confusing to the majority of users who don't have authentication enabled). So it might be worthwhile to make it very clear that the API key is optional, or surface this if and only if we hit the "valid api key must be provided" server error that happens when we try to connect to an Anki Connect server that requires authentication.

I thought perhaps it might be possible to pass the key via the URL (as a query parameter), but Codex informs me it needs to be the `"key"` field in the POST JSON, which asbplayer currently seems to have no support for.

ShanaryS:
fixes #1032

The setting will only show if there is already an api key or if the response says one is required. If the user were to disable the api key in the future, they will need to clear the input (which isn't clear as the same error message is present). We can't tell if this is due to the api key not being required or being different (without requesting without as a test) so I didn't do anything special for this.


common/anki/anki.ts
```diff
@@ -459,6 +463,14 @@ export class Anki {
         return response.result;
     }
 
+    static requiresApiKey(result: any): boolean {
+        if (result === null || result === undefined) return false;
+        if (result instanceof Error) return Anki.requiresApiKey(result.message);
+        if (typeof result === 'string') return result.toLowerCase().includes('valid api key must be provided');
+        if (typeof result !== 'object') return false;
+        return result.requireApikey === true || result.requireApiKey === true || Anki.requiresApiKey(result.error);
```
killergerbah, line 471: `result.requireApikey === true` appears twice?

ShanaryS: ankiConnect actually returns `requireApikey` so I just added the caps to be safe.

killergerbah: Oh ok I didn't see the difference in casing

common/components/AnkiSettingsTab.tsx
```diff
@@ -27,6 +30,7 @@ import { Anki, exportCard } from '../anki';
 import Stack from '@mui/material/Stack';
 
 const defaultDeckName = 'Sentences';
+const maskedAnkiConnectApiKey = '●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●●';
```
killergerbah: I think most ppl will copy and paste but I wonder if this masking can take into account the length of the input, at least while the input is in focus? Otherwise the UX feels a bit explosive. Also if you make this change here, it might make sense to make it re-usable with the WaniKani field as well

https://github.com/user-attachments/assets/ad0013c0-abe8-4e04-a429-5dad4b6ba99d

killergerbah:
Thanks, I tried the new change
- Now when the API key field is in hidden mode, it doesn't respond to keystrokes at all. Which may be reasonable behavior but it's not intuitive
- The circles are super huge but this is just a cosmetic issue

ShanaryS:
> Now when the API key field is in hidden mode, it doesn't respond to keystrokes at all. Which may be reasonable behavior but it's not intuitive

These values shouldn't really be changed after initial setup so I made it uneditable. If it is empty it defaults visible so it should be easy to interact (most likely copy paste) and then never touched again. I wanted to avoid writing logic to constantly masking as typing occurs.

> The circles are super huge but this is just a cosmetic issue

It's probably because of the monospace requirement I added, this should make the hidden and visible the same length. On my browser it didn't change but if it's too distracting then we can use something else.

killergerbah:
> These values shouldn't really be changed after initial setup so I made it uneditable. If it is empty it defaults visible so it should be easy to interact (most likely copy paste) and then never touched again. I wanted to avoid writing logic to constantly masking as typing occurs.

That makes sense, but now if you want to change your API key you need to unhide it, paste, then hide again, which weakens the security 
I guess most users won't change their API key but it's' not an unlikely use case, even if you assume everyone will just copy paste

> It's probably because of the monospace requirement I added, this should make the hidden and visible the same length. On my browser it didn't change but if it's too distracting then we can use something else.

Can we use • instead? On my computer it looks a bit better

ShanaryS:
> That makes sense, but now if you want to change your API key you need to unhide it, paste, then hide again, which weakens the security

The main thing I want to ensure by masking is that users must opt-in to display sensitive data. So if they are sharing their screen or a screenshot, their is no risk of leaking anything.

It seems in order to mask as we go, we would need to use the `password` field, which will cause the browser to prompt to save the token. In order to do it in the text field it may be possible but I don't see how to do it, at least not in a simple way that guarantees accuracy. I think we have to hook into the key events which seems error-prone.

killergerbah:
Hmm as a compromise would it make sense to disable the input when it's in hidden mode? If we do this then at least it'll be more obvious that it's not supposed to respond to anything

ShanaryS:
I also fixed an issue where `SubtitleController` would only render the external readings if annotations were enabled. This was reported in discord earlier today.

killergerbah (approved):

killergerbah:
Sounds good merging so this can be released soon

PR #1039, opened 2026-06-04
Make stats header sticky

### LINKED ISSUE #979: Make statistics header on sidebar sticky
Making the header sticky when scrolling would be a slight UX improvement and would maintain UI consistency with the `Mining History` tab, which already uses a sticky header.

Image

ShanaryS:
fixes #979

### DIFF
--- a/common/components/StatisticsDrawer.tsx
+++ b/common/components/StatisticsDrawer.tsx
@@ -8,18 +8,10 @@ interface Props extends StatisticsProps {
     onClose: () => void;
 }
 
-const StatisticsDrawer: React.FC<Props> = ({
-    open,
-    showBackButton,
-    drawerWidth,
-    onClose,
-    sx,
-    contentPadding = 2,
-    ...statisticsProps
-}) => {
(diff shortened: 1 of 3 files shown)


common/components/StatisticsDrawer.tsx
```diff
@@ -8,10 +8,18 @@ interface Props extends StatisticsProps {
...
 }
 
-const StatisticsDrawer: React.FC<Props> = ({ open, showBackButton, drawerWidth, onClose, sx, ...statisticsProps }) => {
+const StatisticsDrawer: React.FC<Props> = ({
+    open,
+    showBackButton,
+    drawerWidth,
+    onClose,
+    sx,
+    contentPadding = 2,
```
killergerbah: Can some padding be added to the separate popup window version as well?

ShanaryS: I added margin to it but removed it from the extension popup and stats drawers. This better matches what it was like before and the non-popout contexts really need the space.

killergerbah:
Sweet

Issue #1009, opened 2026-05-18
Feature Req: Cache recent search queries and loaded subtitles for online subtitle search


xwxb:
**Is your feature request related to a problem? Please describe.**

I'd like to propose two small enhancements to the online subtitle search feature (which I previously contributed). Currently, when I refresh the page, all recent search queries and previously loaded subtitle files are lost. This is particularly frustrating on websites where video compatibility isn't great — page refreshes and subtitle re-embedding happen often. Video titles on these sites are usually in Chinese, making them unsuitable for searching directly, so being able to re-select a previous successful search would be much more convenient. When subtitle embedding breaks, quickly restoring the most recently loaded subtitle would also be very helpful.

**Describe the solution you'd like**

1. **Cache recent search queries**: Persist recent subtitle search queries so they survive page refreshes, allowing users to quickly re-select a previous search.
2. **Cache recently loaded subtitle files**: Persist recently loaded subtitle files so users can quickly restore them after a page refresh or re-embedding failure.

**Describe alternatives you've considered**

For storage, `localStorage` seems like the most natural fit, but I'm not sure if there are potential issues (e.g., size limits for subtitle file content). I had previously considered introducing a general-purpose "misc" IndexedDB store (which also came up with the file recovery feature), but opening a new IndexedDB just for this might be overkill.

**Additional context**

Before I start implementing, I'd like to discuss:

1. **Is this too niche?** This need stems from my own usage patterns (non-English video sites with compatibility issues). If it's considered too personal, I totally understand.
2. **Storage approach**: If the feature is welcome, I'd appreciate input on the preferred storage strategy within the current architecture.

Happy to implement this myself once the direction is agreed upon.

xwxb:
I've noticed that the browser already handles the first search history, and the second isn't very frequent as subtitles for an episode are usually a one-time thing. Perhaps caching the entire search result list would be a more suitable requirement. I'll reopen this once I've thought it through further. https://github.com/asbplayer/asbplayer/issues/1009

PR #1040, opened 2026-06-05
feat: cache Jimaku subtitle works

xwxb:
## Background
This PR addresses the pain point in #1009 regarding the Jimaku subtitle search workflow. Currently, when watching a series, users often have to manually search for each episode. Even with title auto-filling, it can be inaccurate or require several repetitive steps for every new episode.

This feature introduces a persistent cache for recently accessed Jimaku works. It is intended as a pure UX improvement that allows users to:
1. Instantly see previously used subtitle lists for a series without re-searching.
2. Speed up the workflow for binge-watching series where the automatic matching might not be perfect.

## Summary
- Persist recently opened Jimaku works with cached subtitle file lists
- Show cached files immediately while refreshing in the background
- Keep Jimaku loading state scoped to the currently selected work and complete locale keys

image

## Tests
- `corepack yarn workspace root run verify`

I'm open to feedback on the implementation or the feature itself. If there are concerns about the approach, I'm happy to refine it or split the changes.

### DIFF
--- a/common/global-state/index.ts
+++ b/common/global-state/index.ts
@@ -9,9 +9,15 @@ export enum AnnotationTutorialState {
     hasSeen = 2,
 }
 
+export interface JimakuCachedWork {
+    id: number;
+    name: string;
+}
+
 export interface OnlineSubtitleSourceConfig {
     jimakuApiKey: string;
     jimakuSearchCategory: 'anime' | 'drama';
+    jimakuRecentWorks?: JimakuCachedWork[];
 }
 
 export const initialGlobalState: GlobalState = {
@@ -21,6 +27,7 @@ export const initialGlobalState: GlobalState = {
     onlineSubtitleSourceConfig: {
         jimakuApiKey: '',
         jimakuSearchCategory: 'anime',
+        jimakuRecentWorks: [],
     },
 };
 
--- a/common/locales/de.json
+++ b/common/locales/de.json
@@ -711,6 +711,7 @@
         "jimakuApiKey": "Jimaku API key",
         "jimakuApiKeyAutosaveHint": "Saved automatically. Get an API key <0>here</0>.",
         "entries": "Search results",
+        "recentEntries": "Recent",
         "noEntries": "No search results",
         "availableFiles": "Files",
         "noFiles": "No files",
--- a/common/locales/en.json
+++ b/common/locales/en.json
@@ -672,7 +672,7 @@
             "noDeckBreakdown": "No matching Anki cards were found for this media.",
             "info": {
                 "ankiStatistics": "How your Anki cards match up with the words in this media.",
-                "projectedCards": "Preview your stats by treating past cards or new cards (in the order they will be introduced) as Mature." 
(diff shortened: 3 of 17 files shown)


killergerbah:
How is this feature is intended to be used? I quickly checked this out and it didn't seem to be obviously working based on your description. I haven't read the code yet.

xwxb:
@killergerbah  Sorry for the rush earlier. I've added screenshots and detailed background to the PR description and re-verified that the functional tests are passing.

The main goal is to improve the workflow for watching series where searching for subtitles for each episode is repetitive, especially when the auto-fill title isn't perfectly accurate. This cache allows picking up from where you left off immediately. 

If you have any concerns about the feature design or would prefer a different implementation approach, feel free to let me know—I'm happy to adjust it.

killergerbah:
Thanks @xwxb for the explanation. I like the feature, it's a good idea and users probably waste a lot of time re-searching for the thing they just watched.

General comments on the feature and implementation design:
- I think it's only necessary to cache the top level works, not the file URLs underneath. The limit on extension storage is undocumented and if it's not necessary I don't see a need to store potentially 100s of URLs there. I think the effect on UX would be tolerable as the user would just need to wait for the duration of the automatic search query.
- The "Recent works" label is very specific to Jimaku terminology, I think most users would instantly and intuitively understand the meaning of the label if it were just "Recent"
- If you stop caching the URLs above I think you can also remove the "xyz cached" label as well.

xwxb:
@killergerbah That's a fair point. I did overlooked the possibility of extremely long file lists for certain series.

To be honest, I might have actually considered the storage issue earlier, but I was sidetracked by trying to optimize the logic from the previous PR ([7df2978](https://github.com/asbplayer/asbplayer/pull/1004/commits/7df297893befa3bc82b5f1b842e15d99bec98c87)). I felt that implementation was a bit heavy, but I eventually concluded there wasn't a better way to refactor it without losing its purpose, so I stuck with a similar approach here.

An additional reason that influenced my design is my personal use case: I often try multiple subtitles to find a perfect sync. This specific scenario magnified my focus on eliminating even that split-second network delay. However, I agree that it’s a tolerable trade-off and caching everything for long series is indeed overkill.

I'll refactor this to only cache the works and remove the "cached" labels as suggested.

xwxb:
Actually, I'm now starting to doubt whether this implementation is still worth it. If it's downgraded to caching search terms, it might not be significantly different from the search results the browser already auto-caches.

I'll put this on hold for a while to reconsider how to achieve a better balance between convenience and performance. In the meantime, I reopened #1009.

killergerbah:
I do think the "recent" list is a nice UX improvement regardless of the implementation details

xwxb:
@killergerbah Exactly, I've recently come to the same conclusion. There really isn't a more convenient and efficient UX solution for access. I'll be moving forward with the original development route soon.

xwxb:
@killergerbah done for the simplify, PTAL

killergerbah:
Thanks @xwxb

PR #1041, opened 2026-06-05
Add local video scrub timestamp previews

NirDafnai:
Assisted-by: codex:gpt-5.5

﻿## Summary
- Add a configurable timestamp preview for the local video scrub bar, defaulted on.
- Respect the existing Subtitle above thumbnail layering setting for timestamp previews too.
- Add small progress bar helpers/tests plus settings schema and locale entries.

image
image
image


killergerbah:
@NirDafnai Thanks, this looks really good
Can you remove the setting? I think it's fine to be enabled by default
Later, I think it would be cool and useful to language learners for the subtitle to display as well

NirDafnai:
> @NirDafnai Thanks, this looks really good Can you remove the setting? I think it's fine to be enabled by default Later, I think it would be cool and useful to language learners for the subtitle to display as well

@killergerbah Sure, removed the setting, and I agree!

killergerbah (approved):
Awesome, thanks

PR #1044, opened 2026-06-06
Add "always use subtitle text for sentence field" mining setting

cvivesa:
## Summary

Adds an opt-in mining setting — **"Always use subtitle text for sentence field"** — that forces asbplayer to source the sentence field from the loaded subtitle, ignoring any externally supplied sentence text. Defaults to **off**, so existing behavior is unchanged.

## Motivation

This primarily targets the **one-click mining** setup (Yomitan → AnkiConnect proxy → asbplayer "update last card"), which is where the problem mostly shows up.

Today, when an external tool supplies the sentence, asbplayer keeps that text unless it's empty or a substring of the current subtitle. In the one-click flow this means the resulting card keeps Yomitan's sentence — which can include page "garbage" or extra markup — instead of the clean subtitle text. Mining the same card manually from asbplayer instead uses the subtitle, so the two paths are inconsistent. This setting makes both paths use the subtitle text.

## Changes

- New `alwaysUseSubtitleForSentence` Anki setting (type, default `false`, import/export validation schema)
- Mining settings UI toggle
- New English localization key, propagated to all locales via `scripts/loc/merge-loc-from-en.js`
- Honored at both decision points: the extension video binding and the web-app subtitle player

Assisted-by: Claude:claude-opus-4-8


killergerbah:
@cvivesa Is there a reason this setting is preferable to configuring Yomitan to not supply a sentence instead?

cvivesa:
> @cvivesa Is there a reason this setting is preferable to configuring Yomitan to not supply a sentence instead?

this would work but I personally use Yomitan to mine other sources on the browser where I do want to supply the sentence, so this is a quick setting that allows you to use your existing yomitan setup without needing to modify it

PR #1045, opened 2026-06-06
Add setting to hide page elements during clean screenshots

cvivesa:
## Summary

asbplayer's **Clean screenshot when mining** option only hides asbplayer's own UI (subtitles, controls, etc.). Overlays drawn by *other* extensions — most notably a dictionary pop-up such as [Yomitan](https://yomitan.wiki/)'s — are not asbplayer's to hide, so they get baked into the mined screenshot.

This is especially common in the **one-click mining** flow, where you mine by clicking inside Yomitan's pop-up while it's still on screen, so the pop-up is almost always present when the screenshot is captured.

This PR adds a **Hide elements during clean screenshots** setting (Settings → Streaming video → Mining, shown only when "Clean screenshot when mining" is enabled). It accepts CSS selectors, one per line. While capturing a clean screenshot, asbplayer applies `display: none` to anything matching, then restores it immediately afterward.

`display: none` is used (rather than `visibility: hidden`) so the rule wins the cascade even against overlays that set an inline `visibility: visible !important` — e.g. Yomitan's pop-up iframe.

## Notes on Yomitan

By default Yomitan renders its pop-up inside a closed shadow DOM (its "Use a secure container around popups" security option), which page-side scripts like asbplayer can't reach. To use this with Yomitan, that option must be turned off — ideally scoped to trusted mining sites via a URL-conditioned Yomitan profile. This is documented in the guide and settings reference.

## Changes

- New `streamingScreenshotHiddenSelectors` setting (interface, default `''`, import/export schema).
- Settings UI field in the Streaming video → Mining section.
- `binding.ts` injects/removes a `<style>` rule around clean screenshot capture.
- English localization string (propagated to all locale files via the merge script).
- Docs: settings reference entry + a "Keep the Yomitan pop-up out of screenshots" section in the one-click mining guide.

## Testing

- `yarn workspace @project/extension run build:dev` succeeds (type-checks).
- ESLint and Prettier clean on changed source.
- `settings-import-export` tests pass.

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/common/components/StreamingVideoSettingsTab.tsx
+++ b/common/components/StreamingVideoSettingsTab.tsx
@@ -55,6 +55,7 @@ const StreamingVideoSettingsTab: React.FC<Props> = ({
         streamingRecordMedia,
         streamingTakeScreenshot,
         streamingCleanScreenshot,
+        streamingScreenshotHiddenSelectors,
         streamingCropScreenshot,
         streamingSubsDragAndDrop,
         streamingAutoSync,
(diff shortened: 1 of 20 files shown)


killergerbah:
@cvivesa Thanks for the contribution. I turned off Yomitan's "Use a secure container around popups" and asbplayer seems to hide the popup ok, without this change. 

asbplayer's "clean screenshot when mining" feature is intended to hide everything that makes sense i.e. any DOM that's on top of the video. 

Are you seeing any situations where simply disabling "Use a secure container around popups" and using the main fork of asbplayer, does not work? If so, then I can see the value behind the setting you added. However, if the change is intended to benefit Yomitan users specifically, then the change does not seem necessary.

PR #1048, opened 2026-06-15
feat: add get-subtitles command

mcgrizzz:
* [x] I have read the [[contribution guidelines](https://github.com/asbplayer/asbplayer/blob/main/CONTRIBUTING.md)](https://github.com/asbplayer/asbplayer/blob/main/CONTRIBUTING.md).

As part of https://github.com/asbplayer/asbplayer/issues/840

This PR adds a way to return a piece of media's currently loaded subtitles as SRT over the Websocket/HTTP API.

By default it grabs the loaded subtitles for the current active tab if available, similar to the rest of the WS commands. Otherwise, `mediaId` and `trackNumber` can be specified. These can be obtained by first calling the new `get-bound-media` endpoint.

Implementation notes:

* For app-loaded videos, subtitles are requested from the app when needed using a new internal `request-local-subtitles` extension command.
* For streaming videos, this uses the existing `request-subtitles` `ExtensionToVideoCommand`.
* App/local videos require an explicit `mediaId`. Without one they will not be matched. This only happens when there are no streaming videos to default to and the API is called without a `mediaId`. I'm not sure what a better default should be here. Perhaps if there is only one asbplayer instance we could choose that one, but otherwise it is ambiguous.


common/web-socket-client/web-socket-client.ts
```diff
@@ -63,6 +63,19 @@ interface GetBoundMediaResponseBody {
...
+    command: 'get-subtitles';
+    messageId: string;
+    body: {
+        mediaId?: string;
+        trackNumber?: number;
+    };
+}
+
+interface GetSubtitlesResponseBody {
+    srt: string;
```
killergerbah: I know `srt` is a well-known format but is there a reason to favor this over a structured format like 
```
{
  "subtitles": ["text": "foo", "start": 1234, "end": 5678]
}
```
?

extension/src/services/web-socket-client-binding.ts
```diff
@@ -43,6 +49,55 @@ const cyrb53 = (str: string) => {
...
+    return subtitles.filter((subtitle) => subtitle.track === selectedTrack);
+};
+
+// Requests the loaded subtitles from a local asbplayer app instance
+const requestSubtitlesFromAsbplayer = (
+    tabRegistry: TabRegistry,
+    asbplayerId: string
+): Promise<SubtitleModel[] | undefined> => {
+    return new Promise((resolve) => {
+        const messageId = crypto.randomUUID();
```
killergerbah: Would it make sense to use the `uuid` package instead, for consistency? Can consider moving to `crypto` package in a later update.

docs/docs/reference/external-api.md
```diff
@@ -149,13 +149,49 @@ Returns the media asbplayer is currently tracking, including both `streaming` an
...
+
+```javascript
+{
+    "command": "get-subtitles",
+    // Message ID to correlate with asbplayer's response
+    "messageId": "2d7e8a1b-9c4f-4e3a-8b6d-1f0c5a2e9d7b",
+    "body": {
+        // Optional media id from `get-bound-media`. Defaults to the active tab when omitted.
+        "mediaId": "a1b2c3d4e5f6",
+        // Optional track number from `get-bound-media`. Defaults to the lowest-numbered loaded track when omitted.
```
killergerbah: Would it make sense to allow callers to specify multiple track numbers instead of requiring one track number at a time?

Also, would it make sense to return all the tracks if no track number is specified? To me, defaulting to the lowest track number feels oddly specific, when compared to just returning all of the subtitles.

common/src/message.ts
```diff
@@ -435,6 +435,15 @@ export interface RequestSubtitlesFromAppMessage extends MessageWithId {
     readonly command: 'request-subtitles';
 }
 
+export interface RequestLocalSubtitlesMessage extends MessageWithId {
```
killergerbah, line 438: We might benefit in the future from instead re-using the `request-subtitles` request/response protocol that is already defined for requesting subtitles from the extension

extension/src/services/web-socket-client-binding.ts
```diff
@@ -43,6 +49,55 @@ const cyrb53 = (str: string) => {
...
+        let timeout: ReturnType<typeof setTimeout>;
+
+        const listener = (request: any) => {
```
killergerbah: It seems this is the first time that any code is attempting to request something from an asbplayer app instance synchronously. Can this pattern be handled at a lower level? Maybe `TabRegistry` can implement a new method that returns a response like `publishCommandToAsbplayersAndAwaitResponse` for messages that specify a `messageId`.

killergerbah (commented):
Appreciate the continued work on this and sorry for the delay

common/web-socket-client/web-socket-client.ts
mcgrizzz: Got it. Though I think we should include 'track' as well for each that way we can do 

 ```
 {
   "subtitles": [
          {"text": "foo", "start": 1234, "end": 5678, "track": 0}, 
          ...,
          {"text": "foo2", "start": 1234, "end": 5678, "track": 1},
          ...
      ]
 }
 ```

Then the lines can be cleanly separately on the client.

extension/src/services/web-socket-client-binding.ts
mcgrizzz: Will do.

docs/docs/reference/external-api.md
mcgrizzz: Will do.

mcgrizzz:
One note about the go library change I've done here. While testing with multiple tracks loaded on videos, the messages become large enough that chrome decides to fragment the WS messages. The library that was being used (golang.org/x/net/websocket) does not support fragmented messages from what I could see. I opted for what seemed to be the most used WS library on pkg.go.dev, gorilla websocket.

scripts/web-socket-server/cli/subtitles
```diff
@@ -0,0 +1,11 @@
...
+
+MEDIA_ID="$1"
+TRACK_NUMBERS="$2"
+
+QUERY=""
+[ -n "$MEDIA_ID" ] && QUERY="mediaId=$MEDIA_ID"
+[ -n "$TRACK_NUMBERS" ] && QUERY="${QUERY:+$QUERY&}trackNumbers=$TRACK_NUMBERS"
+
+curl -H "content-type: application/json" \
+    -X GET "http://127.0.0.1:8766/asbplayer/subtitles${QUERY:+?$QUERY}"
```
killergerbah: Can you give this file executable permissions?

scripts/web-socket-server/go.mod
```diff
@@ -7,6 +7,8 @@ require (
 	github.com/labstack/echo/v4 v4.11.4
 )
 
+require github.com/gorilla/websocket v1.5.3 // indirect
```
killergerbah: Can you move this into the `require` block above, run `go mod tidy` in this directory, and commit the changes?

common/web-socket-client/web-socket-client.ts
```diff
@@ -165,6 +186,22 @@ export class WebSocketClient {
...
+                                payload.body?.trackNumbers
+                            );
+                            const response: Response<GetSubtitlesResponseBody> = {
+                                command: 'response',
+                                messageId,
+                                body: { subtitles },
+                            };
+                            const serialized = JSON.stringify(response);
+                            this._socket?.send(serialized);
+                            console.log('[ws] get-subtitles response sent', messageId);
```
killergerbah: Is this a debug log for your own testing? We can keep it but I would prefer for the file to be consistent on this with all of the messages. Or you can remove it, up to you.

docs/docs/reference/external-api.md
```diff
@@ -149,13 +149,51 @@ Returns the media asbplayer is currently tracking, including both `streaming` an
 }
 ```
 
+### `get-subtitles`
```
killergerbah, line 154: Can you update the docs for this block `get-subtitles` and also `get-bound-media` to indicate they are extension-only for now?

killergerbah (commented):
Thank you, did another pass

common/web-socket-client/web-socket-client.ts
mcgrizzz: Yes, sorry about that. Missed it during clean up.

killergerbah (approved):
Awesome, thank you

scripts/web-socket-server/cli/subtitles
killergerbah: FYI this script didn't really work for me because the newlines prevented the shebang from being parsed properly. Was it working on Windows? I'm assuming not unless you use a Linux emulator that ignores the newline format

PR #1052, opened 2026-06-18
Virtualize subtitles to improve performance and memory usage

ShanaryS:
Rather than trying to fight with sub-optimal solutions, I instead just went for virtualization with `react-virtuoso` for `SubtitlePlayer` and `StatisticsSentenceDetailsDialog`. This gives the biggest upflift in all areas and is easier to manage, it also guards against any potential complication in the future. There are a few things that broke due to virtualization that I had to address but there may be more subtle ones.

Since the rows are now virtualized, I added an override for the browsers `CTRL+F` to allow the expected behavior of searching the entire subtitle.
- We will search the `text` and `reading` (considered `frequency` as well but opted not to) on each subtitle with the same normalization helpers found in `WordBrowser` (romaji, lemma, etc). It's not fully permissive like `ANY_FORM_COLLECTED`, for that we would need to store the lemmas that `SubtitleAnnotations` queried from yomitan to lookup against, but I don't think it's worth doing right now.
- I added support for regex searching using `/regex/`.
- As a bonus, users can now search the subtitles when they are loaded in the side panel (still uses the browsers native search if side panel is not focused).

I also made a change in the hover behavior so that its triggered when hovering over the the subtitle row rather than the subtitles themselves. This makes it more stable as the change in annotations can move the pointer off the subtitles causing a flicker between hover and non-hover especially between spaces and multiple lines. I've previously made this change for video subtitles.

I added a smaller buffer for the video annotations as well. Even though these aren't as problematic, it should help with the memory usage as well with no real downside.

Finally, I made the rich text rendering fully JIT. I got rid of `richText` and `richTextOnHover` from the subtitle objects since we are always rendering it only when we need it and it gets destroyed once it's outside the virtualized view for the subtitle list or time range for video subtitles.

Profiling with all these changes shows there is no more long term memory we can clear up. The browser task manager may still show usage up to `1GGB` for large subtitles temporarily but in reality it's not that high. Real memory usage is around the `100MB` mark.

Assisted-by: Copilot:Opus-4.8

### DIFF
--- a/common/app/components/SubtitlePlayer.tsx
+++ b/common/app/components/SubtitlePlayer.tsx
@@ -74,6 +74,8 @@ import { MineSubtitleCommand, WebSocketClient } from '../../web-socket-client';
 import { clampSubtitlePlayerWidth } from './video-subtitle-split';
 import './subtitles.css';
 
+const findDelayMs = 300;
+
 let lastKnownWidth: number | undefined;
 export const minSubtitlePlayerWidth = 200;
 const calculateInitialWidth = () => lastKnownWidth ?? Math.max(350, 0.25 * window.innerWidth);
@@ -262,7 +264,7 @@ const parseRegexQuery = (query: string): RegExp | undefined => {
 
 const subtitleSearchableText = (subtitle: DisplaySubtitleModel): string => {
     const tokens = subtitle.tokenization?.tokens;
-    if (!tokens || !tokens.length) return subtitle.text;
+    if (!tokens?.length) return subtitle.text;
     let readings = '';
     for (const token of tokens) {
         for (const reading of token.readings) {
@@ -1016,7 +1018,7 @@ export default function SubtitlePlayer({
 
             if (cancelled) return;
             setFindExpansion({ query: trimmed, terms: normalizedLookupTerms(...queryForms, ...lemmaTerms) });
-        }, 300);
+        }, findDelayMs);
 
         return () => {
             cancelled = true;
--- a/common/app/components/SubtitlePlayer.tsx
+++ b/common/app/components/SubtitlePlayer.tsx
@@ -1,6 +1,16 @@
-import React, { ForwardedRef, useCallback, useEffect, useState, useRef, createRef, RefObject, ReactNode } from 'react';
+import React, { ForwardedRef, useCallback, useEffect, useMemo, useState, useRef, ReactNode } from 'react';
 import { makeStyles } from '@mui/styles';
 import { type Theme } from '@mui/material';
+import {
+    ContextProp,
+    ItemProps,
+    ScrollerProps,
+    TableBodyProps,
+    TableComponents,
+    TableProps,
+    TableVirtuoso,
+    TableVirtuosoHandle,
+} from 'react-virtuoso';
 import { keysAreEqual } from '../services/util';
 import { useResize } from '../hooks/use-resize';
 import { ScreenLocation, useDragging } from '../hooks/use-dragging';
@@ -8,13 +18,17 @@ import { useTranslation } from 'react-i18next';
 import {
     PostMineAction,
     SubtitleModel,
-    SubtitleHtml,
     AutoPauseContext,
     CopySubtitleWithAdditionalFieldsMessage,
     CardTextFieldValues,
     RichSubtitleModel,
 } from '@project/common';
-import { AsbplayerSettings, TokenAnnotationConfig, tokenAnnotationStyleValues } from '@project/common/settings';
+import {
+    AsbplayerSettings,
+    DictionaryTrack,
+    TokenAnnotationConfig,
+    tokenAnnotationStyleValues,
+} from '@project/common/settings';
 import {
     surroundingSubtitles,
     mockSurroundingSubtitles,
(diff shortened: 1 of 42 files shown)


common/app/components/SubtitlePlayer.tsx
```diff
@@ -194,7 +207,32 @@ const useSubtitleRowStyles = makeStyles<Theme>((theme) => ({
     },
 }));
 
-export interface DisplaySubtitleModel extends RichSubtitleModel {
+const useFindBarStyles = makeStyles<Theme>((theme) => ({
```
killergerbah: I know we're still using this in a lot of places but it seems like the direction MUI is going in is to deprecate `useStyles` and to prefer `sx` or `styled` components. I think as a matter of principle we should just avoid using `useStyles` for new code.

killergerbah (commented):
Tested briefly and it seems to be working great, even with image-based PGS subtitles

I'll give a more careful read when I have a chance

common/app/components/SubtitlePlayer.tsx
```diff
@@ -712,15 +957,179 @@ export default function SubtitlePlayer({
...
+                align: 'center',
                 behavior: 'smooth',
             });
             setHighlightedJumpToSubtitleIndex(jumpToIndex);
             setTimeout(() => setHighlightedJumpToSubtitleIndex(undefined), 1000);
         }
-    }, [jumpToSubtitle, subtitles, subtitleRefs, onSeek, onJumpToSubtitleHandled, clock]);
+    }, [jumpToSubtitle, subtitles, onSeek, onJumpToSubtitleHandled, clock]);
+
+    useEffect(() => {
```
killergerbah: Would it be possible to consolidate find-related hooks into their own file, or at least into a separate hook?

A lot of these seem to be a chain of `useEffects` intended to produce the final match results, with some callbacks into the UI for scrolling. I think it should be possible to encapsulate these cleanly inside a single custom hook.

common/app/components/VideoPlayer.tsx
```diff
@@ -1673,23 +1677,35 @@ export default function VideoPlayer({
...
+    const { getSubtitleDomCache, refreshSubtitleWindowDomCache, updateSubtitleWindowDomCache } = useSubtitleDomCache(
+        subtitles,
+        getSubtitleHtml
+    );
 
     useEffect(() => {
         domCacheRef.current = getSubtitleDomCache();
     }, [getSubtitleDomCache]);
 
+    useEffect(() => {
```
killergerbah: Small trick, you can write 

```
refreshSubtitleWindowDomCacheRef.current = refreshSubtitleWindowDomCache;
```

without the `useEffect` and the result is the same but without having to allocate the function passed into `useEffect`.

common/app/hooks/use-subtitle-dom-cache.ts
```diff
@@ -13,15 +13,33 @@ export const useSubtitleDomCache = (
         return () => domCache.clear();
     }, [subtitles, render]);
 
-    return {
-        getSubtitleDomCache: () => {
-            if (domCache.empty) {
-                for (const subtitle of subtitles) {
-                    domCache.add(String(subtitle.index), render(subtitle));
-                }
+    const refreshSubtitleWindowDomCache = useCallback(
```
killergerbah: Nitpick:
There doesn't seem to be anything particularly window-y about this function so would it be clearer to name it `refreshSubtitleDomCache` or if you really want to make it clear that it's for some subtitles, something like `refreshSubtitleDomCacheForSubtitles`?

The `window` seems to be an implementation detail of how the caller is using the function.

common/app/hooks/use-subtitle-dom-cache.ts
```diff
@@ -13,15 +13,33 @@ export const useSubtitleDomCache = (
...
+            for (const subtitle of windowSubtitles) {
+                const key = String(subtitle.index);
+                if (!domCache.has(key)) domCache.add(key, render(subtitle));
             }
+        },
+        [domCache, render]
+    );
 
-            return domCache;
+    const updateSubtitleWindowDomCache = useCallback(
```
killergerbah: Same here, wonder if `updateSubtitleDomCache` would be a clearer name?

extension/src/controllers/subtitle-controller.ts
```diff
@@ -219,6 +220,44 @@ export default class SubtitleController {
...
+        );
+        if (!windowSubtitles.length) {
+            const { lastShown, nextToShow } = this.subtitleAnnotations.subtitlesAt(now);
+            for (const subtitle of lastShown ?? []) windowSubtitles.push(subtitle);
+            for (const subtitle of nextToShow ?? []) windowSubtitles.push(subtitle);
+        }
+        return windowSubtitles;
+    }
+
+    private _refreshCachedHtmlWindow() {
```
killergerbah, line 237: If I understand this correctly, the cache is being updated for the subtitle window prior to the `_renderSubtitles` call in the subtitle update interval. Would it make sense to do the caching inside `_renderSubtitles` itself? In particular, the subtitles don't need to be cached if subtitles aren't actually displaying

ShanaryS: I thought the cache existed to make sure that subtitles show up as quickly as possible once the timestamp is reached. Before it was caching the entire subtitle and now it's a time range around the current time. I thought I saw the subtitles showing earlier when cached but maybe that was placebo. So should I only cache the showing subtitles for SubtitleController and VideoPlayer? That would make the design much more straightfoward.

killergerbah: Actually I'm wrong. Now that I'm remembering, the original reason the cache existed was so that any mutations on the DOM performed by external tools (like the JPDB extension) would persist even after the subtitle disappears from screen. Any performance improvements were secondary.
So instead caching by window makes sense.

killergerbah (approved):
Did a final pass and nothing jumps out
Can you add the react-virtuoso to the dependency list on the about page?
This is a really nice change

PR #1054, opened 2026-06-20
feat: add subtitle sync to a reference

### LINKED ISSUE #441: Sync subtitles with embedded subtitles
I know automatic subs syncing was requested almost 3 years ago but the specific feature I'm requesting is not for the subtitles to be synced with audio, since that seems more complicated and I'm not sure if that's even possible with the extension. 

Many times on Netflix or YouTube, there's an existing subtitle track but it's not the target language. Then I get some target language subs online but they're often untimed and I always struggle with adding or removing the correct offset. Maybe I don't know how to use the offset buttons well but they seem kind of buggy. 

I remember there's some useful mpv script that can let you sync to an existing subtitle track, I think no matter the language, and that was pretty accurate in the past when I used mpv. Also didn't seem to take long for that to process. Do you think this could be added to asbplayer as a button in the player? I think this would be super useful. Thank you! 

https://animecards.site/subtitles/

zakwarsame:
closes https://github.com/asbplayer/asbplayer/issues/441

Add a control that aligns the loaded subtitle track to a reference by comparing cue timing, then applies the detected offset.

A pure, language-agnostic aligner (Jaccard cross-correlation over a coarse timeline) computes the constant offset between two subtitle tracks plus a confidence score. A dialog lets the user pick the reference (an uploaded file or another loaded track), shows the detected offset and confidence, warns on low confidence, and supports undo.

The trigger is a shared button surfaced in the player controls (web app and over-video player) and in the extension side panel, where it applies the offset to the synced tab via the existing offset command.

(_fyi: i've been using a slightly advanced version of this in [a fork](https://github.com/zakwarsame/asbplayer) which detects and loads cc captions in the drop-down; but decided to abstract just the core algorithm with minimal UI here_)

#### demo

https://github.com/user-attachments/assets/72719a3e-b557-4b4c-8714-87619f45080f

### testing steps used in the demo

1. episode: https://www.miruro.tv/watch/101348/vinland-saga?ep=4
2. switch server to bun ([screenshot for reference](https://github.com/user-attachments/assets/266b3d26-7241-4e11-b759-61c32524fff4))
3. load un-aligned japanese subtitle: https://files.catbox.moe/79mqrx.srt
4. sync with reference to english version: https://files.catbox.moe/zey62f.srt
5. notice offset

idea credit: @Kellenok

### DIFF
--- a/common/src/subtitle-sync/offset.ts
+++ b/common/src/subtitle-sync/offset.ts
@@ -1,0 +1,99 @@
+import type { SubtitleModel } from '../model';
+
+/** Timeline resolution. Coarse enough that a full episode scores in milliseconds. */
+const FRAME_MS = 100;
+/** Largest offset (in either direction) the search will consider. */
+const MAX_OFFSET_MS = 120_000;
+/** Tracks with fewer active frames than this fraction can't be correlated reliably (signs/songs). */
+const MIN_DENSITY = 0.1;
+
+export interface SubtitleSyncResult {
+    /** Milliseconds to add to the primary track's cue timings so it aligns to the reference. */
+    offset: number;
+    /** Jaccard overlap of the best alignment, in [0, 1]. */
+    confidence: number;
+}
+
+type Cue = Pick<SubtitleModel, 'originalStart' | 'originalEnd'>;
+
+/** Binary timeline where a frame is true when any cue is active during it. */
+function buildTimeline(cues: Cue[], durationMs: number): boolean[] {
+    const timeline = new Array<boolean>(Math.ceil(durationMs / FRAME_MS)).fill(false);
+
+    for (const { originalStart, originalEnd } of cues) {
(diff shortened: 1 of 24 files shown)


common/src/subtitle-sync/offset.ts
zakwarsame, line 1: this file is basically the algorithm engine that handles the alignment. the rest of the pr is just UI wiring and tests.

killergerbah:
@zakwarsame This is super cool, thanks for opening this PR, and sorry for taking so long to get to this.

What's the primary use case for subtitle syncing like this? Do you think it would be ok to assume that the user will usually be loading multiple subtitle tracks?

The reason I ask, is because if the user is loading multiple subtitle tracks anyway, then I think the most efficient UX would reuse the existing methods of loading subtitles i.e. the subtitle track selector. I imagine then that the syncing can be accessed via a single button.

PR #1060, opened 2026-07-02
Fix Netflix subtitle detection and furigana positioning

### LINKED ISSUE #1047: Netflix ruby displaying over wrong text
Play any show where a word with ruby follows a word without ruby - instead of displaying over just the word it's intended for, the ruby will display over everything to the left of it (seems until it finds a whitespace/special char.)

Screenshot with native Netflix subs on top and asb subs on the bottom:

Image

いて should render over 痛 rather than 頭痛

Original sub:
```
<p xml:id="subtitle2" begin="462128332t" end="493409582t" region="region0" style="style0" xml:space="preserve">（<span style="style1"><span style="style2">小林</span><span style="style3">こばやし</span></span>）うー 頭<span style="style1"><span style="style2">痛</span><span style="style3">いて</span></span>え</p>
```

Sub in asb without "Detect and display ruby" on:
```
（小林(こばやし)）うー 頭痛(いて)え
```

Not sure how you'd fix this without a big refactor because it looks like parsing the Netflix subs and then parsing them for ruby are seperate steps and in the first step you actually just lose the information of what word the ruby applies to.

I was thinking you might be able to fix this by adding an invisible whitespace char before words with ruby in the initial parse but that feels like a hack/would have other unwanted effects.

**What browser?**
Chrome

Streaming

1.17.0

Netflix

N/A

### LINKED ISSUE #1051: Netflix subtitles no longer detected in Chrome extension v1.18.0
asbPlayer v1.18.0 on Chrome 149 / macOS no longer detects Netflix subtitles.

Symptoms:
- Netflix native Japanese subtitles display correctly.
- asbPlayer works on YouTube.
- On Netflix, asbPlayer detects the video title but all subtitle tracks show Empty.
- Tested multiple Netflix shows that previously worked.
- Disabled all other Chrome extensions, quit/reopened Chrome, and retested with only asbPlayer enabled.
- Cleared Netflix site data.
- Hard-refreshed Netflix after enabling native subtitles.
- asbPlayer has “Read and change all your data on all websites” permission.

Console from Inspect Popup:
SecurityError: User activation is required.
Error: Failed to fetch
at extension-global-sta…ider-BL38Papk.js:32
at e._executeAction (Close-BW4y2HBk.js:4:1827)
at async e.version (Close-BW4y2HBk.js:3:91925)
at async extension-global-sta…ider-BL38Papk.js:32:1592

Service worker console showed no issues.

Netflix console also showed:
Refused to get unsafe header "X-OneTrust-IsBot"
Permissions policy violation: unload is not allowed in this document.

Screenshots attached.

Image
Image

Image

jprostko:
### Problem
* asbplayer stopped detecting subtitles on Netflix.
* Netflix furigana readings rendered over the wrong character.

### Cause
* Detection: Netflix now processes the manifest inside its player core rather than on the main thread, so the subtitle track URLs no longer pass through the `window.JSON` hooks the Netflix page script used to collect them, and `buildResponse()` filters out every track. The subtitle files on this path are now IMSC 1.1 (TTML) instead of webvtt.
* Furigana: the ruby tokenization matched the whole run of kanji and kana ending at a reading, so a reading was bound to the preceding characters along with its base. A reading centers over its base, so widening the base drifted the reading onto the wrong character. In `ひろ子`, `こ` belongs to `子` but rendered over `ろ`.

### Fix
* Read subtitle URLs from the most recent player session state instead of the `window.JSON` hooks, walking that state for timed-text URL objects rather than hardcoding the deeper subtitle URL path. Playback controls are unchanged, and lazy per-language fetch is preserved.
* Add a `.nfimsc` IMSC 1.1 parser: tick timestamps against `ttp:tickRate`, `dur`, namespaced elements and attributes, multiple `<div>`s, `tts:ruby` rendered as `base(reading)`, and a finite-timestamp guard. `_parseTtmlTimestamp` gains tick and offset handling and keeps its clock-time behavior for the Amazon Prime Video path.
* Guard the `.dfxp`/`.ttml2` path against non-finite timestamps so the new tick handling in `_parseTtmlTimestamp` cannot leak `NaN` into that path.
* Register the `.nfimsc` extension.
* Fence each ruby base during IMSC parsing so tokenization binds a reading to only its own base instead of the preceding characters, correcting the furigana position.

### Tested
* `yarn verify` passes
* Unit tests cover the `.nfimsc` parser (tick timestamps, `dur`, namespaced markup, multiple `<div>`s, nested spans, ruby on and off, missing tick rate), the `.dfxp`/`.ttml2` finite-timestamp guard, and ruby fencing (a preceded reading binds to its base, a reading containing a closing paren stays literal).
* On live Netflix: track detection, on-demand language loading, ruby rendering with the setting on and off, and furigana sitting over their own base in cases that previously drifted (`田の実`, `ひろ子`).
* Loaded/tested the Chrome and Firefox builds with no issues.

### Additional notes
No cross-site impact expected. The `.nfimsc` parser is Netflix only, and the shared `_parseTtmlTimestamp` change stays backward compatible for the clock-time timestamps the Amazon Prime Video `.dfxp`/`.ttml2` path uses. The ruby fix is gated behind the existing furigana setting and only changes how a reading binds to its base.

One of the test URLs for the #1047 fix was https://www.netflix.com/watch/82122787 . A couple testable timestamps are 4:14 and 14:06.

Fixes #1047
Fixes #1051

- [X] I have read the [contribution guidelines](https://github.com/asbplayer/asbplayer/blob/main/CONTRIBUTING.md).

### DIFF
--- a/common/subtitle-reader/subtitle-reader.ts
+++ b/common/subtitle-reader/subtitle-reader.ts
@@ -350,6 +350,10 @@ export default class SubtitleReader {
             return await this._parsePgs(file, track);
         }
 
+        if (file.name.endsWith('.nfimsc')) {
+            return this._parseNetflixImsc(await file.text(), track);
+        }
+
         if (file.name.endsWith('.dfxp') || file.name.endsWith('ttml2')) {
             const text = await file.text();
             const parser = new DOMParser();
@@ -366,11 +370,19 @@ export default class SubtitleReader {
                     continue;
                 }
 
+                const start = this._parseTtmlTimestamp(beginAttribute);
+                const end = this._parseTtmlTimestamp(endAttribute);
+
+                // Skip cues whose timestamps did not parse to a finite number.
+                if (!Number.isFinite(start) || !Number.isFinite(end)) {
+                    continue;
+                }
+
                 const text = this._decodeHTML(elm.innerHTML.replaceAll(/<br(\s[^\s]+)?(\/)?>/g, '\n'));
                 subtitles.push({
                     text: this._filterText(text),
-                    start: this._parseTtmlTimestamp(beginAttribute),
-                    end: this._parseTtmlTimestamp(endAttribute),
+                    start,
+                    end,
                     track,
                 });
             }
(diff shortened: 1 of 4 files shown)


common/subtitle-reader/subtitle-reader.ts
```diff
@@ -104,16 +120,24 @@ export default class SubtitleReader {
...
-            return this._deduplicate(allNodes);
-        }
-
         if (this._convertNetflixRuby) {
-            for (const node of allNodes) {
-                this._convertNetflixRubyToHtml(node);
+            if (flatten) {
+                // Flattened output keeps inline base(reading) without tokenizing, so
+                // the ruby base markers are simply dropped here.
+                for (const node of allNodes) {
```
killergerbah, line 127: Do you mind explaining this a bit more?
```
Flattened output keeps inline base(reading) without tokenizing
```
IIRC "flattening" is intended to help with overlapping subtitle tracks that often appear in m3u8 manifests so I'm having trouble making the connection between `flatten` and the tokenization behavior of `subtitle-annotations.ts`

killergerbah (commented):
Great work, thanks again as always

This seems to be working well, and nothing jumps out on the first pass, just left one question

common/subtitle-reader/subtitle-reader.ts
jprostko: The comment in the code is describing behavior that already existed, rather than a new design decision. If you think it is worded strangely, I can take another crack at it, or you can modify it to your liking.

Basically, the connection you spoke of is really just order of operations. The flatten early return used to come before the ruby conversion loop, so flattened output has never been tokenized and always kept the literal base(reading) form. The diff does restructure that block, since the marker strip needs to run before the early return, but the observable behavior is unchanged.

The reason ruby comes up in the flatten branch at all is because the IMSC parser now puts an invisible marker in front of each ruby base. The ruby conversion normally consumes that marker, but flattened output never goes through the conversion, so the loop right under the comment you quoted strips the markers instead. Without it they would leak into the flattened text.

Flatten and the Netflix ruby handling don't even combine in practice right now, since flatten comes from multi-file syncs (the overlapping m3u8 case you mentioned) and Netflix syncs one file per track. The strip is just making the marker safe on every output path.

killergerbah (approved):
Did another pass and nothing jumps out

Thanks again for the critical fix
I always appreciate your very thorough work

common/subtitle-reader/subtitle-reader.ts
killergerbah: Understood, I may update the comment accordingly since I'm not really sure right now how safe it is to just pipe everything through the ruby converter and have `flatten` only used for deduplication

PR #1063, opened 2026-07-05
Refactor `subtitle-annotations.ts` and fix version mismatch

ShanaryS:
The main change is a refactoring of `subtitle-annotations.ts` into multiple files under a new `annotations` folder. This clearly separates responsibilities to reduce the size of the file and makes future expansion of annotations beyond subtitles easier.

I also enabled support for pitch accent from tokenize for the `scanning-parser` since the Yomitan PR has been merged. This allows the optimization to automatically enable on Yomitan's next release.

I made a QoL UX change to clearly show what settings enable annotations for a track.
annotations

Finally I removed the UI guard around the extension version for the new annotation config settings. Since the initial release of the annotations feature in `v1.14`, the extension and app has rendered their own subtitles using the settings they are aware of. So while the app and extension settings for these won't be in sync, they can somewhat render independently
-  If the media is app controlled then everything in `v1.19` works
- If the media is extension controlled:
  - No pitch accent data will get sent to the synced app instance
  - Frequency annotation must be enabled in the extension settings for the App to receive them which is only populated for the statuses the settings specify
  - If the user were to change the non-synced settings in the app settings, the extension will not rebuild annotations since the annotation settings will still seem equal.

### DIFF
--- a/common/components/DictionarySettingsTab.tsx
+++ b/common/components/DictionarySettingsTab.tsx
@@ -51,7 +51,7 @@ import {
 } from '@project/common/settings';
 import { Anki } from '../anki';
 import { WaniKani, WaniKaniUser } from '../wanikani';
-import { Yomitan } from '../yomitan/yomitan';
+import { Yomitan } from '../yomitan';
 import SwitchLabelWithHoverEffect from './SwitchLabelWithHoverEffect';
 import SettingsTextField from './SettingsTextField';
 import SettingsSection from './SettingsSection';
@@ -81,7 +81,12 @@ import {
     percentToHex2,
 } from '../util';
 import DictionaryImport from './DictionaryImport';
-import { computeRichText, getAnnotationsForRender, getAnnotationsHtml, InternalToken } from '../subtitle-annotations';
+import {
+    computeRichText,
+    getAnnotationsForRender,
+    getAnnotationsHtml,
+    InternalToken,
+} from '@project/common/annotations';
 import WordBrowserDialog from './WordBrowserDialog';
 import '../app/components/subtitles.css';
 import ButtonGroup from '@mui/material/ButtonGroup';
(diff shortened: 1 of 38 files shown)


common/components/DictionarySettingsTab.tsx
```diff
@@ -1185,310 +1174,126 @@ const DictionarySettingsTab: React.FC<Props> = ({
...
-                                variant="outlined"
-                                size="small"
-                                label={t(labelKey)!}
-                                value={value}
-                                helperText={
-                                    annotation === 'pitchAccent'
-                                        ? dictionaryTokenPitchAccentUnderlineOverlineStyleWarning
-                                        : undefined
+                        >
+                            {t('settings.dictionaryEnableAnnotations')}
```
killergerbah, line 1195: The highlighting + "enable annotations" text is much more informative than before, but I wonder if there's an even more obvious way to indicate to the user that annotations are enabled. In particular, the current wording "Enable Annotations" feels more suitable to a button that the user would press, since the it feels like a command rather than a statement of something being on/off. 

Also, does the on/off state itself have any concrete affect on UX? Just wondering of what concern it is to the user. The only thing I can think of is, that if annotations are on, then the user needs to configure all the dependent Anki/Yomitan settings for it to work.

killergerbah (commented):
Did a pass and nothing jumps out regarding the refactor.
I assume the refactor involves mostly reorganization of existing code without any adjustment of the underlying logic. I do recognize a lot of very similar-looking (I'm assuming nearly or exactly the same) code. My understanding of the annotation code is still very shallow so I can't make any useful comments without spending a lot of time reading.

I left one comment regarding the UX change in the settings. Would be curious to hear what you think.

common/components/DictionarySettingsTab.tsx
ShanaryS: Yes, if they just turn it on then there will be red strikethrough text on everything since it's likely not configured. It also helps inform the user that these settings must be enabled for anything else for this track to be relevant. Feel free to change it to whatever you think makes sense.

ShanaryS:
> I assume the refactor involves mostly reorganization of existing code without any adjustment of the underlying logic.

The only two behavior changes are the version gate of pitch accent from `scanning-parser` and the new behavior  for older extensions. The refactor is purely copy-pasted code.

killergerbah (approved):

killergerbah:
Feel free to merge when ready @ShanaryS

PR #1065, opened 2026-07-06
Add stronger eslint rules and fix various bugs

ShanaryS:
_This PR should be merged after the `subtitle-annotations.ts` refactor #1063_

The goal is to strengthen the type safety throughout the repo by elimintate whole classes of bugs. It was tedious to update the existing codebase but it should to fairly simple to review as most are trivial type changes.

I added a few extra rules on top of the recommended that I've always found value in. These help identify bugs as usually these errors means something was refactored or never fully thought through.

`@typescript-eslint/no-unnecessary-type-assertion`: Prevents using things such `as` or `!` when it doesn't change the type
`@typescript-eslint/return-await`: No need to await before returning if the function already returns a promise
`@typescript-eslint/await-thenable`: await can only be used when it can be (no awaiting sync code)
`@typescript-eslint/no-floating-promises`: All promises must be explicitly handled (`await fn()`, `fn().catch()`, `fn().then()`) or explictly marked as fire-and-forget through `void fn()`
`@typescript-eslint/no-misused-promises`: Code that expects sync `=> void` callbacks are respected, no passing in `=> Promise<void>`. It must be stripped of the promise by wrapping in a lambda which forces `no-floating-promises` to be respected. Same for `=> Promise<void> callbacks with only `=> void` passed in.

The final commit, `fix violations requiring behavior change`, contains fixes that I think are worth taking a closer look at. These either change the behavior of the code or remants from refactors or unfinished implementations. I'll also add some review comments for other noteworthy things. The `no-async-promise-executor` is also a commit with some slight behavior changes.

Assisted-by: Codex:GPT-5.5


common/app/components/Controls.tsx
```diff
@@ -602,7 +594,6 @@ export default function Controls({
     onPause,
     onPlay,
     onTabSelected,
-    onUnloadVideo,
```
ShanaryS, line 605: It seems the unload was not used in this file.

common/app/components/VideoPlayer.tsx
```diff
@@ -810,7 +811,6 @@ export default function VideoPlayer({
             videoPreview.src = videoFile;
             videoPreview.load();
         } else {
-            settings.thumbnailPreview;
```
ShanaryS, line 813: Should this be used somewhere?

common/components/About.tsx
```diff
@@ -207,15 +207,14 @@ const dependencies: Dependency[] = [
...
-    let count = dependencyPurposeCounts[dep.purpose] ?? 0;
+    const count = dependencyPurposeCounts[dep.purpose] ?? 0;
     dependencyPurposeCounts[dep.purpose] = count + 1;
 }
 
 const About = ({ appVersion, extensionVersion }: Props) => {
     const theme = useTheme<Theme>();
     const { t } = useTranslation();
     const renderedPurpose: { [key: string]: boolean } = {};
-    let purposeIndex = 0;
```
ShanaryS, line 218: Seems like nextPurpose doesn't use this

common/components/AnkiDialog.tsx
```diff
@@ -86,9 +86,6 @@ const boundaryIntervalSubtitleCountRadius = 1;
 const boundaryIntervalSubtitleTimeRadius = 5000;
 
 const boundaryIntervalFromCard = (subtitle: SubtitleModel, theSurroundingSubtitles: SubtitleModel[]) => {
-    let index = theSurroundingSubtitles.findIndex((s) => s.start === subtitle.start);
```
ShanaryS, line 89: These indexes look like they were important

common/components/SettingsForm.tsx
```diff
@@ -210,9 +209,6 @@ interface Props {
     onUnlockLocalFonts: () => void;
 }
 
-// Filter out keys that look like '0', '1', ... as those are invalid
-const cssStyles = Object.keys(document.body.style).filter((s) => !isNumeric(s));
```
ShanaryS, line 214: Doesn't seem relevant anymore

extension/src/controllers/bulk-export-controller.ts
```diff
@@ -7,14 +7,6 @@ export interface BulkExportStartedPayload extends Message {
     total: number;
 }
 
-interface BulkExportCompletedPayload extends Message {
-    command: 'bulk-export-completed';
-}
-
-interface BulkExportCancelledPayload extends Message {
-    command: 'bulk-export-cancelled';
-}
```
ShanaryS, line 16: The messages for completion or cancellation aren't being used.

extension/src/controllers/subtitle-controller.ts
```diff
@@ -685,7 +681,6 @@ export default class SubtitleController {
 
     currentSubtitle(): [IndexedSubtitleModel | null, SubtitleModel[] | null] {
         const now = 1000 * this.context.video.currentTime;
-        let subtitle = null;
```
ShanaryS, line 688: This was being updated but not used

extension/src/ui/components/Tutorial.tsx
```diff
@@ -94,11 +92,7 @@ const ToolbarBubble: React.FC<{ show: boolean; onConfirm: () => void }> = ({ sho
     );
 };
 
-const LoadSubtitlesDialog: React.FC<{ open: boolean; count?: number; onClose: () => void }> = ({
-    open,
-    count,
-    onClose,
-}) => {
```
ShanaryS, line 101: onClose wasn't being used for the Tutorial

ShanaryS (commented):
Here are some more things I think are worth extra consideration.

extension/src/services/subtitle-sources.ts
```diff
@@ -47,47 +47,6 @@ const parseJsonSafely = (text: string): unknown | undefined => {
...
-    }
-
-    return trustedHtmlPolicy ? trustedHtmlPolicy.createHTML(html) : html;
-};
-
-const parseHtmlDocument = (html: string) => {
-    const trustedHtml = createTrustedHtml(html);
-    return new DOMParser().parseFromString(trustedHtml as string, 'text/html');
-};
-
```
ShanaryS, line 90: These seemed to never have been used, maybe the initial implementation required these.

common/app/components/App.tsx
```diff
@@ -162,8 +162,8 @@ async function extractDropFileHandles(items: DataTransferItemList): Promise<File
 }
 
 function extractSources(files: FileList | File[]): MediaSources {
-    let subtitleFiles: File[] = [];
-    let audioFile: File | undefined = undefined;
```
ShanaryS, line 166: This is never reassigned and thus always undefined. Later on it's value it's being used so I kept it but should this be extracted like the subtitles?

common/annotations/subtitle-annotations.ts
```diff
@@ -109,18 +109,18 @@ export class TrackState {
     }
 
     updateDictionaryTrack(dt: DictionaryTrack) {
-        (this.dt as DictionaryTrack) = dt;
+        (this.dt as any) = dt;
```
killergerbah, line 112: Would it make sense to just make `this.dt` not readonly instead? Or the intention is to keep it readonly by default, to avoid accidental mutations, and use this method to mutate it when necessary?

common/app/components/App.tsx
killergerbah: This is leftover from a refactor that removed handling of audio files using `audio` elements. The intention was to use `video` elements for everything, and so audio and video files would be treated the same way.

I think I just forgot to remove this variable, so you can go ahead and remove it

common/app/components/App.tsx
```diff
@@ -1668,7 +1673,7 @@ function App({
                                 ankiDialogOpen={ankiDialogOpen}
                                 seekRequest={videoPlayerSeekRequest}
                                 onSettingsChanged={onSettingsChanged}
-                                onAnkiDialogRequest={handleAnkiDialogRequestFromVideoPlayer}
+                                onAnkiDialogRequest={(...args) => void handleAnkiDialogRequestFromVideoPlayer(...args)}
```
killergerbah: I think it might not make a huge difference in practice, but it might still be preferable to continue to benefit from `handleAnkiDialogRequestFromVideoPlayer` being a `useCallback` function as a matter of caution, because it seems like it's referenced in a lot of `useEffect` dependency arrays. I think an easy way to do this would be for  `onAnkiDialogRequest` to return `void | Promise<void>` or similar.

common/app/components/App.tsx
```diff
@@ -1682,7 +1687,7 @@ function App({
                                     settings={settings}
                                     lastSelectedExportMode={lastSelectedAnkiExportMode}
                                     onCancel={handleAnkiDialogCancel}
-                                    onProceed={handleAnkiDialogProceed}
+                                    onProceed={(params) => handleAnkiDialogProceed(params)}
```
killergerbah: Similarly here, and in other places, if possible would be good to keep the function references stable if they were stable before.

common/app/components/Controls.tsx
killergerbah: That makes sense, it used to be used by a button but that button no longer exists.

common/app/components/Player.tsx
```diff
@@ -1366,7 +1374,7 @@ const Player = React.memo(function Player({
         }
 
         webSocketClient.onSeekTimestamp = async ({ body: { timestamp } }: SeekTimestampCommand) => {
-            seek(timestamp * 1000, clock, true, true);
+            await seek(timestamp * 1000, clock, true, true);
```
killergerbah: This is a behavioral change, since it would cause the exception to propagate if there were one. I would suggest keeping it as `void` instead, since an exception would currently cause the `webSocketClient` to drop its response and cause the originating request from the server to timeout.

common/app/components/Player.tsx
```diff
@@ -1447,10 +1455,9 @@ const Player = React.memo(function Player({
                             playModes={playModes}
                             onPlay={handlePlay}
                             onPause={handlePause}
-                            onSeek={handleSeek}
-                            onAudioTrackSelected={handleAudioTrackSelected}
+                            onSeek={(progress) => void handleSeek(progress)}
```
killergerbah: As mentioned in other places, I think we should keep the function references stable if it's easy to do so. Just worried about possible performance regressions due to assumptions about the reference being stable.

common/app/components/SubtitlePlayer.tsx
```diff
@@ -262,20 +262,24 @@ const SubtitleScroller = React.forwardRef<HTMLDivElement, ScrollerProps & Contex
...
-        {/* Trailing spacer so the last row clears the controls. */}
-        <tbody aria-hidden="true">
-            <tr>
-                <td style={{ height: 75, border: 0, padding: 0 }} />
-            </tr>
-        </tbody>
-    </Table>
-);
+const SubtitleTable = ({ context, children, ...rest }: TableProps & ContextProp<SubtitleRowContext>) => {
+    void context;
```
killergerbah, line 266: What does `void context` do?

common/app/components/VideoPlayer.tsx
killergerbah: I don't think so, looks like a typo

common/components/About.tsx
killergerbah: Yes can remove, must have tried a few things here and forgot to remove it

common/components/AnkiDialog.tsx
killergerbah: It's so long ago I don't remember the point of these lines, I think it's fine to remove at this opint

common/components/SettingsForm.tsx
killergerbah: Yeah `SubtitleAppearanceSettingsTab` has its own copy

killergerbah (commented):
Thanks for doing this, this was a huge refactor. I forgot there was the codebase was this big.

My main comment is that React components which used to take stable callbacks returned from `useCallback` should probably continue to take stable callbacks.. as I recall having to do this at times for performance reasons

On that note, it seems like `@typescript-eslint/no-misused-promises` is the rule that has caused the most changes..  what types of bugs does it prevent?

extension/src/services/subtitle-sources.ts
killergerbah: I think it was being used to parse HTML of some subtitle repository website. That subtitle source was removed, leaving behind only Jimaku API which is accessed via a JSON API instead. So this is safe to remove.

ShanaryS:
> On that note, it seems like @typescript-eslint/no-misused-promises is the rule that has caused the most changes.. what types of bugs does it prevent?

It prevents using promises when sync code is expected. So if you do `if (func())` but the function returns a promise, it will always be true. Similarly using promises in places where sync is expected can cause other subtle issues or even timing ones as maybe the callback or value needs to be known before the rest of the code. So this combined with `no-floating-promises` ensures that the author has to make an explicit decision about whether the async function can be fire-and-forget, processed with .then/.catch or explicitly awaited.

> My main comment is that React components which used to take stable callbacks returned from useCallback should probably continue to take stable callbacks.. as I recall having to do this at times for performance reasons.

These eslint rules are customizable and you can disable it for JSX attributes on `void`/`Promise<void>`. I'll enable that and revert those changes as we always want to do a floating promise in a UI context.

common/app/components/SubtitlePlayer.tsx
ShanaryS: It's for no-unused variables, it makes the clear the variable isn't being used and is only being pulled out. I couldn't find a cleaner way to do it. We could also just use an ignore for the rule on this line.

common/annotations/subtitle-annotations.ts
ShanaryS: Since the this is only updated on a settings event I prefer keeping it readonly since it's accessed probably hundreds of times and don't want any accidents.

common/app/components/SubtitlePlayer.tsx
killergerbah: Does the variable need to be extracted from the props at all?

killergerbah: Oh, I guess by extracting it there, you prevent it from being passed to `Table`

killergerbah (approved):
Huge effort, thank you
Feel free to merge when you're ready @ShanaryS

common/app/components/SubtitlePlayer.tsx
ShanaryS: Yeah, I don't think it's necessary though. I think we could enable a rule to ignore leading underscores which would make these easier to mark as ignored.

PR #1067, opened 2026-07-10
Do not rewind after playback finished.

### LINKED ISSUE #827: When playback finishes, keep position in the end (rather than looping back to the start)
In the dedicated player mode, when playback finishes, the position / timestamp jumps to the beginning when the video finishes.

I think it'd be a nicer user experience to keep the playback position / timestamp at the end when the video finishes. Simply because there can be something interesting right at the end of the video, and to check it out you'd need to manually seek to the end again

artjomsR:

Closes https://github.com/asbplayer/asbplayer/issues/827

The original logic dates back https://github.com/asbplayer/asbplayer/commit/b129ea8d so not sure if it's still applicable. Personally I think it's more useful to have the playback stop at the end because it makes it easier to mine sentences near the end of the stream. What do you think?

### DIFF
--- a/common/app/components/Player.tsx
+++ b/common/app/components/Player.tsx
@@ -1214,9 +1214,6 @@ const Player = React.memo(function Player({
 
             if (progress >= 1) {
                 pause(clock, mediaAdapter, true);
-
-                await seek(0, clock, true, true);
-                setLastJumpToTopTimestamp(Date.now());
             }
         }, 1000);
 


killergerbah:
I think that makes a lot of sense. I didn't really put a lot of thought into the existing UX, for some reason "go back to beginning" seemed to make sense but it doesn't really help anyone since they still need to interact with the player to start again from the beginning. Also, I don't think most ppl would want to replay anyway...

PR #1073, opened 2026-07-11
fix: restore subtitle order when cues share a start time

### LINKED ISSUE #1064: Netflix subtitles are sometimes merged in reverse order when multiple cues have the same timestamp
### Description

When using asbplayer with Netflix, some subtitles are displayed in the wrong word order.

For example, Netflix displays:

> And I caught a bunch of rocks.

However, asbplayer displays:

> a bunch of rocks. And I caught

The sentence is reversed even though the timing is correct.

---

### Example

Show:
The Thundermans

Episode:
S1E8 - "You Stole My Thunder, Man"

Around:
18:02

Netflix subtitle:

And I caught a bunch of rocks.

asbplayer subtitle:

a bunch of rocks. And I caught

---

### Additional information

Looking at the Subtitle List, both subtitle fragments have exactly the same timestamp:

04:37.069

a bunch of rocks.

04:37.069

And I caught

It appears that when two subtitle cues share the same timestamp, their order is reversed.

---

### Expected behavior

The subtitle should be reconstructed as:

And I caught a bunch of rocks.

---

### Actual behavior

The subtitle is reconstructed as:

a bunch of rocks. And I caught

---

### Browser

Google Chrome

### Streaming service

Netflix

### asbplayer version

(please fill in)

### Screenshots

I'll attach screenshots showing:

1. Netflix displaying the correct subtitle.
2. asbplayer displaying the reversed sentence.
3. Subtitle List showing both cues with the same timestamp. 

Image

Image

Otto-Deviant1904:
Fixes #1064.

`SubtitleCollection.subtitlesAt()` passed `tree.search()` results
straight through. `@flatten-js/interval-tree` orders search results by
(start, end) of its internal keys, not by insertion order, so two cues
sharing a start time can come back reversed if the first one in the
file happens to have a later end than the second. That's exactly the
Netflix case in the issue.

Added a stable sort by `start`, falling back to the cue's original
`index` for ties, so display order matches the source file again.

**Testing**
Added a regression test in subtitle-collection.test.ts. Confirmed it
fails against the pre-fix code (reproduces the exact reversal from the
issue) and passes with the fix. I wasn't able to run the full `yarn
run verify` in this environment (needs Yarn 3/Berry), so please run
that in CI; I did verify the fix and test in an isolated Jest sandbox
against the real source files.

Assisted-by: Claude:claude-sonnet-5


common/subtitle-collection/subtitle-collection.ts
```diff
@@ -64,7 +64,13 @@ export class SubtitleCollection<T extends SubtitleModel> {
 
     subtitlesAt(timestamp: number): SubtitleSlice<T> {
         const interval: NumericTuple = [timestamp, timestamp];
-        const showing = this.tree.search(interval) as T[];
+        // IntervalTree#search orders results by (start, end) of its
+        // internal keys, not by insertion order. Two cues sharing a start
+        // time can therefore come back in the wrong order relative to the
+        // source subtitle file whenever the earlier cue's end is later.
+        const showing = (this.tree.search(interval) as T[]).sort(
```
killergerbah: For now I think it'd be preferable to sort where the subtitles are actually being displayed, since putting the sorting here like this creates a special case just for `showing`. `VideoPlayer.tsx` and `subtitle-controller.ts` are already doing their own sorting so it would be easy enough to piggy back on that.

Otto-Deviant1904: Good call, moved it there instead. Added a shared `compareSubtitlesForDisplay` comparator in common/util and used it in both VideoPlayer.tsx and subtitle-controller.ts's existing sort calls, so SubtitleCollection stays untouched. Added tests for the comparator directly.

killergerbah (approved):
Awesome thanks

buzzshima:
Hi!

I noticed that PR #1073 has already been merged. Thank you for implementing it!

I've been waiting for this feature, so I'm really looking forward to the next Chrome Web Store release.

If possible, I'd really appreciate it if the next extension update could be published soon.

Thanks again for your amazing work!

PR #1077, opened 2026-07-12
Feature: remember playback position

### LINKED ISSUE #847: Remember playback position when reopening videos
When reopening videos in the asb website, the playback always restores to the beginning. It would be great if it remembered the last (approximate) timestamp and suggested resuming from that point

Probably the simplest solution would be to bind the timestamp to the file name, anything more complicated is not warranted

artjomsR:

Closes https://github.com/asbplayer/asbplayer/issues/847

Tried to make this as simple as possible while being most useful. App only (no extension), saves position to local storage on every pause and every 10s, stores maximum of last 50 entries. Autoplays on resume

https://github.com/user-attachments/assets/076b06b1-5e45-41aa-b980-83e4e5914eb4

### DIFF
--- a/common/app/components/Player.tsx
+++ b/common/app/components/Player.tsx
@@ -60,10 +60,8 @@ import { SeekTimestampCommand, WebSocketClient } from '../../web-socket-client';
 import { ensureStoragePersisted } from '../../util';
 import { resolveVideoSubtitleSplitLayout, useVideoAspectRatio } from './video-subtitle-split';
 import { FileWithId } from '../../file-selector';
-import { PlaybackPositionRecord, upsertPlaybackPosition } from '../../global-state';
+import { usePlaybackPositionMemory } from '../hooks/use-playback-position-memory';
 import Alert from './Alert';
-
-const playbackPositionSaveIntervalMs = 10000;
 
 const minVideoPlayerWidth = 300;
 const subtitleCollectionOptions = { returnLastShown: true, returnNextToShow: true, showingCheckRadiusMs: 150 };
@@ -175,8 +173,6 @@ interface PlayerProps {
     hideControls?: boolean;
     forceCompressedMode?: boolean;
     webSocketClient?: WebSocketClient;
-    playbackPositions?: PlaybackPositionRecord[];
-    onPlaybackPositionsChanged?: (positions: PlaybackPositionRecord[]) => void;
 }
 
 const Player = React.memo(function Player({
@@ -223,8 +219,6 @@ const Player = React.memo(function Player({
     hideControls,
     forceCompressedMode,
     webSocketClient,
-    playbackPositions,
-    onPlaybackPositionsChanged,
 }: PlayerProps) {
     const { t } = useTranslation();
     const [playModes, setPlayModes] = useState<Set<PlayMode>>(new Set([PlayMode.normal]));
@@ -239,10 +233,6 @@ const Player = React.memo(function Player({
     subtitlesRef.current = subtitles;
     const settingsRef = useRef(settings);
     settingsRef.current = settings;
-    const playbackPositionsRef = useRef(playbackPositions ?? []);
-    playbackPositionsRef.current = playbackPositions ?? [];
-    const restoredPlaybackPositionFileNameRef = useRef<string>(undefined);
-    const [pendingResume, setPendingResume] = useState<PlaybackPositionRecord>();
     const [subtitleCollection, setSubtitleCollection] = useState<
         SubtitleAnnotations | SubtitleCollection<DisplaySubtitleModel>
     >(SubtitleCollection.empty<DisplaySubtitleModel>());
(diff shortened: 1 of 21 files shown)


common/app/components/Player.tsx
```diff
@@ -1228,6 +1261,48 @@ const Player = React.memo(function Player({
         return () => clearInterval(interval);
     }, [clock, mediaAdapter, seek, tab]);
 
+    useEffect(() => {
```
killergerbah: Can the related callbacks/effects be encapsulated by a single custom hook? That would help keep this component a little cleaner

common/app/components/Player.tsx
```diff
@@ -1401,6 +1476,21 @@ const Player = React.memo(function Player({
 
     return (
         <div onMouseMove={handleMouseMove} className={classes.root}>
+            {pendingResume && (
+                <Alert open={true} onClose={handleDismissResume} autoHideDuration={4000} severity="info" anchor="top">
```
killergerbah: Do you think always showing an alert is most optimal for UX? I can think of some alternatives, such as not providing an option at all when resuming the last session.

common/global-state/index.ts
```diff
@@ -29,13 +42,15 @@ export const initialGlobalState: GlobalState = {
         jimakuSearchCategory: 'anime',
         jimakuRecentWorks: [],
     },
+    playbackPositions: [],
```
killergerbah: Can this be stored directly into local storage instead? `GlobalState` is shared between the app and the extension, which may be useful in the future but it seems harmless and safe to just store directly in LocalStorage for now.

common/app/components/Player.tsx
artjomsR: When I considered design alternatives, the popup seemed the simplest, non-intrusive solution. It only comes up if the user deliberately decides to pick media they were already watching (either opening manually or using `Restore last session`). The standard popup is not intrusive and disappears after a while, so I think it's a good fit for cases where the user would want / not want to resume.
  
I've been driving this branch for less than a week now and am personally content with how it turned out.  
  
I've also considered having something like a recent history overview (maybe before even loading files). Something like the screenshot attached, it's a bookmark / history plugin for mpv. But I think this would be (if ever) a separate implementation aside from the popup, not a replacement. What alternatives were on your mind?

killergerbah: Mostly, I was wondering if someone users might find the new alert as something to just get out of the way, so I was thinking that if we could avoid showing it that would be best. One reasonable way to not show it - just restore the last position automatically when "restoring last session."

However, this is just a UX nitpick about a possibility that may not be true. If you found the UX to be improved then that's a much better indicator that what you have is good.

common/app/services/playback-position-store.ts
```diff
@@ -0,0 +1,32 @@
+const storageKey = 'playbackPositions';
+const maxPlaybackPositions = 50;
+
+export interface PlaybackPositionRecord {
+    fileName: string;
+    position: number;
+}
+
+export function upsertPlaybackPosition(
```
killergerbah, line 9: Nitpick - `upsertPlaybackPosition` already sounds like it should be persisting something. The only usage of `savePlaybackPositions` is immediately following a call to `upsertPlaybackPosition` so maybe you can just merge `savePlaybackPositions` into this function and remove `savePlaybackPositions`.

common/app/components/Player.tsx
artjomsR: Well, this would happen only on resuming playback, so wouldn't happen every time. And I think currently it's a nice balance whether the user wants or doesn't want to interact with the popup:
- User wants to resume: it takes 1 click to resume playback
- User wants to start over: popup disappears after a few seconds 

I think automatically resuming is too intrusive because the user could open a really old video and would prefer to start over. When forced to resume, friction to start over is greater - the user needs to manually drag the slider over to 0:00 which is greater effort than the single click I mentioned above

artjomsR:
All good now

ShanaryS:
I have redesigned the entire playback engine in #1089 which essentially requires this implementation to be rewritten. It would also need to be supported for the extension as part of the refactor is to unify playback logic with clear separation of concerns.

I can easily pull this feature into that PR by taking the work you have done here. For example the new `PlaybackEngine` already does things similar like remembering the playback rate and playback modes.

artjomsR:
I don't mind, happy to hand it over :) @ShanaryS 

Just a note that originally I deliberately decided not to support the extension for this. The reason is that video players normally have their own means to remember playback, so I decided there's no point in fighting for control there.

ShanaryS:
Thanks for the PR, this has been implemented largely the same.

> Just a note that originally I deliberately decided not to support the extension for this. The reason is that video players normally have their own means to remember playback, so I decided there's no point in fighting for control there.

I think it's still beneficial even in this case as not every one of them might and this resume is scoped to the last mining session on this video, not just the last viewing of the video.

PR #1079, opened 2026-07-12
feat: add opt-in Anki card browser refresh after updating a note

### LINKED ISSUE #331: Implement workaround for Anki GUI not updating when updating last card
This would fix the issue mentioned here https://github.com/killergerbah/asbplayer/issues/330 and here https://github.com/killergerbah/asbplayer/issues/316
A workaround (which I had missed before) is mentioned by Aquafina at the bottom of this issue: https://github.com/FooSoft/anki-connect/issues/82

mvanhorn:
## Summary

Adds an opt-in Anki setting that refreshes the Anki Card Browser after asbplayer updates a note, so an open browser window shows the updated card instead of a stale one. When enabled, a successful note update runs the known anki-connect workaround (`guiBrowse` with a deselecting query, then `guiBrowse` with `nid:<updatedNoteId>`); refresh failures are logged and never fail the export itself.

## Why this matters

The reporter in #331 kept the Card Browser open while exporting with "update last card" and had to manually re-query to see the updated fields — the browser does not reload a note that changes underneath it. The maintainer noted the workaround has a macOS side effect (it foregrounds the Anki window), which is why this ships as an opt-in toggle, default off, with help text calling that out. Nothing changes for anyone who leaves the setting off.

## Testing

Added cases to `common/anki/anki.test.ts` covering: refresh runs after `updateLast` and `updateSpecific` when enabled, the two `guiBrowse` calls fire in deselect-then-select order, no refresh happens when disabled, and a `guiBrowse` failure does not fail the export. All 11 tests in the suite pass. The new localization key was added to `en.json` and propagated to every locale file so the `loc-keys-match` check passes.

Closes #331

AI was used for assistance.


common/anki/anki.test.ts
```diff
@@ -31,3 +74,78 @@ it('does not inherit marked up html if already marked up', () => {
...
+describe('Anki browser refresh setting', () => {
+    it('defaults to false and survives settings validation', () => {
+        expect(defaultSettings.ankiRefreshBrowserAfterUpdate).toBe(false);
+
+        const importedSettings = validateSettings(
+            JSON.parse(JSON.stringify({ ...defaultSettings, ankiRefreshBrowserAfterUpdate: true }))
+        );
+        expect(importedSettings.ankiRefreshBrowserAfterUpdate).toBe(true);
+    });
+});
```
ShanaryS: I don't think this needs to be tested, it's only checking `validateSettings` which is already covered elsewhere.

common/locales/en.json
```diff
@@ -363,6 +363,8 @@
         "anki": "Anki",
         "ankiConnectUrl": "Anki Connect URL",
         "ankiConnectApiKey": "Anki Connect API Key",
+        "ankiRefreshBrowserAfterUpdate": "Refresh Anki Card Browser after updating cards",
+        "ankiRefreshBrowserAfterUpdateHelperText": "This works around stale fields in an open Anki Card Browser. On macOS, it may bring the Card Browser window to the foreground.",
```
ShanaryS: This foregrounds the window for me as well on Windows, so I think this can be simplified to:

`This works around stale fields in an open Anki Card Browser. It may bring the Card Browser window to the foreground.`

ShanaryS (commented):
I tested this on Windows and it doesn't seem to be necessary. While it won't open the card browser (which I think is desirable even with it on), if you are already in the browser it gets refreshed. This happens regardless if the updated card is already selected or not. It also of not changing the user's query as well.

mvanhorn:
@ShanaryS thanks for checking Windows. If the open browser already refreshes there, the opt-in is thinner than I thought.

I'll drop the extra validateSettings test and simplify the copy to your wording. Happy to drop the feature entirely if you'd rather not have the toggle — say if you want it gone.

mvanhorn:
@ShanaryS dropped the extra validateSettings test and switched the helper text to your wording (all 13 locale files, so the seed script still treats them as untranslated). Toggle stays; say if you want it gone.

ShanaryS:
I tested it on Linux as well, and it also seems to be updating there. This was Anki `26.08.1` for both Windows and Linux. I assume you're on Mac? Regardless, as long as it's still necessary for you on `26.08.1` then we can add it. If nothing else it focuses the card that was just mined.

Edit: Also there seems to be a conflict that needs to be resolved with the test file now.

mvanhorn:
Tested on macOS 26.3 with Anki 26.08.1 and AnkiConnect. The answer is narrower than I expected.

The card list does refresh on its own, matching what you saw on Windows and Linux. What does not refresh is the editor pane for the note currently selected in the browser.

With the browser open on the note and no `guiBrowse` call at all:

- the list row went from `BEFORE_FR...` to `AFTER_FRO...`
- the editor pane still read `BEFORE_FRONT` / `BEFORE_BACK`, and still did 20 seconds later
- `notesInfo` confirmed the collection already held `AFTER_FRONT_CHANGED` / `AFTER_BACK_CHANGED`

Running exactly what this PR does (`guiBrowse nid:1`, then `guiBrowse nid:<id>`) updated the editor pane to the new values.

So it is still necessary here, but for a narrower reason than the PR description claims, and I will reword that. The practical risk is a stale editor rather than a stale list: if the note you just mined is the selected one, the editor holds the pre-update values, and a manual edit there writes them back over the update.

Happy to keep it opt-in and default off. If you would rather not carry it, say so and I will close.

The test-file conflict you flagged is resolved; the branch is rebased and mergeable.

common/settings/settings.ts
```diff
@@ -637,6 +637,7 @@ export type MediaFragmentFormatSetting = 'jpeg' | 'webm';
 export interface AnkiSettings {
     readonly ankiConnectUrl: string;
     readonly ankiConnectApiKey: string;
+    readonly ankiRefreshBrowserAfterUpdate: boolean;
```
ShanaryS, line 667: This should also be added to `SettingsAccessor` in `use-anki.ts` for it to work in all contexts.

ShanaryS:
> What does not refresh is the editor pane for the note currently selected in the browser.

This does refresh for me on Windows but this option is still clearly needed.

ShanaryS (approved):

PR #1080, opened 2026-07-12
add tests for annotations

ShanaryS:
Now that the design for annotations are more stable, it's worth adding tests to enshrine the various invariants and behavior that the current pipeline depends on. I expect future features to change some of these tests as the goal is that breaking any of these should be intentional especially around the database. It also serves as getting a lot of the knowledge and assumptions I have in my head out concretely so it's easier for others to maintain as well.

I also did some refactoring of the playback mode logic to make it more testable in preparation for future integration with annotations.

**Bug fixes**
- Fix getting last anki note id from sorting
- Fix Anki html tag regex so repeat captures retain the entire tag
- Fix using normalized tokens for `getBulk()` lookup, `anyOfIgnoreCase()` does not work for virtual indexes from compound indexes
- Fix SubtitleCollection gap having a last shown when it precedes the first subtitle
- Fix extension controlled media not preferring repeat mode over condensed like the app

**Other code changes**
- `keysAreEqual()` to only compare an objects own keys rather than inherited as well (no behavior change)
- Toggling a mode off doesn't affect other playback modes (not currently possible)
- Ensure normal play mode is removed when other modes are enabled (not currently possible)
- Extract playback mode logic so it's more testable
- Ensure states are sorted for local tokens when saving
- Explicitly ensure invariants around Anki and WaniKani records when saving

Assisted-by: Codex:GPT-5.6

### DIFF
--- a/common/anki/anki.test.ts
+++ b/common/anki/anki.test.ts
@@ -9,8 +9,7 @@ import {
     NoteInfo,
 } from '@project/common/anki';
 import { AnkiSettings } from '@project/common/settings';
-import { CardModel, Fetcher, MediaFragment } from '@project/common';
-import { AudioClip } from '@project/common/audio-clip';
+import { CardModel, Fetcher } from '@project/common';
 import { extractText, sourceString } from '@project/common/util';
 import { afterEach, describe, expect, it, jest } from '@jest/globals';
 
@@ -117,21 +116,34 @@ const makeSubtitle = (text: string, track = 0) => ({
     track,
 });
 
-const makeAudioClip = (overrides: Record<string, unknown> = {}) =>
-    ({
+const makeAudioClip = (overrides: Record<string, unknown> & { base64Result?: string; base64Error?: Error } = {}) => {
+    const media = {
         name: 'clip:name?.mp3',
         error: undefined,
-        base64: jest.fn<() => Promise<string>>().mockResolvedValue('audio-base64'),
+        base64Reads: 0,
+        base64: async () => {
+            media.base64Reads++;
+            if (overrides.base64Error) throw overrides.base64Error;
+            return overrides.base64Result ?? 'audio-base64';
+        },
         ...overrides,
-    }) as any;
+    };
+    return media as any;
+};
 
-const makeImage = (overrides: Record<string, unknown> = {}) =>
-    ({
+const makeImage = (overrides: Record<string, unknown> & { base64Result?: string } = {}) => {
+    const media = {
         name: 'image:name?.jpeg',
         error: undefined,
-        base64: jest.fn<() => Promise<string>>().mockResolvedValue('image-base64'),
+        base64Reads: 0,
+        base64: async () => {
+            media.base64Reads++;
+            return overrides.base64Result ?? 'image-base64';
+        },
         ...overrides,
-    }) as any;
+    };
+    return media as any;
+};
 
 const makeExportArguments = (overrides: Partial<ExportArguments> = {}): ExportArguments => ({
     text: 'line 1\nline 2',
@@ -229,11 +241,8 @@ it('preserves target markup instead of replacing it with shallower source markup
 });
 
 it('maps CardModel data into export params in exportCard', async () => {
-    const exportSpy = jest.spyOn(Anki.prototype, 'export').mockResolvedValue('stored-word');
-    const audioClip = { kind: 'audio' } as any;
-    const image = { kind: 'image' } as any;
-    const audioFromBase64Spy = jest.spyOn(AudioClip, 'fromBase64').mockReturnValue(audioClip);
-    const imageFromBase64Spy = jest.spyOn(MediaFragment, 'fromBase64').mockReturnValue(image);
+    const fetcher = new MockFetcher();
+    fetcher.fetch.mockResolvedValue(ankiConnectResponse('stored-word'));
     const card = makeCardModel({
         text: undefined,
         mediaTimestamp: 0,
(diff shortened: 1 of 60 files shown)


common/anki/anki.test.ts
```diff
@@ -1,33 +1,1495 @@
...
+            extension: 'mp3',
+            paddingStart: 0,
+            paddingEnd: 0,
+        },
+    });
+
+    const result = await exportCard(card, testAnkiSettings, 'gui');
+
+    expect(result).toEqual('stored-word');
+    expect(audioFromBase64Spy).toHaveBeenCalledWith('episode.ass', 1000, 2000, 1, 'audio-base64', 'mp3', undefined);
```
killergerbah: These kinds of tests that verify implementation do their job by making it harder to introduce a regression but they also make it hard to change the implementation because they essentially copy parts of the implementation into tests.

Is it possible to write these so that they mostly test inputs and outputs?

common/anki/anki.ts
```diff
@@ -572,7 +572,7 @@ export class Anki {
             case 'gui':
                 return (await this._executeAction('guiAddCards', params, ankiConnectUrl)).result;
             case 'updateLast': {
-                const lastNoteId = [...recentNotes].sort()[recentNotes.length - 1];
+                const lastNoteId = [...recentNotes].sort((a, b) => a - b)[recentNotes.length - 1];
```
killergerbah, line 576: Wow, feels like this should have been a more noticeable bug

common/annotations/annotations-test-utils.ts
```diff
@@ -0,0 +1,154 @@
...
+    requestStatisticsGeneration: jest.fn(),
+    onRequestStatisticsGeneration: jest.fn().mockReturnValue(jest.fn()),
+    requestStatisticsSeek: jest.fn(),
+    onRequestStatisticsSeek: jest.fn().mockReturnValue(jest.fn()),
+    requestStatisticsMineSentences: jest.fn(),
+    onRequestStatisticsMineSentences: jest.fn().mockReturnValue(jest.fn()),
+    _removeCallback: jest.fn(),
+});
+
+export const makeSettingsProvider = (settings: AsbplayerSettings) =>
```
killergerbah: I don't think it'll make a huge difference in practice, but I do think it is preferable to prefer real implementations/mocks over mocks produced by mocking libraries since you'll get more incidental code coverage. Also, mocking libraries make it easy to "cheat," since being able to mock anything allows you to write code that's hard to test otherwise. Although for asbplayer it makes sense to use it a lot since almost nothing was written to be testable.

For `SettingsProvider` you should be able to use `new SettingsProvider(new MockSettingsStorage())` instead, which would give you some incidental coverage of `SettingsProvider` as well.

common/annotations/subtitle-annotations.test.ts
```diff
@@ -0,0 +1,767 @@
...
+        expect(resetCache).toHaveBeenCalledTimes(1);
+        expect(trackState.yt).toBeUndefined();
+        expect(trackState.ytLastResetAt).toBe(1234);
+    });
+});
+
+describe('SubtitleAnnotations', () => {
+    it('defaults originalText, clones subtitles, preserves cached tokenization, and stores external readings', () => {
+        const { subtitleAnnotations } = makeSubtitleAnnotations();
+        const buildAnnotations = jest.spyOn(subtitleAnnotations as any, '_buildAnnotations').mockResolvedValue(true);
```
killergerbah, line 99: Having the test in the first place is more important of course, but mocking private methods is a sign that the private method could be extracted to a different class and become a public method

killergerbah (commented):
Nice, it's good to have tests over some very old code

Was a lot of this Codex-generated? I have found that AI tends to overuse mocking libraries, maybe because it's an easier alternative to refactoring?

ShanaryS:
> Was a lot of this Codex-generated? I have found that AI tends to overuse mocking libraries, maybe because it's an easier alternative to refactoring?

It's a mixture of a lot, I've been working on the tests on and off since May. The AI did things like separating the tests from the `dictionary-db` refactor and writing similar tests for WaniKani as for Anki. It's more so that I was avoiding refactors as I know how subtle the issues are and how much though it took to verify all the pipelines.

My main goal is to preserve a lot of small but important things at different stages and so mocking made it the easiest to do. I didn't want to change the actual code too much as we know it works and wanted to keep the code and the tests separate.

I think I can still do those while making the tests more input/output oriented. I'll update the non annotations ones to be more like that but I think I'll leave the annotations ones for now. I'll gradually refactor those as the implementations change, basically using the tests as written as hard docs. Since rewriting them now won't really offer any benefit and I'm planning more refactors for the upcoming features.

common/anki/anki.test.ts
ShanaryS: Yeah I'll update the non annotations ones to be more input/output oriented but I'll slowly update the annotations tests as I'm expecting some refactors for the next group of features.

common/anki/anki.ts
ShanaryS: The query for this was `added:1` so only cards from the past day would be sorted. In that case all of the leading digits would be the same and the others would increase quickly since it's based on timestamp. I think the nature of this usage meant the string sorting didn't really matter.

common/annotations/subtitle-annotations.test.ts
ShanaryS: This function was on my list to break up or put in its own class like I did with the other refactors. But think would be hard without introducing some bugs as it manages so much state. I wanted to at least have a strong base before any significant refactors.

extension/src/services/binding.ts
```diff
@@ -384,6 +384,7 @@ export default class Binding {
                         if (
                             this.recordingMedia ||
```
killergerbah, line 410: Do you think it would be easy to use the play mode effects here? I think that would obviate the need for binding tests to cover the play mode logic

killergerbah (commented):
Thanks, the codebase needed tests
As this is mostly test code I don't think it's critical to address any of the comments. The main one is that there is a lot of mocking which could be replaced by refactoring, like the play mode effect refactor you did. Understandably this is difficult for sensitive code that's already working.

killergerbah (approved):

extension/src/services/binding.ts
ShanaryS: I've currently completely refactored the playback modes so that the implementation is consistent between Binding and Player. I basically got rid of all the binding tests except for some specific binding behavior. Though since this is a big refactor I need to spend some time testing it out and simplifying the design as much as possible.

PR #1081, opened 2026-07-12
display subtitles with negative timestamps

ShanaryS:
`react-virtuoso` requires the elements to always have a height to work properly. So with the switch to that it uncovered an existing bug, offsets can make the subtitle timestamp negative. A check in `SubtilePlayer` would return `null` for such rows triggering the crash.

This occurred due to the `lastSubtitleOffset` being negative and getting applied to a new subtitle that starts at `0ms` or less than the offset value.

~~We will now clamp the offset so that the earliest timestamp is `0ms` and the latest timestamp cannot be past the video length.~~

I reverted the original design and ended up simply allowing negative events to be displayed but clamping the seek to within the media length. I think displaying the negative event may even better as context for the offset rather than dropping it.

Assisted-by: Codex:GPT-5.6

### DIFF
--- a/common/app/components/Player.tsx
+++ b/common/app/components/Player.tsx
@@ -31,7 +31,7 @@ import { SubtitleCollection } from '@project/common/subtitle-collection';
 import { HoveredToken, SubtitleAnnotations } from '@project/common/annotations';
 import { SubtitleReader } from '@project/common/subtitle-reader';
 import { KeyBinder } from '@project/common/key-binder';
-import { surroundingSubtitles, timeDurationDisplay } from '@project/common/util';
+import { clampMediaTimestamp, surroundingSubtitles, timeDurationDisplay } from '@project/common/util';
 import BroadcastChannelVideoProtocol from '../services/broadcast-channel-video-protocol';
 import ChromeTabVideoProtocol from '../services/chrome-tab-video-protocol';
 import Clock from '../services/clock';
@@ -278,10 +278,11 @@ const Player = React.memo(function Player({
                 resetPendingAutoRepeatTargetTimestamp();
             }
 
-            clock.setTime(time);
+            const clampedTime = clampMediaTimestamp(time, (channelRef.current?.duration ?? 0) * 1000);
+            clock.setTime(clampedTime);
 
             if (forwardToMedia) {
-                await mediaAdapter.seek(time / 1000);
+                await mediaAdapter.seek(clampedTime / 1000);
             }
 
             autoPauseContextRef.current?.clear();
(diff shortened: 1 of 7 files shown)


common/app/components/SubtitlePlayer.tsx
```diff
@@ -358,7 +358,6 @@ const SubtitleRowCells = React.memo(function SubtitleRowCells({
 }: SubtitleRowCellsProps) {
     const classes = useSubtitleRowStyles();
     const { t } = useTranslation();
-    if (subtitle.start < 0 || subtitle.end < 0) return null;
```
ShanaryS, line 361: This is not necessary for this fix but I don't think we should have it due to how `react-virtuoso` treats it.

killergerbah:
If I'm understanding the change correctly, it will ensure that all subtitles have valid timestamps? I.e. you can't offset a subtitle off of the timeline? I think that makes sense and it fixes the virtuoso issue.

How would we deal with situations where it's desirable to offset the subtitle off the timeline? For example, subtitles are often misaligned by the length of a commercial break which can be 1-2 minutes. To make the subtitles usable in this case, users would need to offset the subs by 1-2 minutes which would likely push the starting subtitles to negative offsets. I think that's the `return null` existed in the first place. As an alternative, could filtering be applied right before passing the subtitle list to virtuoso?

ShanaryS:
> If I'm understanding the change correctly, it will ensure that all subtitles have valid timestamps? I.e. you can't offset a subtitle off of the timeline?

Yes.

> How would we deal with situations where it's desirable to offset the subtitle off the timeline?

My thinking was that these subs would be too bad to be usable or would constantly need adjusting as if they were meant for a different version of the media. But now I'm realizing that it may be more common and normal than expected. Especially if there is a subtitle at the start and the end of the media, then it wouldn't allow shifting by even 1s.

I previously had a speculative fix where I just prevented the zero heights from being possible but since I didn't really know the cause I could really test it. But now I think that's the correct approach by filtering before it reaches the virtualization stage.

ShanaryS:
I reverted the original design and ended up simply allowing negative events to be displayed but clamping the seek to within the media length. I think displaying the negative event may even better as context for the offset rather than dropping it.

Filtering before it reaches virtuoso required maintaining the new indexes in a map and to reference that whenever we need the indexes.

common/app/components/Player.tsx
```diff
@@ -278,10 +278,11 @@ const Player = React.memo(function Player({
                 resetPendingAutoRepeatTargetTimestamp();
             }
 
-            clock.setTime(time);
+            const clampedTime = clampMediaTimestamp(time, (channelRef.current?.duration ?? 0) * 1000);
```
killergerbah, line 281: `channelRef.current` won't be defined if a video element isn't connected, I think you want `calculateLength`

killergerbah (approved):

common/app/components/Player.tsx
ShanaryS: I can switch it but it shouldn't matter. If it's 0 then it won't add an upper bound to the seek but it shouldn't matter without a video.

killergerbah: Ah I see, makes sense for it to be intentional

PR #1083, opened 2026-07-14
Fix/stremio subtitle detection

### LINKED ISSUE #800: Subtitle detection not working on web.stremio.com

After the recent enhancement for subtitle detection on Stremio, subtitles are not being detected.
When playing a video on web.stremio.com, asbplayer shows “no subtitles available” even though subtitles are enabled in Stremio from the start.
I simply opened a Stremio stream with subtitles turned on, no manual toggling or late enabling, and asbplayer failed to detect them.

**What browser?**

Google Chrome 141.0.7390.55 on Windows 10

Streaming video

asbplayer Chrome extension Extension version [1.12.0]

web.stremio.com

Image

eXaminator:

This PR improves the Stremio integration in three ways:

1. It improves the subtitle detection, including fetching all available subtitles from the installed addons that provide subtitles. This works for me using AIOStreams which provides subtitles from OpenSubtitles. It should also improve the old detection by beeing less strict with the URLs subtitles are being loaded from when it comes to the fetch interceptor. This works for me, but obviously I'm limited in what I can test and would love more people to give it a go (especially with different subtitle sources etc.). This should fix #800 (or hopefully at least improve it a lot).
2. I also improved the title detection, because Stremio has a fixed document title. So instead it looks for the proper element on the page.
3. I made sure the font size for subtitles with annotations are inherited, because Stremio adds styles that take priority for some reason that lead to the subtitles getting a lot smaller once annotations are enabled.

As noted in the commits, I used claude opus 4.8 to make these changes. They look fine to me, but let me know if there's anything you want changed. I'm happy to help to get this over the finish line to allow learners like me who would like to use Stremio to get the best possible experience :)

I was also only able to test this in Firefox so far, but I don't see why any of these changes should work differently on Chrome.


extension/src/entrypoints/stremio-page.ts
```diff
@@ -1,62 +1,415 @@
...
+                    detail: {
+                        error: '',
+                        basename: videoName(),
+                        subtitles: undefined,
+                    },
+                })
+            );
+
+            // A single pending timer; repeated requests just reset it.
+            if (emptyResultTimer !== undefined) {
```
killergerbah, line 399: Why was this needed? I think the spinner in the asbplayer UI should timeout on its own?

killergerbah:
@eXaminator Thanks for doing this. The code looks roughly ok. Can you upload a video of the integration working on Firefox and Chrome?

extension/src/entrypoints/stremio-page.ts
eXaminator: Ok, here is Claudes reasoning for this, which I relied on:

> The asbplayer UI doesn't currently time the spinner out on its own — isLoading comes straight from the synced data (isLoading: this._syncedData.subtitles === undefined in _buildModel), and VideoDataSyncUi only flips it from model.isLoading. There's no timer in the controller or the dialog, so if the page reports subtitles: undefined and never sends a final array, the spinner never stops.
> 
> We report undefined on purpose here: Stremio detection is async (we rebuild the subtitle manifest by querying the installed addons after the request), and reporting an empty [] up front would count as a real result and consume the one-shot auto-sync — the auto-sync block in _setSyncedData is gated on subtitles !== undefined — firing it against zero tracks before the reconstructed ones arrive. So the page has to resolve the loading state itself: emit the tracks once found, or an empty array after the timeout when nothing turns up.
> 
> The bit you're commenting on didn't add the timeout, it just collapses it into a single resettable timer, since asbplayer-get-synced-data can fire repeatedly and each request was otherwise stacking a new one.

Does that make sense to you?

eXaminator:
Here are the videos. While creating them I noticed a small bug that I fixed in the last commit.

I can't upload videos larger than 10 MB here, so I uploaded them to proton. I hope this works:

[Firefox](https://drive.proton.me/urls/62PCWFMF3C#eKafg98j2sWT)
[Chrome](https://drive.proton.me/urls/7PWF9TERBW#UvthsgeGOk0u)

extension/src/entrypoints/stremio-page.ts
killergerbah: Makes sense, it is a way to debounce multiple requests
I was wrong about the timeout, that is up to each page script

killergerbah:
Great work, thank you @eXaminator

PR #1086, opened 2026-07-16
Forward detected episode to Jimaku files API to auto-locate the current episode's subtitles

xwxb:
## Summary

Opening this as a **draft** to use it as a discussion issue — I'm mainly looking for your thoughts on the feature direction itself, not on the implementation yet. I'll keep comparing a few implementation approaches on my own and request a real review once it's ready.

### The gap this tries to close

For continuous watching on a streaming site, the current flow already does a lot: a hotkey opens the subtitle dialog, recent works are cached, and the detected title is auto-filled into the search box. But after picking an entry, you still have to manually find the file for the *current episode* inside the file list — that step repeats every episode.

### What this PR does (roughly)

It parses the current episode number from the page-script basename hint and forwards it to the Jimaku files API (`?episode=N`), which the backend already supports via `anitomy` (including anilist relation mapping and movie skipping). So the file list comes back already filtered to the current episode. When no episode can be confidently detected, or the filtered result is empty, it silently falls back to all files (current behavior).

### What I'd like feedback on

Just the direction: is auto-locating the current episode's file worth adding at all, and does "parse from the page title, forward to the existing `episode` API param" feel like the right shape for it? No worries if it doesn't fit — happy to drop or rework it. Implementation details (and a possible follow-up for user-customizable patterns) I'll bring back when I request a review.


killergerbah:
Thanks for opening the discussion

> But after picking an entry, you still have to manually find the file for the current episode inside the file list — that step repeats every episode.

There may be other solutions, like scrolling to/highlighting the last file that was picked

> is auto-locating the current episode's file worth adding at all

I think so, since it sounds like you're a heavy user that finds it desirable

> , and does "parse from the page title, forward to the existing episode API param" feel like the right shape for it?

IMO this depends on how accurate the episode extraction is. If it's possible for it to be inaccurate, then it seems more like a liability, because either the wrong episode will be queried, or another API request will be used fetch the whole list as a fallback.

The extra API request could be solved if you could always fetch everything, and then choose the right episode from an in-memory list, but it sounds like that would require a Jimaku API update.

On the other hand, if the detection is extremely accurate then there is no concern.

xwxb:
@killergerbah 

Thanks for the feedback. About still needing to pick an entry — I don't think that's really an issue. Once the episode number is locked in, it's just a matter of choosing a subtitle track, and most of the time users will pick the one that matches their video source anyway. Automating that further feels tricky since the first result isn't necessarily the best fit, so I'd lean toward leaving that part as is.

Also, you mentioned "If it's possible for it to be inaccurate, then it seems more like a liability" — the API's episode parameter and match algorithm is actually reliable enough. I've done some quick testing and it works fine. that's the original reason i started working with this feature 

Anyway, glad you're on board with the direction. If nothing else, I'll spend another times refining the implementation and then have you take a look at the actual code

xwxb:
Quick update: I ran the `episode` param against a spread of real series to confirm it actually returns results in practice, not just the two examples.

**Bulk probe — 8 series × 3 episodes (ep 1/6/12):**
- Hit rate (non-empty): 22/24. The 2 empties were a movie entry probed for ep6/ep12 (out of range) — wouldn't happen in real use, since movie titles have no episode marker to parse.
- Reduction: avg ~0.08x — the file list drops to ~8% of full, from 100+ down to single digits for long series.

**Accuracy/recall (manual, on the two examples):**
- Frieren S2 E2 (120 files): filter returned 12, all correct, none missed — including the anilist-relation equivalents where S2E2 is numbered as absolute 30.
- Rascal Does Not Dream of Santa Claus E3 (69 files): filter returned 5, all correct, none missed.

So: ~100% hit in real scenarios, ~92% list reduction, 100% precision/recall on the checked cases, and the empty-result fallback path is covered. The feature holds up.

---

entry | total | ep1 | ep6 | ep12
-- | -- | -- | -- | --
Sousou no Frieren | 305 | 12 | 11 | 11
SPY×FAMILY | 72 | 6 | 6 | 6
Kaguya First Kiss | 7 | 4 | 0 | 0
Jujutsu Kaisen | 146 | 6 | 6 | 6
Boku no Hero Academia | 106 | 10 | 9 | 9
Kimetsu no Yaiba | 131 | 6 | 5 | 5
Shangri-La Frontier | 112 | 4 | 4 | 4
Bocchi the Rock! | 95 | 9 | 8 | 7

killergerbah:
@xwxb Sorry, I have been focusing on other PRs and accidentally skipped over this one since it's in draft. In case you were expecting a response, if the regex is working then the approach seems fine

xwxb:
@killergerbah Never mind, the previous reply is fine already. I've also been putting this on hold for a while — when I first started implementing it, I found it hard to write something I was satisfied with. It might need some refactoring, and the changes could be quite significant. So I stepped back to reconsider whether it's worth making things that complex again.

But as it happens, I had some time the day before yesterday to revisit this and think about whether this feature is a good fit for our tool. My conclusion is that I still want to do it — after all, the functionality is basically scoped within the online subtitle module, and there is an existing API for it. It's also genuinely a recurring pain point. So I'm planning to just go ahead and implement it. Once I've determined the scope of changes, if it turns out to be really large, I'll come discuss it with you.

xwxb:
@killergerbah 
Ready for review — this is a rework of the earlier draft. The previous version silently dropped the episode as soon as the user edited the search box, and detection / query cleaning lived in two disconnected functions, so `SxxExx / EP / 第N话` markers were detected but stayed in the query and polluted the search.

Now: detection and query cleaning happen in one pass, so the search query is always clean; the episode is resolved once when the dialog opens and stays fixed for its lifetime, decoupled from the query. The episode is forwarded only to `entries/{id}/files?episode=N` when an entry is selected — it never touches the first-layer search, the cache key, or category auto-research, all identical to main.

User-facing behavior:
- No episode parsed → full list, no chip — identical to main.
- Episode parsed, non-empty result → list with an `EP N` chip next to the entry name.
- Episode parsed, empty playable result → `console.warn` + automatic refetch of all files, chip hidden. The user never faces an empty "no subtitles" list.
- Click `×` on the chip → refetch all files for the current entry only.

Two deliberate decisions worth flagging:

1. **The episode stays fixed for the dialog session rather than being re-derived from the query.** Cross-entry searches keep it, which can look surprising at first glance. We chose this over the old silent-drop because the chip makes the filter visible, `×` makes it clearable, and the empty-result fallback guarantees no regression — strictly not worse than silently losing it.
2. **Failures degrade silently to the full list.** Parse failure → no param, full list, same as main; episode filter returns nothing → warn + refetch all. The feature only ever adds; nothing becomes worse than today.

Both are cheap to revisit — the chip can grow into an editable input, and the fallback policy is a one-line condition. Happy to adjust based on your feel.

killergerbah:
Thankx @xwxb looks really good

xwxb:
Hi @Rapptz, this PR adds episode-aware subtitle lookup to asbplayer, which makes additional requests to Jimaku’s episode files endpoint (entries/{entry_id}/files?episode={episode}). since now asbclient not add ua to req (as discussed in #957) , i may let you know we make another API usage (centainly with api key).
And the `X-Client` plan up to @killergerbah.

Rapptz:
Hi.

Just so you know, the anitomy library I wrote is also usable within javascript as a wasm module. That way you don't have to do the brittle parsing using regex. Maybe at some point in the future it'd be nice to have a single API call that takes in a filename and then returns the entry + files for that filename but that is a long ways away. Either way, the extra `X-` header for user-agent might be worth considering though I'm unsure which one that'd be.

xwxb:
@Rapptz 

The library and future API idea are cool. But this PR is about the main use case here is online video, where we're parsing browser tab titles rather than filenames, so anitomy doesn't really help there. 

For local files, most people already download subtitles alongside the video, so the local video workflow isn't really connected to Jimaku's online subtitle search at this point yet.

That said, a local mode where you could parse a filename and directly get a list of matching subtitle files would be a nice experience down the road. But it's another issue, that depends on where asbplayer is headed — I think this can be discussed together with the X-Client header issue once @killergerbah weighs in.

Rapptz:
I don't see how the anitomy doesn't help when it parses literally everything mentioned in the tests just fine. Either way, it's your tool and your decision. I was just trying to point you to a different direction that didn't involve maintaining a bunch of regex. 

I don't know what you mean by local mode. The API endpoint idea is just to automate this flow entirely, especially for those with e.g. mpv scripts or something. It also didn't really pertain to asbplayer directly, just a general idea of something I want to work on in the future that might end up helping here since the parsing is fundamentally the same.

xwxb:
@Rapptz 
What I actually want to clarify is that the main parsing scenario for jimaku is only online streaming currently — the parsing target is browser tab titles like `Show Name - Episode 1 | Crunchyroll`. Anitomy is definitely an awesome approach, and it's the first one I've come across besides regex (it's actually a good takeaway for me).

But imo, anitomy is filename-specific by design intent. If I wanted to apply something similar to the current scenario, I'd probably need to build a different algo for it. I'll give it more thought.

Rapptz:
It's not inherently filename specific, the way it works is by parsing components that it attempts to find. I don't wanna derail this further, but like, using this as an example:

```js
> parse('Seihantai na Kimi to Boku 2nd Season - Episode 1 | Crunchyroll')
< Object { title: "Seihantai na Kimi to Boku", season: "2", episode: "1", source: "Crunchyroll" }
``` 

or really any combination, should show what I'm talking about. I don't own a crunchyroll subscription to actually test this with a browser tab but I don't see how this would matter fundamentally speaking. Either way, I'm sorry for derailing this.

xwxb:
@Rapptz 
that actually works better than I expected for tab titles. I'll take a closer look at it then, giving it a try and compare them, thx again.

killergerbah:
@Rapptz How does `X-Client-ID: asbplayer-extension` sound to you?

Rapptz:
@killergerbah that sounds fine to me

PR #1088, opened 2026-07-17
[codex] Add Bilibili mainland subtitle support

Dr-TNineS:
## Summary

- add subtitle detection for desktop `www.bilibili.com/video/*` pages while preserving `bilibili.tv` support
- parse mainland manual and AI subtitle metadata, including protocol-relative subtitle URLs
- observe both `JSON.parse()` and `Response.json()` results and isolate tracks by video path and multipart `p` value
- document support for the mainland site and its login-dependent subtitle availability

> [!NOTE]
> Bilibili mainland may expose subtitle track metadata only to **signed-in** users. When signed out, `data.subtitle.subtitles` may be omitted, so asbplayer will not detect any subtitle tracks. This is expected behavior: the integration reuses the page's current authenticated session and does not bypass Bilibili's access controls.

## Root cause

The existing Bilibili adapter only matched `www.bilibili.tv` and understood its `data.subtitles` response shape. The mainland site exposes tracks under `data.subtitle.subtitles` with different field names, and the relevant data may be consumed through `Response.json()` instead of a direct `JSON.parse()` call.

## User impact

Users signed in to the desktop mainland Bilibili site can select detected manual or AI subtitle tracks, load the subtitle list, and use the existing copy and card-creation flows. Videos without accessible subtitle metadata continue to return an empty track list.

## Screenshot

Bilibili subtitle tracks are detected in the selector, and the selected Japanese track is successfully loaded into the asbplayer subtitle list.The screenshot below was captured while **signed in** to Bilibili.

Bilibili mainland subtitle detection

## Validation

- 22 Jest suites passed (109 tests total) across common, client, and extension workspaces
- ESLint passed for `common`, `extension/src`, and `client/src`
- Prettier passed for all files changed by this PR
- extension TypeScript compilation passed
- Chromium MV3 development build passed
- locale-key, locale-code, and page-script reference checks passed via equivalent PowerShell checks
- manually verified subtitle loading and card creation on `https://www.bilibili.com/video/BV1sM4y1V7x1/`

The root `yarn verify` wrapper could not run directly in this Windows checkout because its Bash helpers require WSL/`jq`, and the checkout uses CRLF for untouched files. All constituent checks relevant to this change were run separately; the upstream Ubuntu workflow will run the canonical verification.

### DIFF
--- a/common/subtitle-reader/subtitle-reader.test.ts
+++ b/common/subtitle-reader/subtitle-reader.test.ts
@@ -150,3 +150,24 @@ describe('SubtitleReader dfxp timestamp handling', () => {
         expect(subtitles).toHaveLength(0);
     });
 });
+
+describe('SubtitleReader Bilibili JSON parsing', () => {
+    it('parses Bilibili JSON cues', async () => {
+        const file = {
+            name: 'test.bbjson',
+            text: async () =>
+                JSON.stringify({
+                    body: [
+                        { from: 1.25, to: 3.5, content: 'First subtitle' },
+                        { from: 4, to: 6.75, content: 'Second subtitle' },
+                    ],
+                }),
+        } as unknown as File;
+
+        const subtitles = await createReader().subtitles([file]);
+
+        expect(subtitles).toHaveLength(2);
+        expect(subtitles[0]).toMatchObject({ start: 1250, end: 3500, text: 'First subtitle' });
+        expect(subtitles[1]).toMatchObject({ start: 4000, end: 6750, text: 'Second subtitle' });
+    });
+});
--- a/docs/docs/compatibility.md
+++ b/docs/docs/compatibility.md
@@ -42,7 +42,7 @@ sidebar_position: 7
 | Rakuten Viki            |                                                                                            ✓                                                                                             |
 | osnplus                 |                                             Compatibility with osnplus is currently unknown. Reach out if you have more information on this.                                             |
 | Plex                    | Supports external subtitles. As for internal subtitles, first select them from Plex UI to make them selectable from asbplayer. Configure custom domains from the page-specific settings. |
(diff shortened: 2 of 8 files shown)


killergerbah:
Sorry its taken me so long to look at this. Is this still working for you? I just tried testing on [bilibili.tv](https://www.bilibili.tv/en/play/1063218?bstar_from=bstar-web.homepage.recommend.all) and it's now downloading an `ass` file instead of the `json` that it used to

Dr-TNineS:
Thanks for taking a look! I rechecked the exact `bilibili.tv` title today. You're right that the public subtitle list now points directly to ASS files for most languages.

For season `1063218`, episode `11511707`, the current `/intl/gateway/web/v2/subtitle` response contains direct `data.subtitles[].url` values ending in `.ass` for most tracks (with one JSON track), while `data.video_subtitle` still contains both the older JSON `srt.url` and the newer `ass.url` variants.

The PR already reads `data.subtitles[].url` and derives the subtitle format from the URL extension, so the current ASS response is detected as `ass`; no production-code change was needed. I fetched the English ASS file and parsed it with asbplayer's existing ASS parser successfully (484 dialogue lines). I added regression coverage for this current response shape in cfd163c.

Validation after that commit:
- 39 Jest suites passed (580 tests total)
- extension TypeScript compilation passed
- ESLint passed
- Chromium MV3 development build passed
- the changed test file passes Prettier

For the mainland sample, the public API still exposes seven subtitle tracks but leaves `subtitle_url` empty when signed out, which is consistent with the login-dependent behavior documented in the PR. I wasn't able to complete a fresh signed-in UI check from this environment, so I don't want to overstate that part of the live verification.

killergerbah:
Thanks
The integration is still not working for me on bilibili.tv (logged in or not) so I will try again later and maybe debug a bit

PR #1089, opened 2026-07-19
Redesign playback engine and playback modes

### LINKED ISSUE #215: Auto-pause ms offset preference

### LINKED ISSUE #367: Fast-Forward Mode Minimal Interval Option
There doesn't appear to be any options related to Fast-Forward mode.
Condensed Playback has a setting for minimal skip interval. And I think the same option for Fast-Forward would be very useful. (Or at least minimally a toggle to set the Interval to be extra low or zero).

As a related part of the feature. When Fast-Forward mode is turned on it seems to automatically set the playback speed to a default of x1. (Both for other plugins and for Asbplayer's own speed controls).
It might be more niche but I think it would be good if users could control the base speed -- along with choosing a faster speed for Fast-Forward.
Again, probably too niche to spend time in order to have the option to set a not-Fast-Forwarding / while-subtitles-are-active speed option. But it not forcing 1x speed, when Fast-forward mode is enabled, would be good.

Thank you for reading and sorry if either of these were already covered somewhere else.

### LINKED ISSUE #750: Playback rate notification pops up on every pause/unpause when a different extension controls the speed
Today out of nowhere ASBPlayer started showing the "playback rate" popup every time I pause or unpause the video (which is very annoying, needless to say), which is only happening when the extension Video Speed Controller (used for, well, controlling video speed) is enabled. If not a fix, then at least an option to turn this notification off would be very appreciated.

### LINKED ISSUE #784: Auto-pause at the start and the end
add option to enable both
Image

### LINKED ISSUE #847: Remember playback position when reopening videos

### LINKED ISSUE #960: some streaming website has white block on overlay when pause

>go to https://www.miruro.to/
>watch an anime
>pause
>have white block overlay

Image

**What browser?** 
brave
streaming video 
v1.15.0
the youtube.com is fine.https://animepahe.pw/ is fine.https://kaa.lt/ is fine.https://animetsu.net/ is fine
other website have this issue i tested are:
https://allmanga.to/
https://animepahe.pw/

### LINKED ISSUE #1091: JPDB Reader with Version 1.19 only parses for one minute
**Before submitting an issue check out the common issues section of the docs:**
https://docs.asbplayer.dev/docs/common-issues

Before the 1.19.0 update, the JPDB Reader extension would parse the full video/movie. Since the update, I noticed that it only parses a minute worth, from when the parsing happens. I downloaded the previous version, which parses the whole thing, but Netflix subtitles don't show up. I don't know if this is a bug with JPDB Reader, but downgrading asbplayer fixed it, so I'm assuming it's a change in asb.  

**What browser?**

Tried with both Vivaldi and Firefox (both on on Endeavour and Fedora) 

Streaming

1.19.0

Youtube and Netflix 

``` 
The resource <URL> was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
video.js:61 Anki permission request failed: Error: Failed to fetch
    at e._executeAction (video.js:61:2563)
    at async e.requestPermission (video.js:60:14220)
    at async qs._refreshAnki (video.js:61:66484)
_refreshAnki @ video.js:61
await in _refreshAnki
(anonymous) @ video.js:61
video.js:61 Error checking Anki recently modified cards: Error: Anki not initialized
    at qs._checkAnkiRecentlyModifiedCards (video.js:61:65437)
    at qs._refreshAnki (video.js:61:67201) 
```

I don't have Anki, so I'm assuming these errors are normal.

ShanaryS:
fixes #215 fixes #367 fixes #750 fixes #784 fixes #847 fixes #960 fixes #1091

# Overview

This started with the goal of making the playback modes testable and to unify the extension and app behaviors, then it expanded to a full rewrite of the playback modes. Now it's a complete redesign of the playback engine.

The current method is insufficient since it relies on polling which requires tons of work arounds due to how imprecise it is. It also breaks down a higher playback rates as we would need to poll frequently with expensive checks degrading performance. The solution is to move to an event based system much like how rendering is typically handled nowadays with `requestAnimationFrame()` but using `requestVideoFrameCallback()` (baseline 2024).

# Design

`requestVideoFrameCallback()` is the lower bound of the video frame rate or `requestAnimationFrame()`. It also comes with the `mediaTime` and `expectedDisplayTime` (when the browser thinks it will be displayed). For synthetic playback we will use `requestAnimationFrame()` with `Clock` instead when only subtitles are loaded. This is all powered by a new structure, essentially a replacement for `SubtitleCollection` called `PlaybackEngine`.

We will precompute all the relevant edges for showing subtitles, auto-pause, fastforward sections, etc once and use `requestVideoFrameCallback()` to know when we cross a threshold on the timeline. This allows the correct behavior even if multiple events are triggered at once, immediate pause and seek right when the frame is displayed, and is far more scalable for expansion in the future. For auto-pause and repeat playback, we actually queue this base on the `requestVideoFrameCallback()` rate (frame rate at 1x speed) in advance so they land exactly when they should. We do all other calcuations reactively since they could have worse side effects or aren't as noticeable.

The one caveat is that browsers will not update `requestVideoFrameCallback()` and `requestAnimationFrame()` when `document.hidden`. But in that case we just start listening for `timeupdate` which is far precise for detection latency but doesn't matter since the user has audio only in this case. The `PlaybackEngine` doesn't care and has zero knowledge of this an will accurately perform its duties.

`Binding`/`VideoPlayer`/`Player`
- ├── Clock (`VideoPlayer`/`Player`)
- └── PlaybackEngine
  - ├── VideoFrameTimingDriver (`Player`: AnimationFrameTimingDriver)
  - ├── PlaybackModeController
  - ├── PlaybackPlan
    - └── PlaybackTimelineCompiler
  - └── PlaybackPlanExecutor
    - ├── PlaybackTimeline
    - └── PlaybackTimelineRunner
  	  - ├── PlaybackTimeline
  	  - └── PlaybackTimelineCursor

`PlaybackEngine` is design to be a hard separation of responsibilities to ensure unified handling between all the media owners so implementations can never drift. We don't pass in the `HTMLVideoElement` directly, it takes callbacks so that each media owner can have special handling (e.g Disney+ media time). The only other interface is the timing driver, which handles if we have a video or just subtitles. `PlaybackEngine` then owns all the different playback concepts, most of which media owners do not know about or care, and so the surface for which they interact is very small, explicit, and simple.

The way I think of it is that the `PlaybackEngine` receives events about the media from the media owners and `PlaybackEngine` issues commands. When events are raised (such as settings change, subtitles changed, duration changed, etc), we will rebuild a plan and compare it to the current. We will only do further work if they are different. ~~For offset changes, we only shift the timeline rather than rebuilding the entire plan.~~

I measured the time it takes to rebuild a plan at it was around `0.1ms` for `VideoPlayer` and `1ms` for `Binding`. Since this is so negligible, it's not worth keeping the shift around. I also measured the per frame workload to ensure nothing bad is happening. The hot path (no state changes, 99% of the time) the time is about `0.05ms` with a max around `0.2ms`. The more complicated paths are still `<1ms`.

External seeks are the biggest problem for coordinating playback, we simply prefer an internal seek if they happen to be at the same time. This is the easiest solution as none of them are perfect and we can see if this is an issue. Though it shouldn't since it should be clear that X is happening because of Y and the user just happened to seek then.

Here is a visual representation of the timeline (able download the timeline html for any subtitle):
image

# General improvements

- All of the calculation is done ahead of time once when the subtitles first load and will rarely change
- The work per frame is low compared to the work per loop before
- `Clock` now uses `performance.now()` as its monotonic and more accurate
- `timeupdate` event is now sent from `requestVideoFrameCallback()` which no only keeps everything in line but is also more consistent.
- We can display showing subtitles precisely when needed since we are always in the loop on every new frame
  - This removes a lot of complexity around buffers and ranges for showing subtitles
- We should have better playback changes (auto pause, seeking, rate changes, etc) due being in the loop
  - We estimate the current `requestVideoFrameCallback()` rate (frame rate at 1x speed) and queue a pause/seek for auto-pause/repeat to ensure they land as close as possible using `expectedDisplayTime`
  - We process other events reactively (fast forward, condensed, showing subtitles, etc) as those could have worse side effects from a misprediction
- The behavior is now completely deterministic, rather than happening randomly due to interval/playback rate timing or more
  - Race conditions should be far less likely and we should be able to make stronger guarantees
  - Ability to download the `SubtitleTimeline` which can aide debugging
  - The design is far more testable and possibly allows for integration or end-to-end testing in the future
- The  `StatsOverlay` will no longer overlap with the `MobileOverlayController`, it will move down to make space when it's present
- Prevent `Alert` notifications from stacking on each other by pushing old ones down
- Fixed the white overlay background that appears when dark mode is enabled by extensions/browsers on the `MobileOverlayController` and `StatsOverlay`

# Playback modes improvements

- Reorganized the playback mode settings from `Keyboard Shortcuts` to their own section in `Misc`
  - Displays a helper text below the second to link to the shortcuts
- Added `Playback rate` which controls the media speed during non-fastfoward (even when fast-forward is enabled)
- `Show playback rate notification` switch
- `Remember last playback rate` switch, controls whether changing the playback rate on a video updates the setting, always defaults to it regardless
  - If changed rate while in fast forward mode, that setting will change. If not currently fast forwarding, the normal playback rate will be changed
- `Remember last playback modes` switch (similar to `Remember subtitle offset`)
  - Always saves the last playback modes, the switch just controls whether we load them
  - If enabled and the last mode is not normal, then we will show the playback overlay so the user is aware of it on intial load
- `Auto-pause preference` supports enabling both at the same time
  - If both are enabled and repeat is also enabled, the `atStart` pause will be skipped on repeats (but still pauses on start before repeating)
- `Repeat count preference` which controls how many repeats before continuing (0 for infinite)
- `Playback mode start/end offset` to control where the Subtitle triggers happen (auto-pause and repeat)
  - Legality is determined on a per event basis, they cannot cross the media boundary or into other events
  - If start is too positive and end is too negative such that they cross each other, we swap their roles
  - 0ms is the first ms and last ms the subtitle is visible
- `Fast-forward minimum skip interval` (similar to `Condensed playback minimum skip interval`)
- `Playback mode start/end gap` to control where the Gap triggers happen (fast-forward and condensed)
  - Legality is determined on a per event basis, they cannot cross the media boundary or into other events
  - Start is <=0 and end is >= 0 so that users can only extend the normal playback time rather than reduce it
  - 0ms is the last ms and first ms the subtitle is not visible
- Playback mode overlay shows up briefly when changed through a keybind or if not normal when remembering modes on inital load
  - TTL normally 3s, stay open when hovered, TTL 1s when unhovered

image

# Remaining work

- [x] Cleanup/refactor/simplify now that final design is in place
- [x] Replace more `SubtitleCollection` usage and rely on the `PlaybackTimeline` for VideoPlayer (fixes #1091)
- [x] Remember last playback position (closes #1077)
- [x] Retest all possible mode transitions and behaviors (previously tested, since refactored)
- [x] Verify play/pause/seek/playbackRate works on sites with an unreliable `HTMLVideoElement`
  - [x] Netflix
  - [x] Disney+

# Likely future PR
- `PlaybackEngine` is the sole owner of the video element and all playback related access must be mediated through it (some exceptions)

Assisted-by: Codex:GPT-5.6


common/app/components/Player.tsx
```diff
@@ -274,30 +266,83 @@ const Player = React.memo(function Player({
...
+                { type: 'text/html' }
+            ),
+            `${title}.html`
+        );
+    }, [playbackRate, playbackTimelineFileName, playbackTimelineModeLabels, playbackTimelineOptionLabels, settings]);
 
     useEffect(() => {
-        playModesRef.current = playModes;
-    }, [playModes]);
+        if (downloadSubtitleTimelineRequest > handledDownloadSubtitleTimelineRequestRef.current) {
```
killergerbah: I know that this pattern was being used already here in `Player` already but after gaining a bit more experience with react I think, in order to explicitly trigger actions from outside, in general it's preferable to either pass a service-like object as a prop or use a ref to trigger some kind action.

Or if we had something like redux, we could just dispatch an event from here but for now we can do a similar thing with a service-like thing.

common/locales/en.json
```diff
@@ -477,6 +486,16 @@
         "firefoxExtensionShortcutHelp": "Edit this shortcut from the Plugin manager at about:addons.",
         "postMinePlayback": "Post-mining playback state",
         "pauseOnHoverMode": "Auto-pause when mousing over subtitles",
+        "playbackModes": "Playback Modes",
+        "playbackModesHelperText": "Configure playback mode <0>keyboard shortcuts</0>.",
+        "subtitleTriggers": "Subtitle triggers",
+        "gapTriggers": "Gap triggers",
```
killergerbah: Do you think there's a more intuitive way to express what you mean here? Even if our users are fairly technical I think they'd have trouble understanding what gap triggers refers to

common/locales/es.json
```diff
@@ -9,6 +9,7 @@
         "downloadAudio": "Descargar Audio",
         "downloadImage": "Descargar Imagen",
         "downloadSubtitlesAsSrt": "Descargar Subtítulos como SRT",
+        "downloadSubtitleTimelineAsHtml": "Download subtitle timeline as HTML",
```
killergerbah, line 12: Is this developer-facing only? I am fine with not localizing strings that aren't intended for real users. Or at least, not queueing them up for localization since human localization is very slow on this project

common/locales/en.json
```diff
@@ -369,11 +370,19 @@
         "autoCopy": "Auto-copy current subtitle to clipboard",
         "autoPauseAtSubtitleEnd": "At Subtitle End",
         "autoPauseAtSubtitleStart": "At Subtitle Start",
+        "playbackModeStartOffset": "Playback mode start offset",
+        "playbackModeEndOffset": "Playback mode end offset",
+        "playbackModeOffsetHelperText": "Auto-pause and repeat playback triggers earlier or later than when subtitles are shown",
+        "playbackModesStartGap": "Playback modes start gap",
```
killergerbah: Does start gap mean that modes that only occur in gaps, like condensed mode, would be "pushed back" by a padding applied to the subtitle's start timestamp? Similar to the comment below, I think the "gap" being used here is not going to be easily understood by most users. Supplementary documentation could help, but ultimately it would be good to use expressions that are likely to be understood at a glance IMO.

killergerbah (commented):
One thing I noticed while messing around, is that applying new offsets seems to reset my playback rate setting

https://github.com/user-attachments/assets/39aa0af1-7ccf-4e03-8266-185896156e14

common/locales/en.json
killergerbah: That said I can't think of an obvious alternative. Maybe "subtitle gap" might be clearer?

common/locales/en.json
ShanaryS: Yes, gap refers to `Condensed` and `FastForward`, modes that apply in the gaps between subtitles. UI and naming definitely needs work and these mainly exist to show the settings surface.

common/locales/en.json
ShanaryS: Yeah I struggle with them too, at least initially trying to understand how start/end offsets should affect each playback mode. I think just mentioning what modes are affected may be the best, or just group pause/repeat in a section and condensed/fastforward in another. From a design point of view those groups are treated differently.

But `Subtitle gap` I think is clear and maintains the gap primer. But these names right now are very lazy.

common/locales/es.json
ShanaryS: I added it to the download subtitle button on the Header and SidePanel. It's meant to be there in production. I'm thinking this feature could help troubleshoot about "why X play back mode is at Y" but I don't think such bugs would make it to prod (hopefully). So it's mainly just a cool tool, I did originally make it for me and you to help visualize what's happening.

ShanaryS:
> One thing I noticed while messing around, is that applying new offsets seems to reset my playback rate setting

I can't seem to replicate that now, I also don't see how in the current code (I just simplified this section of code). Sorry I've been refactoring the code a lot so I'm not surprised if something broke or got fixed. I've gone through like 10 iterations of this idea so there has been a lot of failed designs I needed to clean up. I'm mostly done outside of the `common/playback` folder and so I'm gonna focus on cleaning that up now.

But the concepts of the PR are set and so it's stable enough for review. I'll probably make one new commit with some more refactors/cleanup then retest the App/Extension then.

killergerbah:
It seems to be happening because of this block

```
    useEffect(() => {
        const playbackEngine = playbackEngineRef.current;
        if (!playerChannelSubscribed || !playbackEngine) return;
        playbackEngine.settingsChanged({ ...settings, ...miscSettings });
    }, [miscSettings, playerChannelSubscribed, settings, video]);
```

`playbackRate` is now a setting but it's not fully treated that way. Setting the playback rate in `VideoPlayer` doesn't seem to persist it back to settings. So when the subtitle offset is changed (which does change the settings), the playback engine is being refreshed with the original `settings.playbackRate`. I can see how this could happen mid-refactor.

EDIT: I see now that it only reproduces when "remember last playback rate" is off

common/settings/settings-provider.ts
```diff
@@ -150,12 +150,23 @@ export const defaultSettings: AsbplayerSettings = {
...
+    playbackModeStartOffset: 0,
+    playbackModeEndOffset: 0,
+    playbackModesStartGap: 0,
+    playbackModesEndGap: 0,
     seekableTracks: 1, // Bitset with first bit flipped i.e. first track
     autoCopyableTracks: 1, // Also bitset
     subtitleHtml: SubtitleHtml.remove,
     seekDuration: 3,
     speedChangeStep: 0.1,
+    playbackRate: 1,
```
killergerbah, line 162: Why did playback rate need to become a setting? That doesn't seem like something that should obviously persist

common/app/components/notification-stack.ts
```diff
@@ -0,0 +1,12 @@
+export interface Notification<T> {
```
killergerbah: There's only ever one type used for `T`. Would it make sense to not have a generic at all? I can see how this could be reused in the future because it represents a stack, but if we just wanted a stack implementation it would be even more re-usable if it were simply named `Stack`.

common/app/components/Alert.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        const rows = Array.from(stack.children) as HTMLElement[];
+
+        expect(rows).toHaveLength(4);
+        expect(rows.map((row) => row.textContent)).toEqual([
+            'Playback rate: 1.0',
+            'Fast-forward disabled',
+            'Playback rate: 2.0',
+            'Fast-forward enabled',
+        ]);
+        expect(getComputedStyle(stack).position).toBe('fixed');
```
killergerbah: This test does accomplish the goal of making it hard to change the styles to something other than a flex column. But I have reservations about tests like these because it's essentially copying the implementation into the test. Where do you draw the line between styles that are copied into the tests and those that stay in the source?

common/app/components/Alert.tsx
```diff
@@ -30,21 +51,124 @@ interface Props {
...
+                          severity: props.severity,
+                          disableAutoHide: props.disableAutoHide ?? false,
+                      },
+                  },
+              ]
+            : []
+    );
+    const nextNotificationIdRef = useRef(props.open ? 1 : 0);
+    const previousPropsRef = useRef<
+        { children: React.ReactNode; severity: AlertColor | undefined; disableAutoHide: boolean } | undefined
```
killergerbah: Could you use `AlertNotification` here instead?

common/app/components/Bar.tsx
```diff
@@ -233,6 +250,32 @@ export default function Bar({
...
+                    <ListItem disablePadding>
+                        <ListItemButton onClick={handleDownloadSrt}>
+                            <ListItemIcon>
+                                <SaveAltIcon />
+                            </ListItemIcon>
+                            <ListItemText primary={t('action.downloadSubtitlesAsSrt')} />
+                        </ListItemButton>
+                    </ListItem>
+                    <ListItem disablePadding>
+                        <ListItemButton onClick={handleDownloadTimeline}>
```
killergerbah, line 270: It's good that this feature exists and it can be used, but is there a better place for it than the download SRT button? Depending on the person, I'm sure they might use the timeline button quite a lot, but I think it's safe to say that the SRT button has a higher chance of being used by pretty much everyone. As a result, it feels like we come out net negative on UX by adding a button click to the SRT download just so we could get the timeline download.

common/app/components/VideoPlayer.tsx
```diff
@@ -372,15 +351,21 @@ export default function VideoPlayer({
     const [playerChannelSubscribed, setPlayerChannelSubscribed] = useState<boolean>(false);
     const { fullscreen, requestFullscreen } = useFullscreen();
     const playing = () => !videoRef.current?.paused || false;
-    const [length, setLength] = useState<number>(0);
+    const [lengthMs, setLengthMs] = useState<number>(0);
```
killergerbah, line 354: Thanks, much better name

killergerbah (commented):
Btw found another regression - when loading different files the clock isn't resetting back to 0

common/app/components/Alert.test.tsx
ShanaryS: There is a few tests for `.tsx` added as I felt they were useful, but I find them in general harder to test cleanly. I don't have much experience testing UI code so I'm not sure what patterns I should be using.

common/app/components/Bar.tsx
ShanaryS: Yeah it can be moved to its own button, I wasn't sure if it was worth getting its own.

common/app/components/VideoPlayer.tsx
ShanaryS: Yeah in general I took the time to make everything in milliseconds and make the names explicit. I would love for the entire program to deal in ms but some messages between extension/app are currently in seconds. Changing that is on my list if we ever implement App versioning.

common/settings/settings-provider.ts
ShanaryS: I think it's useful as typical content consumption in a comfortable language will be different from a learning one. For example I watch most Youtube videos in 2x if the content allows but I would never do that while mining, at least not yet. This allows a user to default to a lower rate or even higher rate when mining based on their level.

You could also now with the fast forward make the "normal rate" like 0.5 and the fast forward rate 1x so the dialog moments are slower but outside of it it's normal speed. I can see such a thing being useful for non-mining immersion by artificially making the content easier to processes without pausing.

ShanaryS: Note that this rate will always be used, the `Remember last playback rate` only affects if changing the rate outside of settings will take (for both normal and fast forward).

ShanaryS:
> Btw found another regression - when loading different files the clock isn't resetting back to 0

What do you mean? When I load a different subtitle file on this branch or prod they both keep the playback position. When I load a different video file they both reset.

killergerbah:
>What do you mean? When I load a different subtitle file on this branch or prod they both keep the playback position. When I load a different video file they both reset.

Yeah I'm an idiot. It's not a regression.

common/settings/settings-provider.ts
killergerbah: Got it, makes sense

killergerbah: For what its worth, I see now that the resetting playback rate issue on this PR only reproduces when `Remember last playback rate` is off

common/app/components/Bar.tsx
killergerbah: I think it could get its own button but not at the expense of other buttons, as its mostly useful for development. Perhaps the only-on-hover side panel controls could be brought into the app. But this doesn't need to be addressed for this PR.

ShanaryS:
> One thing I noticed while messing around, is that applying new offsets seems to reset my playback rate setting

This should be fixed now

ShanaryS:
I made some simplifications, the important ones are that shifting the timeline on offset is no longer being done. Instead we just bake it into the plan like everything else. This comes with a lot of simplification to the runtime lookups.

I measured the time it takes to rebuild a plan at it was around `0.1ms` for `VideoPlayer` and `1ms` for `Binding`. Since this is so negligible, it's not worth keeping the shift around.

I also measured the per frame workload to ensure nothing bad is happening. The hot path (no state changes, 99% of the time) the time is about `0.05ms` with a max around `0.2ms`. The more complicated paths are still `<1ms`.

I made some changes around seeking to make it more robust agaisnt external seeks. Though this can never be fully accurate, especially if the external seek is near the location of the internal seek. This is the only area where we can't be fully sure. The worse case of this race condition is ending up in the wrong spot initially or a user seek doesn't cancel timeline behaviors. But these should be exceedingly rare.

common/playback/timing/clock.ts
```diff
@@ -0,0 +1,92 @@
...
+    removeEvent(eventName: ClockEvent, callback: () => void): void {
+        this.remove(callback, this.callbacks[eventName]);
+    }
+
+    private elapsedMs(): number {
+        return (this.now() - this.startedAtMs!) * this.playbackRate;
+    }
+
+    private fireEvent(eventName: ClockEvent): void {
+        for (const callback of [...this.callbacks[eventName]]) callback();
```
killergerbah, line 92: Why is it necessary to copy the array?

common/playback/playback-engine.ts
```diff
@@ -0,0 +1,318 @@
...
+ *         ├── PlaybackTimeline
+ *         └── PlaybackTimelineRunner
+ *             ├── PlaybackTimeline
+ *             └── PlaybackTimelineCursor
+ */
+export default class PlaybackEngine<T extends IndexedSubtitleModel> {
+    private settings: AsbplayerSettings;
+    private subtitles: readonly T[];
+    private ready: { settings: boolean; subtitles: boolean };
+    private subtitleOffsetMs: number;
```
killergerbah: How are you benefitting from tracking `subtitleOffsetMs` separately from the subtitles themselves?

common/playback/playback-engine.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    durationChanged(durationMs: number): void {
+        if (!Number.isFinite(durationMs) || durationMs === this.plan.timeline.durationMs) return;
+        this.rebuildPlan();
+    }
+
+    subtitleOffsetChanged(offsetMs: number): void {
+        if (!Number.isFinite(offsetMs) || this.subtitleOffsetMs === offsetMs) return;
+        this.subtitleOffsetMs = offsetMs;
+        this.rebuildPlan();
```
killergerbah, line 267: I know that it's necessary to respond to all possible events that might affect playback but it looks like by just changing the subtitle offset, the plan gets rebuilt 4 times, from the various hooks that get called. In particular, both this method and `subtitlesChanged` get called. Also, the settings hook gets called twice with the new offset.

EDIT: ^^ measured the above in the VideoPlayer's playback engine

common/playback/timing/clock.ts
ShanaryS: I think this was async at one point and so could have had the event removed mid loop awaiting each.

common/app/components/SubtitlePlayer.tsx
```diff
@@ -719,52 +790,56 @@ export default function SubtitlePlayer({
...
-                        virtuosoRef.current?.scrollToIndex({
-                            index: smallestIndex,
-                            align: 'center',
-                            behavior: 'smooth',
-                        });
-                    }
+        const indexesChanged =
+            currentSubtitleIndexes.size !== highlightedSubtitleIndexesRef.current.size ||
+            [...currentSubtitleIndexes].some((index) => !highlightedSubtitleIndexesRef.current.has(index));
+        if (indexesChanged) {
```
killergerbah, line 808: This logic looks pretty much identical to before but for some reason the scrolling doesn't work anymore when seeking off screen using keyboard shortcuts on the subtitle player

https://github.com/user-attachments/assets/4fb8216a-381a-45a2-aff5-6c011039a842

common/playback/playback-engine.ts
ShanaryS: This was due to shifting the timeline which used the offset at runtime. It was also there because I wasn't sure if the subtitles being passed in would always be up to date and wanted the ability to have it independent.

But I didn't think any of those matter anymore.

common/playback/playback-engine.ts
ShanaryS: I plan to run tests on all the operations to ensure the plans aren't being built unnecessarily. It's part of the testing I have where I'd cleanup these aspects. I'm currently doing one more pass of the playback pipeline before I'm done and will start optimizing.

common/app/components/SubtitlePlayer.tsx
ShanaryS: I made some separate changes to the scroll detection as it wasn't behaving properly moving away from intervals. It's the `SubtitleScroller` which I see I didn't mention in the description. So it's probably a regression from that.

Just thinking about it seems right as the new scroll conditions are more explicit.

common/playback/controllers/playback-mode-controller.ts
```diff
@@ -0,0 +1,115 @@
...
+        }
+    };
+
+    const notifications: string[] = [];
+    for (const mode of transition.removed) {
+        if (mode === PlayMode.normal) continue;
+        notifications.push(getLocKey(mode, false));
+    }
+    for (const mode of transition.added) notifications.push(getLocKey(mode, true));
+    return { notifications, join: ' | ' };
```
killergerbah, line 81: Nitpick, it's possible for a notification like below to show, but it would have the same meaning with fewer words if it just displayed "all play modes off"

Image

common/components/PlayModeSelector.tsx
```diff
@@ -21,7 +21,7 @@ interface Props extends PopoverProps {
 
 const ListItem = ({ children, ...props }: ListItemProps) => {
     return (
-        <MuiListItem disablePadding dense {...props}>
+        <MuiListItem disablePadding dense sx={{ width: 'auto' }} {...props}>
```
killergerbah, line 24: With the local video player and mobile simulator the playback modes now look like this

Image

which is actually more usable but it still would look a lot better if it were completely a column

common/playback/plan/playback-plan-executor.ts
```diff
@@ -0,0 +1,461 @@
...
+    private readonly callbacks: PlaybackPlanExecutorCallbacks<T>;
+    private repeatedBlock?: RepeatedBlock;
+    private pendingTarget?: PendingTarget;
+    private startPauseSuppression?: StartPauseSuppression;
+    private condensedOperation?: number;
+    private operationGeneration = 0;
+    private updateOperationGeneration = 0;
+    private showingSubtitles: readonly T[] = [];
+    private _isFastForwarding: boolean;
+    private expectedDiscontinuity?: ExpectedDiscontinuity;
```
killergerbah, line 82: Is it accurate to say that expected discontinuities only result from internal seeks due to the a playback plan? Could this be made clearer from naming and/or documentation?

killergerbah:
>This should be fixed now

Looks fixed on my end as well

common/playback/controllers/playback-mode-controller.ts
ShanaryS: Yeah I have that fixed locally.

common/playback/plan/playback-plan-executor.ts
ShanaryS: Yeah, they are repeat or condensed jumps. Also could be auto pause correction.

common/playback/timing/clock.ts
ShanaryS: Yeah I think it makes sense to keep to ensure all callbacks receive their event if one unsubscribes during processing.

ShanaryS:
I changed the way the race condition resolution with internal/external seeks, we simply now prefer internal seeks. This should be a rare occurrence but simplifies the code a bit and I don't think is confusing for the user. But we can monitor this going forward.

A big-ish design change is to now use `timeupdate` when `document.hidden`. Browsers will not update `rVFC` and `rAF` in this case and only play audio. This means that the playback modes will not get triggered at all while in the background. Using `timeupdate` for this is straightforward but we lose the next frame prediction for perfect application of playback modes but it doesn't matter since the user can't see the video anyways. `PlaybackEngine` has no knowledge of this switch as it's isolated to the timing drivers and everything works exactly the same, we just get time events at a different rate from a different source.

ShanaryS:
(#1091) I removed the window subtitles and the last of `SubtitleCollection` from `VideoPlayer`. Now the only areas that `SubtitleCollection` is being used is for `Player` and `SubtitleAnnotations` as those shouldn't depend on `PlaybackEngine` directly.

(#1077) I implemented this PR into the `PlaybackEngine` in its current state. Some notable changes are:
- A minimum time of 60s to remember playback position (removes existing if below)
- Triggers save on seek and pause in addition to the 10s interval
- If seek is within showing subtitles it will instead seek to the start of the subtitles
- The snackbar stays open on hover. It also opens at the bottom as I believe that's easier to reach and less crowed with other overlays and notifications at the top.
- Support for extension media

For extension media, we use the subtitle names (as they appear in the selector) to remember positions by since the video src is usually a uuid. This should work well for sites that use the titles from the media player but I cannot confirm them all. It will use the minimum time from any of the subtitle files for suggestion and will update them all at once as needed.

common/app/services/video-channel.ts
```diff
@@ -87,6 +91,7 @@ export default class VideoChannel {
     private loadFilesCallbacks: (() => void)[];
     private loadSubtitlesCallbacks: (() => void)[];
     private playModesCallbacks: ((modes: Set<PlayMode>) => void)[];
+    private pendingPlayModes?: Set<PlayMode>;
```
killergerbah: Since you're maintaining some additional state like `paused` would it make senes to continue that pattern with `playModes` so that `Player` can inspect existing state when subscribing to play modes, instead of making a special cause to publish pending state here? Complexity would be shifted to the `Player` but I think it would be worth it to keep the code inside here somewhat consistent.

common/components/PlayModeSelector.tsx
```diff
@@ -77,7 +124,18 @@ export default function PlayModeSelector({
...
             {...restOfPopoverProps}
         >
-            <List disablePadding dense sx={listStyle}>
+            <List
+                ref={handleListRef}
+                disablePadding
+                dense
+                sx={{
+                    ...listStyle,
+                    flexDirection: useColumnLayout ? 'column' : 'row',
```
killergerbah: For now I think you should consider to use a row for streaming video
I've been wanting to fix up the overlay for streaming video, so that it can dynamically resize its containing iframe like the statistics overlay does, but until that's done the column isn't going to fit

Also, it looks like the menu is auto-closing, and the overlay refreshes to its starting state, which looks a bit weird.

https://github.com/user-attachments/assets/db775b7f-f3bc-435b-b40d-63fb5671a95e

killergerbah:
Reporting an issue I found when using condensed mode on Disney Plus. Video is missing since it's hidden when I screen record.


https://github.com/user-attachments/assets/9dfe8824-c03c-4884-8491-caaa81bc7c3f

I think this happening because with our integration with Disney Plus (and maybe other sites), seeking is imprecise, so seeking to the next subtitle seeks to a point at or before the previous gap, which causes condensed mode to seek back to the subtitle, ad infinitum.

EDIT: I tried reproducing this a main branch build and it still reproduces, so this not an issue related to this change, and related mostly to the seeking imprecision.

Attached the subs in case it's helpful. 

[スター・ウォーズエピソード1／ファントム・メナス - English [CC].vtt.txt](https://github.com/user-attachments/files/30368071/1.-.English.CC.vtt.txt)


[スター・ウォーズエピソード1／ファントム・メナス - English [CC].vtt.html](https://github.com/user-attachments/files/30368051/1.-.English.CC.vtt.html)

common/playback/playback-engine.ts
```diff
@@ -0,0 +1,359 @@
...
+            onPlaybackStarted: () => this.executor.playbackStarted(),
+            onError: callbacks.onError,
+        });
+    }
+
+    bind(): void {
+        if (this.timingDriver.bound) return;
+        if (!this.ready.settings || !this.ready.subtitles) return;
+
+        const transition = this.playbackModeController.setModes(this.playbackModeController.playModes);
```
killergerbah, line 166: When does this return a nonempty transition?

common/playback/playback-engine.ts
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    playbackRateChanged(playbackRate: number): { readonly notify: boolean; readonly playbackRate: number } {
+        const setting = this.executor.isFastForwarding ? 'fastForwardModePlaybackRate' : 'playbackRate';
+        const normalizedPlaybackRate = normalizePlaybackRate(playbackRate);
+        if (normalizedPlaybackRate === undefined || this.settings[setting] === normalizedPlaybackRate) {
+            return { notify: false, playbackRate: this.settings[setting] };
+        }
+        this.settings = { ...this.settings, [setting]: normalizedPlaybackRate };
+        this.rebuildPlan();
```
killergerbah: How is the playback rate affect the behavior of the plan? It value appears to be maintained as state, and is passed around, but it doesn't seem to affect the plan output, other than to appear as a field. Would it be simpler to not to interact with playback rate at all from this class?

ShanaryS:
> seeking to the next subtitle seeks to a point at or before the previous gap, which causes condensed mode to seek back to the subtitle, ad infinitum.

I think we can apply a corrective seek to condensed as well since the gaps will be huge. Perhaps repeat too though that would probably be distracting if necessary.

In addition we could ban the same trigger being used again before a play, pause, non corrective seek, or another trigger.

EDIT: I think what's being described is different from what I thought. So it's trying to say seek to `15s` but is seeking to `8s` which is the previous gap? Idk if that can be addressed, maybe by compensating for the difference if the seeks are precise but just not accurate.

common/playback/playback-timeline-runner.ts
```diff
@@ -0,0 +1,90 @@
...
+} from '@project/common/playback/playback-timeline';
+import PlaybackTimelineCursor from '@project/common/playback/playback-timeline-cursor';
+
+export interface PlaybackTimelineActionResult {
+    readonly autoPaused: boolean;
+    readonly seeked: boolean;
+}
+
+export interface PlaybackTimelineRunnerCallbacks<T extends SubtitleModel> {
+    onStart(event: PlaybackTimelineEvent): boolean | Promise<boolean>;
```
killergerbah: I can see from the usage that `onStart` returns whether a particular timestamp is autopaused. Could this be documented? It's not obvious from the method signature returning a `boolean` that this is the case

common/playback/timeline/playback-timeline-runner.ts
```diff
@@ -0,0 +1,90 @@
...
+    readonly autoPaused: boolean;
+    readonly seeked: boolean;
+}
+
+export interface PlaybackTimelineRunnerCallbacks<T extends SubtitleModel> {
+    onStart(event: PlaybackTimelineEvent): boolean | Promise<boolean>;
+    onEnd(event: PlaybackTimelineEvent): PlaybackTimelineActionResult | Promise<PlaybackTimelineActionResult>;
+    correctAutoPause(timestampMs: number): Promise<void>;
+    /** Reconciles persistent state only: subtitles, playback rate, and other state that must survive seeks. */
+    onState(state: PlaybackTimelineState, segment: PlaybackTimelineSegment<T>): Promise<void>;
```
killergerbah, line 22: I see that `onState` leads to `reconcileShowingSubtitles` in `PlaybackPlanExecutor`. Is this comment out of date, or are the showing subtitles considered persistent state?

common/playback/playback-timeline-runner.ts
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+
+        const initialUpdate = this.initialUpdate;
+        this.initialUpdate = false;
+        if (groups.length === 0 && !initialUpdate) return;
+        if (groups.length > 0) await this.applyState(timestampMs);
+        if (groups.some((group) => group.direction === 'backward')) return;
+        const stateChangedPosition = await this.callbacks.onAfterState(timestampMs);
+        if (stateChangedPosition) this.cursor.reset(timestampMs, { includeAtTimestamp: true });
```
killergerbah: Reading the concrete implementation of `onAfterState`, I see that it returns whether the executor seeked to the next condensed target. Why is the response to that to reset the cursor to `timestampMs`? Isn't `timestampMs` now the wrong timestamp?

common/playback/playback-timeline-runner.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export interface PlaybackTimelineRunnerCallbacks<T extends SubtitleModel> {
+    onStart(event: PlaybackTimelineEvent): boolean | Promise<boolean>;
+    onEnd(event: PlaybackTimelineEvent): PlaybackTimelineActionResult | Promise<PlaybackTimelineActionResult>;
+    correctAutoPause(timestampMs: number): Promise<void>;
+    /** Reconciles persistent state only: subtitles, playback rate, and other state that must survive seeks. */
+    onState(state: PlaybackTimelineState, segment: PlaybackTimelineSegment<T>): Promise<void>;
+    /** Applies non-edge continuous behavior such as condensed playback after persistent state is current. */
+    onAfterState(timestampMs: number): boolean | Promise<boolean>;
```
killergerbah: Could this return value be documented as well?

common/components/PlaybackRateInput.tsx
```diff
@@ -9,10 +10,10 @@ interface Props extends InputProps {
     disableKeyEvents?: boolean;
 }
 
-const valueToPrettyString = (v: number) => '×' + String(v.toFixed(2));
+const valueToPrettyString = (v: number) => '×' + String(v.toFixed(3));
```
killergerbah: Why 3 decimal points now? It doesn't look that great on my computer:

Image

common/components/PlaybackRateInput.tsx
```diff
@@ -26,7 +27,10 @@ export default React.forwardRef(function PlaybackRateInput(
             valueToPrettyString={valueToPrettyString}
             stringToValue={stringToValue}
             numberValue={playbackRate}
-            onNumberValue={onPlaybackRate}
+            onNumberValue={(value) => {
+                const normalized = normalizePlaybackRate(value);
```
killergerbah: It's not obvious to me why but this change is causing the playback rate input to be unselectable for me

extension/src/ui/components/NotificationUi.tsx
```diff
@@ -78,6 +109,29 @@ const NotificationUi = ({ bridge }: Props) => {
...
+                <Snackbar
+                    open={resumeSnackbar.open}
+                    anchorOrigin={{ vertical: 'bottom', horizontal: 'center' }}
+                    onClose={resumeSnackbar.close}
+                    onMouseEnter={resumeSnackbar.onMouseEnter}
+                    onMouseLeave={resumeSnackbar.onMouseLeave}
+                >
+                    <Alert
+                        severity="info"
+                        onClose={resumeSnackbar.close}
```
killergerbah, line 122: Would it make sense to include the logo icon here as well so that the alert is easily recognizable as asbplayer UI? Since we use a standard UI framework it's not unlikely that a notification like this would blend in with the website it's appearing on.

killergerbah:
> I think what's being described is different from what I thought. So it's trying to say seek to 15s but is seeking to 8s which is the previous gap? Idk if that can be addressed, maybe by compensating for the difference if the seeks are precise but just not accurate.

Maybe fine not to address at all for now, guessing that amount of inaccuracy would be rare

common/components/MiscSettingsTab.tsx
```diff
@@ -41,6 +45,14 @@ function regexIsValid(regex: string) {
     }
 }
 
+const responsiveSettingsStackSx = {
```
killergerbah: Might be worth making this re-usable via a custom component

extension/src/controllers/statistics-overlay-controller.ts
```diff
@@ -226,15 +323,15 @@ export class StatisticsOverlayController {
                 this._applyOverlayContainerStyles(container);
             },
         });
+        const colorScheme = frameColorScheme();
         const colorSchemeClass = frameColorSchemeClass();
         this._overlay.setHtml([
             {
                 key: 'ui',
                 html: () => {
-                    console.log('rendering iframe');
```
killergerbah, line 234: oops

common/playback/timeline/playback-timeline-html.ts
```diff
@@ -0,0 +1,801 @@
...
+    const left = ((value.startSeconds - rowStartSeconds) / 10) * 100;
+    const width = ((value.endSeconds - value.startSeconds) / 10) * 100;
+    const renderedWidth = value.className === 'normal' ? width : Math.max(width, 0.15);
+    return `<div class="event ${value.className}" style="left:${left}%;width:${renderedWidth}%;background:${value.color}"${
+        value.title === undefined ? '' : ` title="${escapeHtml(value.title)}"`
+    }>${value.className === 'normal' && value.label !== undefined ? escapeHtml(value.label) : ''}</div>`;
+};
+
+const renderModeToggle = (className: string, label: string, color: string, checked: boolean): string =>
+    `<label class="mode-toggle"><input type="checkbox" data-mode="${className}"${checked ? ' checked' : ''}><span class="legend-swatch ${className}" style="background:${color}"></span>${label}</label>`;
```
killergerbah, line 326: Any reason why react couldn't be used to generate the HTML?

common/playback/playback-engine.ts
ShanaryS: The `rebuildPlan()` issue is mainly due to the `VideoPlayer` settings dep changing calling `settingsChanged()`, `Binding` does not have the behavior. I think to fix it for `VideoPlayer` would require being aware of the settings changes so that we can ignore the emitter to not re-process, but I don't think such complexity is warranted.

I readded the check to see if plans are equal before replacing them. It won't stop the extra checks but we minimize the unnecessary work. The plan build + check is about `0.1ms` for `VideoPlayer` so this is fine.

common/components/PlaybackRateInput.tsx
ShanaryS: What do you mean? It seems to be working for me, both the arrows and typing. It does get weird when it's at the minimum though for me.

common/components/PlaybackRateInput.tsx
ShanaryS: Off by one error... What I wanted was the value it was already at.

common/playback/timeline/playback-timeline-html.ts
ShanaryS: I felt that using pure functions were easier to test. I want to prevent this code from drifting from the runtime code as much as possible.

common/playback/playback-engine.ts
ShanaryS: Never, but `playbackModesChanged()` takes a transition and I think it's better than manually creating one.

common/playback/playback-engine.ts
ShanaryS: I want all the details that affect the timeline to exist in the plan and for the plan to be immutable. I prefer rebuilding the plan more than necessary for that considering it's so cheap.

> Would it be simpler to not to interact with playback rate at all from this class?

I'm not sure what you mean. The `PlaybackEngine` class is only checking to see if we are currently fast forwarding to know which playback rate setting to update.

common/playback/timeline/playback-timeline-runner.ts
ShanaryS: Yes, `showingSubtitles` and `playbackRate` are the persistent state. The `PlaybackEngine` handles those as well as one time triggers such as `autoPause` and seeks for `repeat` and `condensed`. After seeking or a state change, we need to check if we should update the `showingSubtitles` or `playbackRate` since they are a range rather than a point.

common/playback/playback-timeline-runner.ts
ShanaryS: Yes that was a bug. I seems like it could be related to the D+ seeks but I doubt it.

common/components/PlayModeSelector.tsx
ShanaryS: The auto close was intentional but it probably doesn't need it when manually opened.

ShanaryS:
I've now pretty much tested every option and how they interact with every other. I've made some changes to make certain edge cases explicitly defined and some changes:
- When the adjusting fast forward rate we display a specific localization for it
- Update a numeric settings inputs to use a new component that better handles numeric input. This was lifted from the screenshot delay implementation so that every setting can benefit. This treats the inputs as strings initially so that users can type `-` or `.1` among others.
- Explicitly test that overlapping showing subtitles are treated as one block for auto pause and repeat. We can revisit this in the future, it originally made the timeline less complicated
- Clarified how overlapping offsets from different subtitles interact. Auto pause will only happen once if a end offset is meets and another start offset.
- Suppressing the start auto pause (when auto pausing on start and end with repeat enabled) is now always done regardless of showing subtitle semantics. Previously it would only suppress if the showing subtitles are the same at the start and end pauses but this restriction isn't necessary for the main benefit of preventing double pauses for repeats.
- When the start offset for auto pause is before the start of the subtitle, condensed will now seek to it instead of the start of the subtitle and will no longer end up in an infinite loop of seeking and auto pausing
- Similarly, condensed mode will no longer jump at the end of a subtitle if auto pause or repeat is delayed until after the subtitle
- Fast forward does not have the above condensed suppressions since it's not a destructive operation
- Compiled more runtime checks into the timeline. Also now using a lookahead cursor explicitly for next frame prediction.
- Isolated the state and action timestamps so runtime lookups don't need to iterate over unnecessary items
- Normalizes all the inputs when compiling the timeline to ensure that `PlaybackEngine` is robust against unexpected data

common/components/PlaybackRateInput.tsx
killergerbah: Clicking on the input doesn't seem to do anything

https://github.com/user-attachments/assets/c6345508-4c8d-4342-8855-2677e589c417

I debugged a bit and this change seems to fix it. I guess extra renders are causing the input to get deselected.

```
     { inputRef, playbackRate, onPlaybackRate, ...rest }: Props,
     ref
 ) {
+    const handleNumberValue = useCallback(
+        (value: number) => {
+            const normalized = normalizePlaybackRate(value);
+            if (normalized !== undefined) onPlaybackRate(normalized);
+        },
+        [onPlaybackRate]
+    );
     return (
         <VideoControlInput
             ref={ref}
@@ -27,10 +34,7 @@ export default React.forwardRef(function PlaybackRateInput(
             valueToPrettyString={valueToPrettyString}
             stringToValue={stringToValue}
             numberValue={playbackRate}
-            onNumberValue={(value) => {
-                const normalized = normalizePlaybackRate(value);
-                if (normalized !== undefined) onPlaybackRate(normalized);
-            }}
+            onNumberValue={handleNumberValue}
             rejectValue={rejectValue}
             placeholder={placeholder}
             {...rest}
```

common/playback/playback-engine.ts
killergerbah: I see that the engine uses the values to know which playback rate number to use 
But from my reading the plan itself seemed like it didnt need any knowledge of the values since it just determines when to go faster or slower

common/components/PlaybackRateInput.tsx
ShanaryS: Okay I see it now, I was looking at the playback rate in the settings not the VideoPlayer.

common/playback/playback-engine.ts
ShanaryS: I think I understand your point now, we could just update the rate values as well as immediately change the rate and that would be fine. But I would still want it in the plan for completeness and we would need a separate flow just for this.

I don't think it's worth it since building a new plan, comparing it, replacing it, and reconciling it only takes `0.1ms`. The other input changes needs this rebuild so we will be stuck with it for some aspects no matter what so to me it makes no difference if this is unnecessary. I don't think it's worth the minor complexity and future maintenance needed to update this without a full rebuild.

It's also why for the rebuild I'm just building a new plan and deep comparing it to the current as it guarantees safety against any future issues doing it this way. Such a simple direct comparison can never be stale or wrong since we see if any information is different in anyway. I think `PlaybackEngine` should be the most defensively programmed area of asbplayer since it's so fundamental.

common/components/PlayModeSelector.tsx
ShanaryS: I removed the auto close but the flashing of the number seems more fundamental. It's probably best for you to tackle that when doing the other fixes.

common/components/KeyboardShortcutsSettingsTab.tsx
```diff
@@ -541,7 +455,7 @@ const KeyboardShortcutsSettingsTab: React.FC<Props> = ({
         }
 
         return (
-            <div key={key}>
+            <div key={key} id={keyBindName === 'toggleAutoPause' ? 'playback-mode-key-bindings' : undefined}>
```
killergerbah: Nit, this could be another keybind property

common/components/MiscSettingsTab.tsx
```diff
@@ -143,6 +177,14 @@ const MiscSettingTab: React.FC<Props> = ({
         <>
             <Stack spacing={1}>
                 <SettingsSection>{t('settings.ui')}</SettingsSection>
+                <Stack direction="row" spacing={1}>
```
killergerbah, line 180: Did you move these here to make them more visible? It's under the UI subsection now

common/hooks/use-snackbar.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+        });
+    };
+
+    it('closes after six seconds', () => {
+        const onClose = jest.fn();
+        renderSnackbar(true, onClose);
+
+        act(() => jest.advanceTimersByTime(5_999));
+        expect(onClose).not.toHaveBeenCalled();
+        expect(container.querySelector('button')?.dataset.open).toBe('true');
```
killergerbah, line 49: Is there a way to test if the button is actually visible? This seems to depend on an implementation detail of the button, where if the button is visible, the dataset happens to have an open key equal to true. Maybe this is something jest or react testing library can help with.

killergerbah (commented):
Sorry I've only had 15-30 minutes to look at this every day and I end up looking at what I happen to scroll by
I'd like to read more deeply into the playback engine pretty soon

common/playback/timeline/playback-timeline-html.ts
killergerbah: Ok, at first thought it doesn't seem like the choice of rendering library/method would affect how easy it is for the code to drift. It also looks like this file would be difficult to make changes to because of code being inside of strings rather than in code.

killergerbah: Do you think it's possible for the file at least be written to include the playback engine code directly? That would automatically accomplish your goal of minimizing drift, without having to use tests to do the same thing.

One idea off the top of my head, would be to use a similar strategy to what's already in the code base. You could build a separate entrypoint for both webapp and extension, and include that entrypoint, along with the raw subtitle data, in the HTML file you generate. This way the code is inside entrypoint and the only point of the generated file is to pass the data to the code. You would also get complete access to all of playback engine, asbplayer UI components etc.

common/components/MiscSettingsTab.tsx
ShanaryS: Yeah, at least one person couldn't find it so I though to make the import/export more prominent. I tried giving it it's own section but didn't really fit with the `UI` section being there, so I just basically use that which kinda fits.

common/playback/timeline/playback-timeline-html.ts
ShanaryS: Yeah I think it can be rewritten and have `PlaybackEngine` updated to support this better. But this has not been a priority and probably won't be until after this PR. This was originally not meant to be shipped but it's of low concern and currently still accurate for the options we have.

common/hooks/use-snackbar.test.tsx
ShanaryS: I don't have much experience with this, I think we can go over some libraries to test the UI side. I don't know what the pattern of testable react code looks like compared to normal classes and functions.

common/app/hooks/use-subtitle-dom-cache.ts
```diff
@@ -9,24 +9,11 @@ export const useSubtitleDomCache = (
 
     useEffect(() => {
         const domCache = new OffscreenDomCache();
+        for (const subtitle of subtitles) domCache.add(String(subtitle.index), render(subtitle));
```
killergerbah: Looks like this is re-adding all the subtitles to the dom cache on every pause. This is noticeable slowdown when testing with movie-length BluRay-format `sup` subtitles. Can't upload here but easy to find some on Kitsunekko.

killergerbah (approved):
The playback code is a lot better now, more accurate and composable with new behavior. I appreciate all the work you put into this.

I won't plan to read more deeply into the code, the way you split up the responsibilities and composed the various parts seems reasonable, looks clean, and appears to be well tested. I haven't encountered any new playback bugs in my own testing including Netflix and Disney Plus.

There are some UI/UX issues that would be nice to fix before release. Also it would be nice if the timeline html export could become more maintainable.

Can you address the last comment on the subtitle cache hook? Other than that feel free to merge when you're ready.

common/components/MiscSettingsTab.tsx
killergerbah: I may just move it up above the UI header to satisfy my OCD

common/hooks/use-snackbar.test.tsx
killergerbah: I don't have much experience either, so my instinct is to just copy how its done in other places
As a matter of principle I do think we should prefer testing desired behavior without depending on specific implementation details, regardless of whether its UI or pure logic. I don't know why AI is so bad at that.

common/app/hooks/use-subtitle-dom-cache.ts
ShanaryS: This identified a few inefficiencies:

The dom cache was being cleared even when no meaningful change to the rendered subtitles. This didn't affect much but an easy optimization.

Settings persistence for values such as `lastPlaybackPositions` (your case), `lastPlaybackModes`, `lastSubtitleOffset`, `playbackRate`, and `fastForwardModePlaybackRate` were triggering effects due to their change causing lots of rerenders. Since `lastPlaybackPositions` and `lastPlaybackModes` are internal to `PlaybackEngine` and the rest can be updated through a keybind during playback (causing frequent updates), I added a new option when saving settings called `saveOnly` which defaults to false.

This new option only affects whether the new settings object will be replaced through `setSettings` which triggers all the effects through the app. These settings should all be owned by `PlaybackEngine` and relevant subscribers should use it as the single source of truth. These values being in the settings is now considered an implementation detail for `PlaybackEngine`.

This also fixes the previous issue of `rebuildPlan()` being called multiple times in the https://github.com/asbplayer/asbplayer/pull/1089#discussion_r3634496292

killergerbah: Nice, makes sense if the playback engine settings values are never published from the outside
I wonder if a similar approach could reduce re-renders while typing into settings forms

PR #1090, opened 2026-07-19
WebSocket API: media targeting for seek-timestamp/mine-subtitle and note targeting for mine-subtitle

mcgrizzz:

As part of the issue here: https://github.com/asbplayer/asbplayer/issues/1087

- Adds `mediaId` to both `seek-timestamp` and `mine-subtitle`. 
- Adds `noteId` to `mine-subtitle`

Without mediaId, it falls back to the active media (as before). 
Without a noteId, it falls back to the last updated note (as before).

Manually tested before opening this PR:

* `seek-timestamp` targets the specified `mediaId`, even when another tab is active.
* Without a `mediaId`, `seek-timestamp` still defaults to the active tab.
* `mine-subtitle` can update a note other than the last-added note.
* Without a `noteId`, `mine-subtitle` still defaults to the last-added note.
* `mine-subtitle` targets the specified `mediaId`, even when another tab is active.
* The proxied `addNote` endpoint still works. It now updates the note using the ID returned by `addNote`, avoiding a race if another note is added immediately afterward.


scripts/web-socket-server/main.go
```diff
@@ -193,7 +194,20 @@ func (forwarder forwarder) forwardToAnkiConnect(buf *bytes.Buffer, c echo.Contex
 	}
 
 	c.Blob(ankiConnectResponse.StatusCode, ankiConnectResponse.Header["Content-Type"][0], ankiConnectResponseBuf.Bytes())
-	return nil
+	return ankiConnectResponseBuf.Bytes(), nil
+}
+
+// Extracts the created note's id from an AnkiConnect addNote response
+func addNoteIDFromResponse(response []byte) (int64, bool) {
```
killergerbah: `addNoteIDFromResponse` feels like it could be interpreted in different ways e.g. "add the note id from the response" or "`addNote` ID from response. You probably wouldn't even need the comment doc if you just named the method using the same words as the comment e.g. `extractNoteIdFromAnkiConnectResponse` or similar.

scripts/web-socket-server/main.go
```diff
@@ -193,7 +194,20 @@ func (forwarder forwarder) forwardToAnkiConnect(buf *bytes.Buffer, c echo.Contex
...
+	parsed := struct {
+		Result *float64 `json:"result"`
```
killergerbah: Nitpick, `float64` seems to suggest that a non-integer could come back, maybe `int64` would more directly represent what the response is supposed to be

extension/src/services/web-socket-client-binding.ts
```diff
@@ -92,30 +135,101 @@ export const bindWebSocketClient = async (settings: SettingsProvider, tabRegistr
...
+                        asbplayer
+                    ): ExtensionToAsbPlayerCommand<CopySubtitleWithAdditionalFieldsMessage> => ({
+                        sender: 'asbplayer-extension-to-player',
+                        message: {
+                            command: 'copy-subtitle-with-additional-fields',
+                            word,
+                            definition,
+                            text,
+                            postMineAction,
+                            customFieldValues,
```
killergerbah: Is updating a specific note ID supposed to work when targeting local media as well? `noteId` is missing here

extension/src/services/web-socket-client-binding.ts
```diff
@@ -92,30 +135,101 @@ export const bindWebSocketClient = async (settings: SettingsProvider, tabRegistr
...
+                        videoElement.src !== target.videoElement.src
+                    ) {
+                        return undefined;
+                    }
+
+                    const extensionToVideoCommand: ExtensionToVideoCommand<CopySubtitleMessage> = {
+                        sender: 'asbplayer-extension-to-video',
+                        message: {
+                            command: 'copy-subtitle',
+                            word,
```
killergerbah: Nitpick, since you just pass this collection of values as is, you could just spread the `...ankiFieldValues` to reduce the number of lines here

extension/src/services/web-socket-client-binding.ts
mcgrizzz: Not currently, I believe I was thinking of doing it in another PR because it's a bit more involved and that way I can work on it in the same place I'm porting `get-bound-media` and `get-subtitles` to the local player.

common/app/components/SubtitlePlayer.tsx
```diff
@@ -1100,11 +1100,17 @@ export default function SubtitlePlayer({
...
             return extension.subscribe((message: ExtensionMessage) => {
-                if (!document.hasFocus() || message.data.command !== 'copy-subtitle-with-additional-fields') {
+                if (message.data.command !== 'copy-subtitle-with-additional-fields') {
                     return;
                 }
 
                 const copySubtitleMessage = message.data as CopySubtitleWithAdditionalFieldsMessage;
+
+                // Commands targeting specific media are routed to a single asbplayer
+                if (!copySubtitleMessage.targetedByMediaId && !document.hasFocus()) {
```
killergerbah: Why is this necessary? The `web-socket-client-binding` is already targeting a specific asbplayer right?

mcgrizzz: It does target a specific asbplayer, but `document.hasFocus()` still drops a targeted mine while reporting `published: true`.

So this change prevents that drop when targeted, and keeps the legacy behavior when not.

Didn't notice until I went back through the PR yesterday.

extension/src/services/web-socket-client-binding.ts
```diff
@@ -92,30 +136,96 @@ export const bindWebSocketClient = async (settings: SettingsProvider, tabRegistr
...
+                .filter((entry) => entry !== undefined)
+        );
+        return { word, definition, text, customFieldValues };
+    };
+
     client.onMineSubtitle = async ({
-        body: { fields: receivedFields, postMineAction: receivedPostMineAction },
+        body: { fields: receivedFields, postMineAction: receivedPostMineAction, mediaId, noteId },
     }: MineSubtitleCommand) => {
+        if (mediaId !== undefined) {
```
killergerbah: Sorry to keep nitpicking this late into the PR but I am realizing that this function is very long, and has 4 different cases depending on presence of media ID and sending to video element vs asbplayer. The lines responsible for forwarding the messages are also very similar.

Would it be possible to reduce the number of branches?

extension/src/services/web-socket-client-binding.ts
```diff
@@ -187,7 +292,33 @@ export const bindWebSocketClient = async (settings: SettingsProvider, tabRegistr
             return toggleVideoSelectCommand;
         });
     };
-    client.onSeekTimestamp = async ({ body: { timestamp } }: SeekTimestampCommand) => {
+    client.onSeekTimestamp = async ({ body: { timestamp, mediaId } }: SeekTimestampCommand) => {
+        if (mediaId !== undefined) {
```
killergerbah: Similarly here, can the branching be reduced?

common/app/components/SubtitlePlayer.tsx
killergerbah: I see now, thanks for explaning

I am just guessing what I was thinking when I wrote it - `document.hasFocus()` filter was an easy check to ensure that the WebSocket command would only mine the instance the user is actually viewing. Now that we would like to ignore this check, I think it's preferable to simply remove it, and replace it with more specific targeting logic in the `web-socket-client-binding`. It should be possible to only target tabs that are actually in focus using the `tabs` API instead.

The reason I'm being kind of picky here is because `copy-subtitle-with-additional-fields` is a message that crosses extension/app boundaries, and so once a field is introduced it can't be dropped for free. Not likely to matter too much in practice, but if we can achieve the same behavior without making that change then I would prefer doing that instead.

extension/src/services/web-socket-client-binding.ts
mcgrizzz: Yup, I'll work on both the seek and mine handlers. Not nitpicking at all, I definitely see your concern here.

killergerbah (approved):
Looks really good, thanks for your patience

PR #1097, opened 2026-07-26
Fix Disney+ mining pause bypass and subtitle auto-detection race

Ayase-the-Dark:
## Summary
- `recording-finished` and the subtitle-hover auto-pause called `this.video.pause()` directly, bypassing the Disney+/Netflix player-API dispatch that `pause()` already provides for seek/play. Route both through `pause()` for consistency.
- Disney+'s subtitle auto-detection hook (`JSON.parse`/manifest interception in `disney-plus-page.ts`) was installed behind a `setTimeout(0)`, which could lose the race against Disney+'s own bundle parsing its playback manifest on the same tick, silently leaving `lastM3U8Url`/`lastBasename` unset. Install the hooks synchronously instead, and additionally patch `Response.prototype.json` (its internal parser never calls the patched `JSON.parse`) as defensive coverage.

## Test plan
- [x] `yarn workspace @project/extension test` (43 tests passing)
- [x] `yarn workspace @project/extension compile` (tsc --noEmit, no errors)
- [x] `yarn workspace @project/extension dev` build succeeds
- [x] Manually verified via chrome-devtools-mcp against live Disney+ playback: player-API seek/pause/resume sequence holds up across a simulated mining cycle (seek → wait → play → pause → resume) with no drift in `player.timeline.info.playheadPositionMs` vs wall clock
- [ ] Would appreciate a maintainer/community check on real-world subtitle auto-detection reliability improvement, since the underlying race is timing-dependent and hard to guarantee eliminated in CI

Assisted-by: Claude:claude-sonnet-5

### DIFF
--- a/extension/src/entrypoints/disney-plus-page.ts
+++ b/extension/src/entrypoints/disney-plus-page.ts
@@ -150,13 +150,16 @@ export default defineUnlistedScript(() => {
     document.addEventListener(playEventName, () => disneyPlusPlayer()?.play());
     document.addEventListener(pauseEventName, () => disneyPlusPlayer()?.pause());
 
-    setTimeout(() => {
+    // Install the JSON.parse/Response.json hooks synchronously (not deferred behind a
+    // setTimeout) since Disney+'s own bundle can fire and parse its playback manifest
+    // request within the same tick the page script finishes loading. A deferred install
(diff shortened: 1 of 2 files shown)


ShanaryS:
The pause changes in `Binding.ts` were fixed in #1089, so I believe the only thing remaining for this PR is the race conditions.

Issue #1059, opened 2026-06-28
[Feature Request] Auto-expand/merge subsequent subtitle lines until sentence-end punctuation (. ! ?) during mining


buzzshima:
### Is your feature request related to a problem? Please describe.
When language learning (Sentence Mining), sentences are often split across multiple subtitle lines/timestamps. For example:

**Line 1 (Currently on screen):** "Oh, you finally" (Mining here)
**Line 2:** "put aside all that"
**Line 3:** "supervillain nonsense,"
**Line 4:** "and it really paid off." (Sentence ends here)

Currently, if I press the mining hotkey on Line 1, it only captures that single split line. I want the whole sentence combined into one card, but manually dragging across multiple lines in the sidebar or using the "manually record" shortcut (which relies too much on precise timing) disrupts the learning flow and is not practical.

This directly relates to the pain point discussed in **Issue #767**, but instead of manually selecting lines via keyboard, I am proposing a completely automated approach.

### Describe the solution you'd like
I would like an option to **"Auto-expand selection to sentence end"**. 

When this feature is enabled and I mine a subtitle line, asbplayer should automatically check the subsequent subtitle lines, parse their text, and automatically merge them (both text and audio timestamps) until it detects a sentence-end punctuation mark (such as `.`, `!`, `?`, `！`, `？`). 

In the example above, pressing the hotkey on Line 1 would automatically capture everything from Line 1 to Line 4 in a single card, hands-free.

### Describe alternatives you've considered
- Manually dragging cursor over multiple lines in the side panel (disrupts immersion).
- Using "manually record" shortcuts, but it is impossible to get the timing perfectly right without cutting off the beginning or end.

### Additional context
Since asbplayer is a tool dedicated to language learning, automating the merging of split sentences based on punctuation would be a massive quality-of-life improvement for anyone doing Sentence Mining. It would perfectly solve the underlying issue in #767. Let me know your thoughts!

buzzshima:
Oops, my bad. Here is the correct description.

killergerbah:
I like this solution. I wonder how configurable it should be since there may be situations where a "sentence" should be demarcated by something other than punctuation

This would fix https://github.com/asbplayer/asbplayer/issues/396

ShanaryS:
[...]

Issue #640, opened 2025-02-04
Increase the amount of subtitle length covered on Youtube


CroakingToadElf:
While watching and mining youtube, it's great that you can pad the length of the audio automatically in the settings. 

What's troublesome though, is that the standard length of the subtitles rarely seems to fit the length of the audio. It usually cuts off too early, leaving the words unfinished.

For example: I want to mine 遜色ない
The subtitle on youtube ends on 遜色. Thanks to the padding option I can at least get the audio fully, but I rarely get the full sentence. 
This is the subtitles rendered by asbplayer

"大手ゲーム会社の社長と比較しても遜色

ないどころか任天に近いだそうですちなみ"

I know you can do it manually, but that process turns a quick mining action into a lengthy one.

This is especially problematic because 遜色ない is a common phrase and yomichan wouldn't recognize it if I wasn't simultaneously using the youtube native subtitles.

Is there a setting I am missing, otherwise I would love this as a feature.

ZornHadNoChoice:
I want to suggest some ways to fix this:

1. The subtitle "shifts" with the audio so that words keep moving left (or right for RTL languages) and fading away.
2. The previous subtitle is shown alongside the current one.

In both cases, we should be able to mine all the visible subtitles with a shortcut.

PR #1101, opened 2026-07-30
Feature: a simple solution helps Yomitan pick up a full sentence when a sentence spans multiple subtitles

BilderLoong:

# Why
This change helps **Yomitan** to be able to pick up a full sentence when a sentence is broken up between subtitles.

# How
I added the previous subtitle text and the after subtitle text as a hidden span for Yomitan to pick up. Now it is just a prototype. I haven't added any settings to configure this functionality.

----
I don't know whether this is a good direction to solve this problem, but now it works quite well for me. Feel free to provide any advice to help solve this problem.

# Related
Partially related to: #640, #1059, etc.

### DIFF
--- a/extension/src/controllers/subtitle-controller.test.ts
+++ b/extension/src/controllers/subtitle-controller.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from '@jest/globals';
+import { IndexedSubtitleModel } from '@project/common';
+import { buildSubtitleContextText } from './subtitle-controller';
+
+const subtitle = (text: string, index: number): IndexedSubtitleModel => ({
+    text,
+    index,
+    start: index * 1000,
+    end: (index + 1) * 1000,
+    originalStart: index * 1000,
+    originalEnd: (index + 1) * 1000,
+    track: 0,
+});
+
+describe('buildSubtitleContextText', () => {
+    it('returns empty strings for single subtitle', () => {
+        const result = buildSubtitleContextText(0, [subtitle('A', 0)]);
+        expect(result).toEqual({ before: '', after: '' });
+    });
+
+    it('returns correct before and after text', () => {
+        const subs = [subtitle('A', 0), subtitle('B', 1), subtitle('C', 2)];
+        const result = buildSubtitleContextText(1, subs);
+        expect(result).toEqual({ before: 'A', after: 'C' });
+    });
+
+    it('trims trailing space from subtitle text', () => {
+        const subs = [subtitle('A ', 0), subtitle('B', 1)];
+        const result = buildSubtitleContextText(1, subs);
+        expect(result.before).toBe('A');
+    });
+
+    it('trims leading space from subtitle text', () => {
+        const subs = [subtitle('A', 0), subtitle(' B', 1)];
+        const result = buildSubtitleContextText(0, subs);
+        expect(result.after).toBe('B');
+    });
+
+    it('prevents double space at before boundary', () => {
+        const subs = [subtitle('A ', 0), subtitle('B', 1)];
+        const result = buildSubtitleContextText(1, subs);
+        expect(result.before).toBe('A');
+        expect(result.before).not.toMatch(/ {2}/);
+    });
+
+    it('prevents double space at after boundary', () => {
+        const subs = [subtitle('A', 0), subtitle(' B', 1)];
+        const result = buildSubtitleContextText(0, subs);
+        expect(result.after).toBe('B');
+        expect(result.after).not.toMatch(/ {2}/);
+    });
+
+    it('joins multiple before subtitles with single space', () => {
+        const subs = [subtitle('A', 0), subtitle('B', 1), subtitle('C', 2)];
+        const result = buildSubtitleContextText(2, subs);
+        expect(result.before).toBe('A B');
+    });
+
+    it('joins multiple after subtitles with single space', () => {
+        const subs = [subtitle('A', 0), subtitle('B', 1), subtitle('C', 2)];
+        const result = buildSubtitleContextText(0, subs);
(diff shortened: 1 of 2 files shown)


ShanaryS:
This adds a lot of hidden data to the DOM as this is per index, even though it would probably be fine performance wise. I think this could be limited to only on hover, and with our own processing of the sentence boundaries precomputed for each index. This way the hidden data is minimal and only loaded when needed.

But I also wonder if this would be necessary if asbplayer can handle the context itself. If we improve asbplayer's handling of sentences across subtitles as the other issues state, I think we could avoid this by just having the user update the last card with asbplayer data. Would there be a reason to need yomitan to handle the sentence properly without having asbplayer update it?

BilderLoong:
> This adds a lot of hidden data to the DOM as this is per index, even though it would probably be fine performance wise. I think this could be limited to only on hover, and with our own processing of the sentence boundaries precomputed for each index. This way the hidden data is minimal and only loaded when needed.

Thanks for the advice. 
1. I think it is a good idea to load the context only on hover and unload the context on unhover.
2. For the sentence boundaries, I'm more inclined to leave it to Yomitan to handle it because different languages have different strategies for defining sentence boundaries. In my setup, I have different Yomitan profiles for different languages, which are configured with different `Sentence termination characters`.

--- 

> But I also wonder if this would be necessary if asbplayer can handle the context itself. If we improve asbplayer's handling of sentences across subtitles as the other issues state, I think we could avoid this by just having the user update the last card with asbplayer data. Would there be a reason to need yomitan to handle the sentence properly without having asbplayer update it?

3. Sorry, I don't get what you mean by this? Do you mean restructuring the subtitle according to some rules to make the displayed subtitle have better sentence boundaries? Could you elaborate a bit more?

ShanaryS:
> For the sentence boundaries, I'm more inclined to leave it to Yomitan to handle it

I was referring to your code calling `.substring` every time we need the context, rather than caching that value per index. But I think it may actual not be worth it without proper benchmarking to see the performance/memory tradeoff.

> Sorry, I don't get what you mean by this? Do you mean restructuring the subtitle according to some rules to make the displayed subtitle have better sentence boundaries? Could you elaborate a bit more?

The fundamental issue that this PR tries to address and the issues relating to it is that a subtitle event may not be the full sentence and thus missing context. I believe Yomitan needs to use the sentence termination characters since it works on arbitrary text, not just subtitles. But asbplayer specializes in subtitles, so to me the proper solution is to have asbplayer handle it correctly rather than trying to trick Yomitan by using hidden data.

The other issues have proposed solutions which I think would be sufficient the majority of the time. That is using punctuation to determine what should be the sentence (want you are forcing Yomitan to do) rather than only taking the subtitle event at face value. This coupled with some heuristics either bounded by the number of subtitles events to span and/or the time range for maximum context.

This would solve the issue for everyone, as long as asbplayer gets the final say for the sentence field. To me this feels natural since anyone would expect asbplayer to be the best tool for mining subtitles and parsing such edge cases since it's specialized to subtitles. So going through asbplayer isn't much of an ask over modifying the DOM so that others can handle it, especially since the users are using asbplayer anyways.

BilderLoong:
Yes, the current implementation is more tailored to the Yomitan (or similar). It doesn't improve the asbplayer itself. If we let asbplayer reconstruct subtitles, there is a limitation to this. We can only concatenate subtitles, but can't split subtitles cause most subtitles don't have word-level timestamps. 

Such as, if punctuation appears in the middle of a sentence.

- "Hello. What's your name? My"
- "name is Bar. Where are"
- "you from? I'm from China"

As far as I can tell, the best thing we can do with the above subtitles is concatenate them into a big one:
- "Hello. What's your name? My name is Bar. Where are you from? I'm from China"

An integration with ASR models to generate better subtitles with word-level timestamps could be a solution, but that's a big change.

ShanaryS:
> If we let asbplayer reconstruct subtitles, there is a limitation to this. We can only concatenate subtitles, but can't split subtitles cause most subtitles don't have word-level timestamps.

Doesn't this same limitation apply to Yomitan?

> As far as I can tell, the best thing we can do with the above subtitles is concatenate them into a big one

My understanding is that we would only expand past the current subtitles and never shrinking based on punctuation. I think that becomes very tricky due to audio and lacking word by word timings like you mentioned above.

> An integration with ASR models to generate better subtitles with word-level timestamps could be a solution, but that's a big change.

Yeah I would want to integrate this feature with any ASR implementations in asbplayer as there is a few features that could benefit from it.


My comment about having asbplayer handle it is mainly to avoid mutating the DOM while the user still gets all the benefits with no downside. It's not clear to me how mutating the DOM so that Yomitan handles the parsing is better than if asbplayer does it by itself.

Since users have to use asbplayer to mutate the DOM for Yomitan to see the benefits, that means it should be no issue for them to just have asbplayer update the sentence on the card directly. Now it wouldn't help in other cases outside of mining, but I'm not sure if you need that behavior. If not, then I think keeping it within asbplayer makes the most sense.

BilderLoong:
> Doesn't this same limitation apply to Yomitan?

I don't think it applies in the same way if Yomitan is only responsible for mining the sentence text. Yomitan doesn't need to preserve the subtitle timestamps when determining the sentence boundaries.


> it should be no issue for them to just have asbplayer update the sentence on the card directly.

Yeah, I think this could be better. If asbplayer only updates the sentence text on the card, we also don't need the word-by-word timing.

> Now it wouldn't help in other cases outside of mining, but I'm not sure if you need that behavior. If not, then I think keeping it within asbplayer makes the most sense.

I currently need this behavior because too many subtitles have incomplete sentences. But if the DOM mutation method isn't a good way to solve this problem, I can keep a temporary fork for my own use.

---

Thanks for taking the time to talk about this. I will close this PR for now.

PR #1103, opened 2026-07-31
Stricter subtitle sanitation

### LINKED ISSUE #1102: Missing clarification about CVE-2025-69771     and general .srt and .ass exploit safety
Hello,

I should first emphasize how much I love this add-on, it's indispensable in my Japanese learning. Great work!

However, I'm a bit worried about the safety of using downloaded subtitle files (I'm on Windows) with this add-on. I download everything off jimaku.cc only, and I always try to analyse the file contents in Notepad, but I'm really not a malware expert at all!
(((I tried using a VM as an extra defense layer just for the extension, but the performance of watching videos is awful, so it's not viable and I'm stuck using my main system)))

What worries me even more is that ASBplayer officially had a CVE, and that this hasn't (as far as me and chatgpt could search) been officially acknowledged here on Github or publicly marked as patched.
I'd assume that this has all been patched ''silently'', because the CVE only talks about versions of this extension that are already like 1,5 years old by now, but the CVE information was never updated.

_Please don't interpret this as a personal attack, I just think it's a good topic to get some clarity about. I'm just genuinely curious and openness of communication helps everyone in the community.

So, just to help you, there are two (but intertwined) issues here:
1. CVE-2025-69771 and how this was fixed. is any suspicious code in the subtitle file automatically sanitized?
2. Malware risks of using .srt and .ass files with this extension even when it's fully up-to-date

*Using Chrome, by the way. In Firefox, my main browser, subtitles don't show up in full screen, but that's another issue and I found a workaround by using Chrome instead, even though I don't like that browser.

ShanaryS:
fixes #1102

This strengthens the original fix in #905 by:
- Updating `DOMPurify` as the previous version had an `mXSS` vulnerability
- Strict allow list over the default `DOMPurify` config, asbplayer only needs presentation html, attributes such as `` could be used for tracking.
- Apply the sanitization as the final step after all parsing to prevent `mXSS` opportunities
- Added regression tests

The other source we should look at is the data from Yomitan (which mainly comes from the user's dictionaries). We do verify that the subtitle text remains the same after tokenization so the base text is safe, but the readings could exhibit the same issue. This is of course far less problematic than arbitrary subtitle files and may be fine implicitly trusting it so I left it out. We could easily reject any reading that contains html since that's obviously invalid.

Assisted-by: Codex:GPT-5.6, Claude:Fable-5

### DIFF
--- a/common/subtitle-reader/subtitle-reader.ts
+++ b/common/subtitle-reader/subtitle-reader.ts
@@ -6,6 +6,19 @@ import { XMLParser } from 'fast-xml-parser';
 import { SubtitleHtml, SubtitleTextImage, Token, Tokenization } from '@project/common';
 import DOMPurify from 'dompurify';
 
+/**
+ * Subtitle files are untrusted input.  Keep this list deliberately small: subtitle
+ * markup is for presentation only and must not be able to navigate, fetch, or run
+ * code.  In particular, do not add `style` or URL-bearing attributes here.
+ */
+export const sanitizeSubtitleHtml = (html: string) =>
+    DOMPurify.sanitize(html, {
+        ALLOWED_TAGS: ['b', 'strong', 'i', 'em', 'u', 's', 'del', 'br', 'ruby', 'rt', 'rp', 'span'],
+        ALLOWED_ATTR: [],
+        ALLOW_DATA_ATTR: false,
+        ALLOW_ARIA_ATTR: false,
+    });
+
 const vttClassRegex = /<(\/)?c(\.[^>]*)?>/g;
 const assNewLineRegex = RegExp(/\\[nN]/, 'ig');
 // Character classes shared by the Netflix ruby regexes below so they cannot drift apart.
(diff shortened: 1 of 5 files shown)


common/subtitle-reader/subtitle-reader.ts
killergerbah: Not completely related to this change but to simplify it you could do something like this

```
        if (flatten) {
          allNodes = deduplicate(allNodes);
        }

        if (this._convertNetflixRuby) {
          for (const node of allNodes) node.text = node.text.replaceAll(netflixRubyBaseMarker, '');
        }

        // Sanitize after all parser, filter, decoding, and flattening transformations.
        // Ruby tokenization runs afterwards because it relies on positions in this
        // sanitized text and does not introduce any new markup.
        for (const node of allNodes) {
          node.text = sanitizeSubtitleHtml(node.text);

        if (this._convertNetflixRuby) {
            for (const node of allNodes) this._convertNetflixRubyToHtml(node);
        }

```

The `flatten` behavior was a little broken to begin with. All it needs to do is put everything in the same track and deduplicate but it was skipping a bunch of other logic because Netflix ruby is assumed to never come from websites that require flattening. In the end it just made the logic more complicated.

killergerbah (approved):

common/subtitle-reader/subtitle-reader.ts
ShanaryS: I had to modify your suggestion a little but I think it accomplishes the intent:

1. Ruby conversion is always performed regardless of flatten
2. This means that we no longer need `node.text.replaceAll(netflixRubyBaseMarker, '')` since that's handled by ruby conversion
3. `deduplicate` happens at the end as sanitization could make nodes equal (or other parsing).

PR #1108, opened 2026-08-03
Update playback settings init, notifications, and state updates

### LINKED ISSUE #1110: App: Playback state popup overlays subtitles
@ShanaryS  In https://github.com/asbplayer/asbplayer/pull/1089 a popup with playback states was added. I appreciate the work that goes into this, but I have a slight annoyance with the new popup - it overlays the subtitles. So I think improvements are possible here, so I'm curious about what you think
- (probably simplest) move the popup a bit higher, so it sits above the subtitles
- as it stands, when changing playback mode, there's notification popups at the top, and also state popup at the bottom. This causes duplication and visual noise. I'd argue that we don't need both: top popups could be removed, and the state popup moved from bottom to top (need to make sure it doesn't overlay the subtitles though)
- (better yet) we could float the new pop-up to the right, in the middle of the screen. By doing so, we could standardise its position, whether it was evoked by pressing a playmode shortcut, or by clicking the icon on the bottom bar

ShanaryS:
fixes #1110

These do a few things, mainly cleaning up things around playback settings and their notifications.

- Move playback preferences to `playback/controllers/playback-preference-controller.ts`
  - Offset is now managed by `PlaybackEngine`
  - All other preferences are managed separate from `PlaybackEngine` as they concern Player semantics
- Using `settingsProvider` to directly init `PlaybackEngine`
  - `PlaybackEngine` needs to know exactly when settings are ready in order to correctly transition between states
  - `Player` and `VideoPlayer` init with default settings and don't get their final until 2 `settingsChanged()` calls later. Binding is explicitly created with default settings and the first `settingsChanged()` call give the final settings. Instead of working around this, we just manually await the settings so it's consistent and future proof
- `saveOnly` option is removed from `setSettings` as coordinating it across multiple instances would take too much work to ensure it's honored
- `saveOnlyKeys` is used instead where specific keys are designated to never emit events, for now these are the internal playback settings of `PlaybackEngine`
- Now displaying the offset and playback rate notification on load if it's different from `0ms` or `1x` along with the playback modes
  - These notifications are persistent for `6s` instead of the normal `3s`
- Using explicit `playModesDisabled` flag for `SyntheticPlayback` rather than repurposing the `playModesSuppressed` flag meant for temporary reverting to normal mode (such as during recording).
- Fixed potential race condition when saving playback postitions for multiple videos at once
- Removed the temporary playback mode selector, the new notifications solves this behavior (#1110)
- Now using authoritative `PlaybackState` to control showing subtitles rather than based it off clock in the local context
  - This ensure that highlighted subtitles on the subtitle list are prefectly consistent with `PlaybackEngine`'s intentions (e.g auto-pause/repeat don't flash next subtitle)
  - `settime` on clock is now atomic with whether it should also `start`/`stop`
- `timeupdate` is too slow (Chrome in fact always fires at 4Hz (250ms)) and so we will also poll the time at 50ms in addition to timeupdate when audio-only/hidden. The `timeupdate` events exist as a fallback as timers are unreliable when hidden.

- [X] I have read the [contribution guidelines](https://github.com/asbplayer/asbplayer/blob/main/CONTRIBUTING.md).

Assisted-by: Codex:GPT-5.6


common/settings/settings.ts
```diff
@@ -8,9 +8,17 @@ import {
 } from '../src/model';
 import { arrayEquals } from '../util';
 
-export interface SaveSettingsOptions {
-    readonly saveOnly: boolean;
-}
+// Settings visible in UI probably shouldn't ever be here to prevent user confusion.
+export const saveOnlySettings: readonly (keyof AsbplayerSettings)[] = [
+    'lastSubtitleOffset',
```
killergerbah, line 16: `lastSubtitleOffset` was added as a setting before we had `global-state`. To be honest I think all of these keys are better off as global state since it doesn't really make sense your `lastSubtitleOffset`, `lastPlaybackModes` etc  should be dependent on your settings profile.

It might be too late to be suggesting this but another way to achieve the goal of this PR would be to switch these values to `global-state` and make all `global-state` updates from the app `saveOnly`. The current method works too though.

common/app/components/VideoPlayer.tsx
```diff
@@ -648,22 +681,51 @@ export default function VideoPlayer({
...
+                initialPlaybackSettingsChanged: ({
+                    autoHideDuration,
+                    playbackRate,
+                    playbackRateNotificationEnabled,
+                    subtitleOffset,
+                    fastForwarding,
+                    playbackModeTransition,
+                    join,
+                }) => {
+                    const offsetAndRateNotifications: string[] = [];
```
killergerbah: This code looks like a less functional version of the `Player.tsx` `initialPlaybackSettingsChanged` callback. And both of these look very similar to the one in `binding.ts`. I wonder if there's a way to deduplicate some code.

common/app/components/VideoPlayer.tsx
```diff
@@ -483,15 +485,33 @@ export default function VideoPlayer({
     const onErrorRef = useRef(onError);
     onErrorRef.current = onError;
 
+    const showAlerts = useCallback((notifications: readonly AlertNotification[]) => {
+        setAlertNotifications([...notifications]);
+        setAlertOpen(true);
+    }, []);
+
+    const showInfoAlert = useCallback((message: string) => showAlerts([{ message, severity: 'info' }]), [showAlerts]);
```
killergerbah: Nitpicky suggestion, you could get a helper function like this automatically if you combine `alertOpen` with `alertNotifiations` into a single state object.
Then you can do stuff like `setAlert(alert => ({open: true, notifications: [{message, severity: 'info']}))`

common/app/components/VideoPlayer.tsx
```diff
@@ -483,15 +485,33 @@ export default function VideoPlayer({
...
+
+    const formatOffsetNotification = useCallback((offset: number) => {
```
killergerbah: Is there a way to deduplicate this code? Some version of it exists across `Player.tsx`, `VideoPlayer.tsx` and `binding.ts`. Maybe bring it into `PlaybackEngine` which already mandates some formatting like the join splitter?

killergerbah:
I'm seeing this on commit `fe2a6e4d7362b900459b75217b065997a5da12c5`

image

Also, not able to load a video:


https://github.com/user-attachments/assets/23cc2f8f-68d7-4475-95b9-0bc223d6be0e

common/settings/settings.ts
ShanaryS: Yeah I think it's too late to change this since these settings already exist in the extension snapshot which I consider final, so the change would have to be worth that migration.

I do think scoping to profile does have value if there are multiple users on the same system using different profiles. They could theoretically watch the same content so the playback positions could differentiate between them. The last playback mode would also be a reflection of their preferences which could differ.

ShanaryS:
> I'm seeing this on commit fe2a6e4d7362b900459b75217b065997a5da12c5

I couldn't replicate this myself but could in the new test so it should be fixed now.

ShanaryS:
I removed the temporary play mode selector that appears on startup and when changing modes. With the new notifications this is less necessary and simplifies a lot of the code around the selector and mobile overlay controller.

The one area this change is deficient in is when a playback mode is changed, the user will no longer see all the current modes enabled. But I don't think this is a big enough concern now that the notifications display them upon load. Also, the user won't be able to change other playback modes through the use of the mouse after triggering with a keybind.

killergerbah:
the infinite re-render seems to be fixed but I'm still getting `VideoPlayer` to crash


https://github.com/user-attachments/assets/bd9831db-ffaa-44ab-943b-3d446e4c1c3b

I'm on commmit `36606bfbd19875729a0d386b65260f4325a43975`

I tried debugging a bit and couldn't find the exception. When breakpointing at the bottom of the `render` function there wasn't a crash, however `PlaybackEngine` looks like it's missing the playback plan

image


Pasting top and bottom as well in case it helps

image

image

I can try to help debug more in a few days, it's a long weekend and I'll be out of town for a little bit

ShanaryS:
> however `PlaybackEngine` looks like it's missing the playback plan

I think this was the cause, the plan should be valid at all times now. We have a specific ready flags so there is no reason to ever be in an invalid state, even if it means default settings.

ShanaryS:
I went ahead and implemented an authoritative `PlaybackState` to control showing subtitles rather than based it off clock in the local context. This was always the goal to ensure the highlighted subtitles on the subtitle list are perfectly consistent with `PlaybackEngine`'s intentions (e.g auto-pause/repeat don't flash next subtitle), the same way `SubtitleController` and `VideoPlayer` naturally are.

I also made `Clock.setTime` take in the paused state so that it can set it atomically preventing any slight inaccuracies. Finally, in addition to `timeupdate` for audio-only/hidden, we will now also poll the time at `50ms` intervals as `timeupdate` is too slow (Chrome in fact always fires at `4Hz` (`250ms`)). This shouldn't matter as much with the showing subtitles change, as audio-only/hidden does not give granular enough feedback for the user to notice. `timeupdate` is still used instead of polling exclusively since timers can behave weird when hidden in the browser.

killergerbah:
I just got home like 30 minutes ago so haven't had time to look at this closely. I did try the diff again locally (2eba1adef43aeca1ebaa599a81917caf5b2b1845) and the video player is still crashing. I've been able to narrow the reproduction steps a bit - the crash only happens in incognito mode if I load my settings.

common/playback/controllers/playback-position-controller.ts
killergerbah: Not related to this specific diff but why 50? Not sure if the value of remembering 50 videos back is justified by the storage space for 50 longish strings in each profile

ShanaryS: This was lifted from the original PR implementing the playback positions.

ShanaryS:
> the crash only happens in incognito mode if I load my settings.

I could not reproduce on Chrome 150.0.7871.114 Linux

Edit:

I can on Firefox

ShanaryS:
The crash is from commit 46da65e8b296f90789d3357d6b69d27d2147e421 and is triggered by changing the language from english. I can reproduce by:

1. Private browsing in Firefox (I cannot reproduce in chrome)
2. Load a video which works
3. Change language from english
4. Load the same video which crashes

The commit introduced `formatPlaybackRateNotification` which depends on `t`, this causes `updatePlaybackRate()` which is called by `playerChannel.onPlaybackRate` to change when the language is not english closing the `PlayerChannel`.

The simplest fix for this is to use a ref for `updatePlaybackRate()` but that just means the failure will happen again in the future. I'll leave it to you come up with a better lifecycle for `PlayerChannel` with the react deps since I'm new to react.

killergerbah:
Thanks for looking into it, I pushed the localization down a bit so that the callback doesn't need to depend on `t`. It's possible that the other notifications could also just pass the loc keys down instead of prelocalizing as well.

common/playback/controllers/playback-position-controller.ts
ShanaryS: I dropped it down to 25 as I don't think people will need history for mining that long. We can change this number at any time and the next save will trim as needed. I also dropped the minimum time from `1 minute` to `30s` as I feel that's a better user experience. Finally I made it so that playback positions are no longer saved after the last subtitle rather than just the video ending.

killergerbah (commented):
It looks like with the playback state update the subtitle list no longer renders the current subtitle when synced with a streaming video across the extension
Otherwise I don't see any obvious issues

ShanaryS:
> It looks like with the playback state update the subtitle list no longer renders the current subtitle when synced with a streaming video across the extension

It's working for me, maybe your extension isn't updated and isn't sending it? Actually I think we do need to address this due to the extension snapshot not sending it out either. I also just found while it gives the latest state to late subscribers, in the case of opening the SidePanel late it needs a new state before it starts highlighting.

I think using clock instead of state until the first playback state comes in should fix both.

killergerbah:
> It's working for me, maybe your extension isn't updated and isn't sending it? Actually I think we do need to address this due to the extension snapshot not sending it out either. I also just found while it gives the latest state to late subscribers, in the case of opening the SidePanel late it needs a new state before it starts highlighting.

Yeah probably, I didn't rebuild the extension, but I guess we need some kind of branching logic somewhere because we would need to deploy the app without breaking ppl on 1.19 and earlier.

> I think using clock instead of state until the first playback state comes in should fix both.

Sort of related, not a regression with this diff but it is something I just noticed when testing with non-YT subtitles that have gaps. The SubtitleList will now highlight nothing if the current timestamp is in a subtitle gap, which sort of makes sense but IMO decreases the usability of the `SubtitleList` since it makes it less obvious where the current playback actually is. IMO we should restore the previous behavior or at least ensure that we have a way to restore it. I think the PlaybackState protocol should support this already but just mentioning this in case it affects any implementation decisions now.

ShanaryS:
> Yeah probably, I didn't rebuild the extension, but I guess we need some kind of branching logic somewhere because we would need to deploy the app without breaking ppl on 1.19 and earlier.

What it was doing before was guarding `PlaybackState` by extension version, so it would work with an outdated extension using clock. The issue was if you were on 1.20 and using an outdated commit (e.g the snapshot or your case). 

> The SubtitleList will now highlight nothing if the current timestamp is in a subtitle gap

Yeah I missed this function of the old using `lastShown`. It now uses `lastShown` when there are otherwise no indexes to highlight from `PlaybackState`.

killergerbah:
> What it was doing before was guarding PlaybackState by extension version, so it would work with an outdated extension using clock. The issue was if you were on 1.20 and using an outdated commit (e.g the snapshot or your case).

Oh got it, I missed that

killergerbah (approved):
Lgtm, feel free to merge when ready

PR #1111, opened 2026-08-07
Play Dolby Audio (E-AC-3, AC-3, DTS) via ffmpeg in Web Player

rookes:
*This update adds some support for ffmpeg to the web player. ffmpeg support is an issue that has been discussed (for example, in [this issue](https://github.com/asbplayer/asbplayer/issues/329) and [this issue](https://github.com/asbplayer/asbplayer/issues/709)). However, for various reasons it hasn't been included yet. I made the decision to include it as an external library at runtime for the web player only. Updating the extension was out of scope, as it would require embedding ffmpeg directly rather than using a CDN to download. Please feel free to request an extension update as well if necessary.*

# The Basics

With this update, when the user loads a video file with audio that isn't supported by the browser, a little prompt appears to ask the user if they want to convert the audio. If they say yes, ffmpeg is loaded and transcoding runs on the audio file. A loading bar shows progress while the user waits. The user can cancel this process at any time and resume playback normally. After the audio is re-encoded, users can play the video with audio as if it were natively playable, and use the audio for other functions such as Anki exports.

image
image

**We only prompt the user if it will actually help.** The prompt displays only if all of the following are true:

1. The video is playable in the current browser (just not the audio).
2. The video has an E-AC-3, AC-3, DTS, TrueHD, or MLP audio codec that requires transcoding to play in the current browser.
3. The video has no other playable audio tracks to fall back on.
4. If the user hasn't downloaded ffmpeg yet, then they must be currently online.

If the user dismisses the prompt, then ffmpeg is not loaded and there is no change to the existing functionality. Otherwise, if the user opts to convert, then ffmpeg is downloaded (if necessary), and the transcoding process begins.

### Transcoding Basics

* The preferred audio output encoding for transcoding is **Opus** (fallback to Vorbis, then AAC if necessary). I selected Opus after testing various other options because it appears to be the most performant by a significant margin. Encoding takes up the majority of the transcoding time, so this is an important optimization.
* Surround sound audio is downmixed to stereo.
* If `deviceMemory` doesn't report less than than 4 GB, there are ≥ 4 cores available, and the audio is ≥ 120 s and ≥ 45 s per projected chunk, then audio is chunked into up to 4 pieces and re-spliced with parallel transcoding for a big performance increase (~2.8x). Single-pass is enabled as a fallback. Asynchronous, streamed transcoding during playback could be implemented in the future for even more responsiveness. *Note: splicing the audio may induce a small amount of delay (~24ms after 3 splices).*
* Once transcoding is complete, the old audio is simply muted and a separate `<audio>` element is created for the transcoded audio. Timing and volume consistency are ensured by a drift watchdog process (250 ms polling, 150 ms dead zone, 0.5 s resync) .Web player controls updates are in `common/app/hooks/use-external-audio-track.ts`.
* After transcoding, the ffmpeg instances are terminated.

## Browser + Wrapper + Encoding Behavior Map

Here's a table of which codecs trigger the conversion prompt (in a container the browser can open):

| Audio codec | Chrome | Edge | Firefox |
|---|:---:|:---:|:---:|
| E-AC-3 | **Prompt** | — | **Prompt** |
| AC-3 | **Prompt** | — | **Prompt** |
| DTS | **Prompt** | **Prompt** | **Prompt** |
| TrueHD / MLP | **Prompt** | **Prompt** | **Prompt** |
| AAC, Opus, Vorbis, FLAC, MP3, PCM, ALAC | — | — | — |

Edge already supports EAC3 and AC3, so no need for a prompt there. We could consider adding a compatibility note for users on Chrome/Firefox in case they want to switch.

**The container gate comes first**. There's no prompt at all if the browser can't open the file. Container compatibility:

| Container | Chrome | Edge | Firefox |
|---|:---:|:---:|:---:|
| `mp4` / `m4v` | ✓ | ✓ | ✓ |
| `mkv` | ✓ | ✓ | ✓ |
| `avi` | ✗ | ✗ | ✗ |

The gate is whatever `canPlayType` reports, not a hardcoded list, so a browser that doesn't claim a container simply never prompts, and behaves exactly as it does today.

A file with *any* decodable track never prompts. An MKV with E-AC-3 **and** AAC already plays, because browsers fall back to a track they can decode.

# Modules Changed

* **A potential point of contention:** ffmpeg (WASM) is downloaded through a CDN at runtime. When the user requests a conversion for the first time, `@ffmpeg/core@0.12.10` is pulled from `unpkg.com` and cached in Cache Storage so no re-download is needed on subsequent use. In order to ensure code integrity, a SHA-256 check is performed before anything runs. I opted to use this third-party source for ffmpeg rather than bundling the 32MB directly into the web player because adding more than 10x filesize to the web player could cause unexpected strain on existing Github pages and other hosting solutions. Nonetheless, remote code is off limits in Chrome's [extension manifest](https://developer.chrome.com/docs/extensions/develop/migrate/what-is-mv3), so extension implementation would need to take a different approach.
* ffmpeg (`@ffmpeg/ffmpeg`) is now marked as a dependency in `common/package.json`.
* To determine what codecs are being used, I added a step to probe video files on load and check the wrapper, codec, and browser combination to determine next steps.
* Added helpers for MKV, AVI, and ISO-BMFF/MP4 wrappers to allow for consistent handling across browsers.

| Area | Change |
| --- | --- |
| `common/media-probe/` | **New.** Matroska/EBML, ISO-BMFF and AVI/RIFF header parsers. Reads only headers, so it's cheap on a 5 GB file. |
| `common/audio-transcode/` | **New.** Lazily-imported ffmpeg-wasm wrapper; parallel chunked encoding. |
| `common/app/hooks/use-transcoded-audio.ts` | **New.** Probe → prompt → convert → blob lifetime. |
| `common/app/hooks/use-external-audio-track.ts` | **New.** Keeps the `<audio>` glued to the video. |
| `common/app/components/AudioConversionModal.tsx` | **New.** Progress + cancel, modelled on `BulkExportModal`. |
| `Player.tsx` / `VideoPlayer.tsx` | Prompt, hidden `<audio>`, muting, volume routing. |
| `message.ts`, `video-channel.ts`, `player-channel.ts` | One new `transcodedAudio` message. |
| `audio-clip.ts`, `model.ts` | `FileModel.transcodedAudioBlobUrl`, preferred when recording. |

## Tests

Added unit tests to cover the new container parsers, the conversion prompt and track selection logic, and audio blob selection for Anki mining. See the new `common/media-probe/audio-codec-probe.test.ts` and updates to `audio-clip.test.ts`.

I have personally done end-to-end testing in Chrome, Firefox, and Edge. End-to-end testing included files up to 106 minutes in length, and I have confirmed that seek, volume, and mining audio work as before. I also verified ffmpeg download vs. caching, offline behavior, loading bar behavior, and prompt appearance behavior both personally and through Claude. My tests were run on Windows.

## Other Updates and Notes

* Updated 13 locale files with *English* versions of the various prompts needed for this PR. The contribution notes say that localization is handled through Crowdin, so I opted to keep English placeholders rather than machine translate these.
* Updated documents `docs/docs/compatibility.md` and `docs/docs/common-issues.md` with information on this change.
* **No settings update**. There are no new settings keys in this update, so that is left as-is.
* ffmpeg license is LGPL, while this project's license is MIT. Since we are loading ffmpeg externally, I assume this should be no problem, but noting here.
* If a video has multiple audio tracks, only the first is considered for conversion. The player's existing `AudioTrackSelector` relies on `HTMLMediaElement.audioTracks`, which Chrome, Edge and Firefox don't implement, so in exactly the browsers that need conversion, there's no track switching available anyway. Per-track conversion would need its own selector, which felt like scope for a follow-up rather than something to include in this PR.

#

Assisted-by: Claude:claude-opus-5

- [x] `yarn run verify` passes


killergerbah:
@rookes Thanks this is pretty cool. I don't think this can be merged as-is unless the `ffmpeg` build being downloaded is the LGPL version. Even if we knew that to be case, I would be scared of violating any copyleft licenses by accident because I don't understand the licenses completely. Probably the safest way to integrate something like this would be to ask the user to supply their own ffmpeg wasm binary, unless you can convince me this is completely safe from a legal point of view. Also, is it safe to include `@ffmpeg/ffmpeg` in the project? I didn't read the code but I'm assuming it's safe because it's MIT-licensed and probably includes only bindings to the actual binary.

rookes:
> @rookes Thanks this is pretty cool. I don't think this can be merged as-is unless the `ffmpeg` build being downloaded is the LGPL version. Even if we knew that to be case, I would be scared of violating any copyleft licenses by accident because I don't understand the licenses completely. Probably the safest way to integrate something like this would be to ask the user to supply their own ffmpeg wasm binary, unless you can convince me this is completely safe from a legal point of view. Also, is it safe to include `@ffmpeg/ffmpeg` in the project? I didn't read the code but I'm assuming it's safe because it's MIT-licensed and probably includes only bindings to the actual binary.

Yes, the JS wrapper `@ffmpeg/ffmpeg` is MIT so it should be fine to include. 

The WASM we're loading is `@ffmpeg/core@0.12.10`, which is ["GPL-2.0-or-later"](https://app.unpkg.com/@ffmpeg/core@0.12.10/files/package.json), so unfortunately not LGPL as-is. I actually think we should be clear even in this case - in my mind, using a third-party to load to cache like this is basically equivalent to asking a user to supply their own ffmpeg. The GPL2+MIT license version of the web player only ever exists on a user's private machine - it is not vendored by us and doesn't exist anywhere in the code we are distributing. Since GPL is about restricting distribution rather than usage, I think this should be OK. However, I would rather you not take my very unqualified legal opinion too seriously, and I defer to you all for the final decision.

Probably the correct way to go about this would be to just compile our own LGPL WASM version of ffmpeg. This would have the benefit of being much more lightweight, since we only need very specific features (x264, x265, webp etc. could be left out), but it would still add a few MB to the web player. We could even publish it to unpkg.com and use the exact same setup that this build uses (though perhaps there's another core build out there we could use that already accomplishes this). We are already using `lamejs`, which is LGPL-3.0, so unless that is also an issue, I think this should be fine.

So my question would be if we have an LGPL version, would that be enough reassurance to use for this change? Or would we rather take a different approach? I am open to setting it up as user-installed as well.

killergerbah:
> So my question would be if we have an LGPL version, would that be enough reassurance to use for this change? Or would we rather take a different approach? I am open to setting it up as user-installed as well.

Some changes would still be required since we would be considered a distributor of the binary. We would still need to link to the exact source code used to build the binary and give proper attribution to ffmpeg.

> open to setting it up as user-installed

I think this is optional because our source code is available for the user to reverse engineer and change the link but I will read through the license again (and consult with AI) and come back to you.

ShanaryS:
I honestly think the only path to integrating ffmpeg into asbplayer is by creating a custom build with everything disabled including the problematic licenses (building without `--enable-gpl` and `--enable-nonfree` makes it LGPL). asbplayer will only need a very small subset of ffmpeg features since we aren't using for general encoding/decoding. There are WASM builds that come out to ~5MB which is about what we need, only audio processing.

This current method is incompatible with the extension as noted. I don't thing it's feasible either to use a "remote" ffmpeg for the extension either through the site or native messaging with a user's ffmpeg locally. Sending the data and the input files needed across would be such a headache and I'm not even sure practical.

The only downside to integrating the size for the extension, the website is negligible since we should be able to defer downloading until needed. The extension would go from `4.5MB` to `~10MB`, which is probably worth it for all the tickets it will close. For reference, Yomitan is `15.1MB`. The only other issue is speed due to WASM over native, but it's better to have and be slow than not have it.

I think this method solves everything cleanly as it's best to develop with, best for the user, and easiest to handle licensing with since we control everything.

PR #1113, opened 2026-08-07
Prefer valid video elements for selector and auto sync

### LINKED ISSUE #1013: Duplicate Subtitle Instances on YouTube (v1.17.0, Brave browser)
https://github.com/asbplayer/asbplayer?tab=readme-ov-file#common-issues

When I load a youtube video multiple instances of ASBplayer subtitles are shown. One set of the subtitles functions normally, but the second set will be offset to one side of the video and continue to play whether or not the video is paused. This issue occurs on some youtube videos and not on others. 

**What browser?**

Brave (Chromium 148.0.0.0)

Streaming video

Extension version [1.17.0] 

Youtube

1. Failed to load resource: the server responded with a status of 403 ()

2. Failed to load resource: net::ERR_BLOCKED_BY_CLIENT

3. Failed to load resource: net::ERR_UNSAFE_REDIRECT

4. Error with Permissions-Policy header: Unrecognized feature: 'browsing-topics'

5. Error with Permissions-Policy header: Unrecognized feature: 'run-ad-auction'

6. Error with Permissions-Policy header: Origin trial controlled feature not enabled: 'join-ad-interest-group'

7. Error with Permissions-Policy header: Unrecognized feature: 'private-state-token-redemption'

8. Error with Permissions-Policy header: Unrecognized feature: 'private-state-token-issuance'

9. Error with Permissions-Policy header: Unrecognized feature: 'private-aggregation'

10. Error with Permissions-Policy header: Unrecognized feature: 'attribution-reporting'

11. Failed to execute 'postMessage' on 'DOMWindow':
    The target origin provided ('https://studio.youtube.com')
    does not match the recipient window's origin ('https://www.youtube.com')

12. Dictionary Anki cache build error (1 - undefined)

13. background.js:1 Uncaught (in promise) Error:
    Could not establish connection.
    Receiving end does not exist.

14. Banner not shown:
    beforeinstallpromptevent.preventDefault() called.
    The page must call beforeinstallpromptevent.prompt() to show the banner.

15. LegacyDataMixin will be applied to all legacy elements.
    Set `_legacyUndefinedCheck: true` on element class to enable.

16. youtubeHook.bundle.js:1 [5:36:54 PM] player hooked

17. Failed to load resource: /api/stats/qoe

18. Failed to load resource: /ptracking

19. Failed to load resource: /generate_204

20. Failed to load resource:
    googleads.g.doubleclick.net/pagead/id

21. Failed to load resource:
    www.google.com/pagead/lvz

22. Failed to load resource:
    rr4---sn-q4fl6nsk.googlevideo.com/videoplayback

Image

### LINKED ISSUE #1099: YouTube: asbplayer detects an extra blob video element on first page load and loads subtitles twice
Recently, asbplayer started detecting an additional video element on YouTube pages and automatically loading subtitles for both videos, even so the second element is not even a video.

The issue only happens when opening a YouTube video page for the first time. If I refresh the page, the problem disappears and asbplayer works normally, however it means that I need to refresh the page literally every time I open a new youtube video.

Steps to reproduce:
1. Open any YouTube video page.
2. Let asbplayer automatically detect the video and load subtitles. For some reason it will detect 2 elements at the same time and load subtitles for 2 of them.

There are only 2 workarounds, I could refresh the page every time I open a new video or i could press ctrl+shift+F combination and manually select what video element to use for every single video I watch.

Image
(the first blob element is not an actual video)

The issue does not happen after refreshing the page.

I tested this with all other browser extensions disabled. The issue still occurs with only asbplayer enabled.

This is what it looks like with "auto-load detected subtitles" option enabled. 2 asbplayer control boxes appear on the screen, one is correctly displayed, the second one should not even exist in the first place. Because of that it is just sitting there in the middle of the screen blocking an actual video.

Image

---

## What browser?

I tested it in Microsoft Edge and Brave browsers on 2 different devices (my windows 10 laptop and my windows 11 PC).

---

## Is it happening with local video or streaming video?

Streaming video. YouTube specifically.

---

## If the issue is happening with streaming video, what version of the extension are you using?

Latest one:  1.19.0
---

## Please provide logs from your developer console. At the very least, copy and paste the error from the "Copy Last Error" button (top-right) if the error is occurring in the asbplayer webapp.

No error messages appear.
The issue seems to be related to incorrect video element detection rather than an exception/error.

ShanaryS:
fixes #1013 fixes #1099

I found that adding a log in `shouldIgnore()`:
```ts
    shouldIgnore(element: HTMLMediaElement) {
        if (this.config.videoElementSelector !== undefined && !element.matches(this.config.videoElementSelector)) {
            console.log(`[asbplayer] Ignoring video element because it does not match the selector: ${this.config.videoElementSelector}`);
            return true;
        }
```

shows that this element starts showing up after hovering a over a thumbnail which triggers a preview. It then gets logged every `~1s`. This explains why it doesn't happen on refresh on a video page, but the `1s` trigger keeps happening after youtube goes to the video since the user has to hover and click the video.

This solution seems to be the best since we already target `#movie_player video` in the implementation, I did not find any other reliable way to detect regardless (but there were other differences).

This is the same issue that causes duplicate subtitles as found in #1013, to solve this we need to ensure that only a single binding can auto sync.


extension/src/services/pages.ts
```diff
@@ -146,6 +149,10 @@ export class PageDelegate {
     }
 
     shouldIgnore(element: HTMLMediaElement) {
+        if (this.config.videoElementSelector !== undefined && !element.matches(this.config.videoElementSelector)) {
```
killergerbah: Could this logic be switched to being preferential in case of multiple video elements instead of requiring it to match? Just worried if YT deploys a change to the HTML and causes this selector to start failing. Normally I wouldn't care that much but YT is probably the most-used service for asbplayer so being cautious feels justified

ShanaryS: Yes that makes sense. I was worrying about that too but was too focused on the always ignoring and missed the idea of priority.

ShanaryS: Moving to preference means we need to change how auto sync works since both would be auto synced. We will now only auto load a single video element rather than all for a frame. This also addresses #1013 which had the same root cause, the offset duplicate subtitles was due to the thumbnail positioning in the window at the time the user clicked on it from this invalid element.

The down side is that other sites may have multiple legitimate (or some invalid elements) where only a single one will be auto loaded. It's fine if there are multiple legitimate and only a single loads, but the concern is if the invalid element is selected. We would need explicit logic like for Youtube in that case.

I think we should do it regardless as the invariant of a single auto load makes intuitive sense to do and we should handle other sites on a case by case basis if they have issues. The preference does not need to be limited to css selectors, any arbitrary logic per site is fine.

The current logic also will switch to the higher priority element if it appears and unloading the previous. I think it makes sense to add a time limit to this but considering only Youtube has a priority it's always desirable. I opted not to further complicate this code but it could be important to set in the future.

EDIT:

Went ahead an implemented a `5s` time limit since it was straightforward

ShanaryS: I ended up reverting the preference changes. I could not get youtube to properly auto sync the right video element consistently. So instead I went with a different middle ground, auto sync won't happen for the currently known bad elements, but it will still be available for manual selection (below the valid element). This means if youtube were to change this, only auto sync would break but users would still be able to use it. Which is far better than both automatically loading.

extension/src/services/pages.ts
```diff
@@ -37,6 +37,9 @@ interface PageConfig {
     // Whether video elements with blank src should be bindable on this page
     allowVideoElementsWithBlankSrc?: boolean;
 
+    // CSS selector for videos that may auto-sync and should sort first in manual video selection
+    videoElementSelector?: string;
```
killergerbah: The naming in `PageConfig` is already pretty bad but I wonder if the name `videoElementSelector` can be improved to match its behavior. As it is right now, if the developer doesn't read the comment, would be easy to misinterpret the field as selecting a specific video element for binding.

extension/src/entrypoints/video.content/index.ts
```diff
@@ -167,6 +171,13 @@ export default defineContentScript({
                     }
                 }
 
+                if (page !== undefined) {
+                    bindings.sort(
```
killergerbah, line 175: I guess this works but it's interesting that AI went with this solution instead of just picking the first one that matches the selector

ShanaryS: How do you mean? I also wanted to reorder the elements so that the preferred one is at the top for the manual selector and the likely bad elements at the bottom. Determining preference then just reuses that work as it abstracts away what did the sorting.

killergerbah: I see, yeah that makes more sense. Sorry, I assumed that an agent wrote the sorting bit since it seemed like it was doing too much but I see that it brings the sorting into the dialog as well.

killergerbah (approved):

PR #1114, opened 2026-08-07
jellyfin/emby treat same url video as new

### LINKED ISSUE #1082: Jellyfin previous episode subtitles remain when selecting next episode
When switching to the next episode, the extension keeps using the subtitle tracks from the previous episode instead of loading the subtitles for the newly selected episode.

**What browser?**
Brave 1.91.180

Streaming

1.19.0

Jellyfin Web 10.11.11

ShanaryS:
fixes #1082

Also, `PlaybackEngine` now clears its subtitles whenever `Binding.reset()` is called which was a missed area for subtitles changing.

### DIFF
--- a/extension/src/services/binding.ts
+++ b/extension/src/services/binding.ts
@@ -152,6 +152,12 @@ const startAudioRecordingErrorResponse: (e: any) => StartRecordingResponse = (e:
     return errorResponse;
 };
 
+export interface BindingOptions {
+    readonly hasPageScript: boolean;
+    readonly frameId?: string;
+    readonly videoSrcChangesIndicateNewVideo: boolean;
+}
+
 export default class Binding {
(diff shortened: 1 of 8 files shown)


extension/src/services/binding.ts
```diff
@@ -692,30 +702,39 @@ export default class Binding {
         this.subtitleController.onMouseOut = (mouseEvent: MouseEvent) => this.hoveredToken.handleMouseOut(mouseEvent);
 
         if (this.hasPageScript) {
+            let forceRefreshForSameLocation = false;
```
killergerbah: I think its possible for `debounced` to allow functions with parameters, in which case this boolean wouldn't be necessary to manage outside

killergerbah (approved):

PR #1115, opened 2026-08-07
Deduplicate Youtube subtitles

### LINKED ISSUE #1018: Subtitles are duplicated in youtube / chrome
https://github.com/asbplayer/asbplayer?tab=readme-ov-file#common-issues

**What browser?**
Chrome
no
latest
youtube

Image

ShanaryS:
fixes #1018

I was able to replicate this on this youtube video https://github.com/asbplayer/asbplayer/issues/1018#issuecomment-5180040257. It seems it happens to all tracks and is 100% youtube sending the duplicate subtitles. Perhaps they had a temporary bug and haven't reparsed or it's some race condition.

Either way, the solution for us is simple. We just always run `deduplicate()` within a track before subtitle reader returns. This should always be desirable with no side effects.

### DIFF
--- a/common/subtitle-reader/subtitle-reader.ts
+++ b/common/subtitle-reader/subtitle-reader.ts
@@ -143,14 +143,14 @@ export default class SubtitleReader {
             for (const node of allNodes) this._convertNetflixRubyToHtml(node);
         }
 
-        return flatten ? this._deduplicate(allNodes) : allNodes;
+        return this._deduplicate(allNodes);
     }
 
     private _deduplicate(nodes: SubtitleNode[]) {
         const deduplicated: SubtitleNode[] = [];
 
         for (const node of nodes) {
-            if (deduplicated.length == 0 || !this._isSame(node, deduplicated[deduplicated.length - 1])) {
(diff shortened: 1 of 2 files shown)


common/subtitle-reader/subtitle-reader.ts
```diff
@@ -159,11 +159,8 @@ export default class SubtitleReader {
     }
 
     private _isSame(a: SubtitleNode, b: SubtitleNode) {
-        if (a.textImage || b.textImage) {
-            return false;
-        }
-
-        return a.start === b.start && a.end === b.end && a.text === b.text;
+        if (a.textImage || b.textImage) return false;
+        return a.start === b.start && a.end === b.end && a.text === b.text && a.track === b.track;
```
killergerbah, line 163: This feels like it could be expensive for very long subtitle files but if we're already doing it for Disney then I guess we're not doing anything worse than before

ShanaryS: I don't think it should be a big issue, we are basically doing `n-1` check with just the `start` for each subtitle, which is an int equality check. Most normal subtitles will differ by this alone.

PR #1117, opened 2026-08-10
Export / import all profiles at once.

### LINKED ISSUE #681: Exporting settings should export all profiles
I've recently restored a setting from the .json back up but realised that it only had settings for a single profile, whereas I was using 2 profiles

I think ideally, it'd be great to back up all user profiles, and likewise restore all the profiles when loading the .json file

Alternatively, at the very least, it should be made obvious to the user that all profiles will _not_ be synced. E.g. change the button text `Export settings` -> `Export current profile`

artjomsR:

Resolves https://github.com/asbplayer/asbplayer/issues/681


common/app/services/app-extension-settings-storage.ts
```diff
@@ -11,12 +11,14 @@ export class AppExtensionSettingsStorage implements AppSettingsStorage {
         this._extension = extension;
     }
 
-    get(keysAndDefaults: Partial<AsbplayerSettings>): Promise<Partial<AsbplayerSettings>> {
-        return this._extension.getSettings(keysAndDefaults);
+    // Only pass a profile when the extension supports it - older extensions ignore the field and
```
killergerbah, line 14: Should this comment be moved to where the version check is?

common/settings/settings-import-export.ts
```diff
@@ -739,13 +739,113 @@ const withIgnoredKeysRemoved = (settings: any) => {
...
+
+    for (const { name, settings } of imported.profiles) {
+        if (name !== undefined && !existingProfiles.has(name)) {
+            await settingsProvider.addProfile(name);
+        }
+
+        await settingsProvider.setForProfile(settings, name);
+    }
+
+    if (imported.activeProfile !== activeProfile) {
```
killergerbah, line 844: I think there might be a UI/UX issue to solve. Namely, if you import a profile with a name that you already have, then it gets silently skipped. Also, if it happened to be the active profile, you would switch that profile. The end result would be to switch to profiles but not import anything, which IMO should be surprising to the user.

EDIT: I looked at this again and the above is not true. All profiles are properly imported. But I still think there is a UX concern that could be addressed in another change.

Previously, the import/export feature did not need to handle any edge cases like this so there was no explanation in the UI about its behavior. But the edge case above + potentially importing lots of profiles justifies having a clearer UX.

IMO it would make sense to display a profile selector both when exporting and importing. Also, when importing, it should be obvious to the user that they are importing a profile name they already have, and that it could get overwritten, or skipped.

extension/src/services/extension-settings-storage.ts
```diff
@@ -27,26 +29,23 @@ export class ExtensionSettingsStorage implements SettingsStorage {
         this._storage = storage ?? browser.storage.local;
     }
 
-    async get(keysAndDefaults: Partial<AsbplayerSettings>) {
-        const activeProfile = await this.activeProfile();
+    async get(keysAndDefaults: Partial<AsbplayerSettings>, profile?: TargetProfile) {
```
killergerbah, line 32: Instead of introducing profile-specific getter and setter parameters I think you could get away with a lot less code change if the interface provides a single method to switch which profile it's reading/writing from without persisting the "active" profile. Then you could use the existing interface as-is to do all of your reading/writing, instead of expanding the interface with profile-specific getters and setters. The target profile could be maintained as additional state in the class.

Something like 

```
export interface SettingsStorage {
    // ...
    targetingProfile: (name: string | undefined) => SettingsStorage;
    // ...
}

```

where the returned SettingsStorage maintains internal "target profile override" state. Then you just have only one method to implement across `ExtensionSettingsStorage`, `LocalSettingsStorage` and `SettingsProvider`.

PR #1121, opened 2026-08-13
Better logging and type imports

ShanaryS:
- Added `asbLog`, `asbInfo`, `asbWarn`, and `asbError` functions to replace `console.*` logging
  - This prepends the logs with `[asbplayer][label]` so that it's easy to filter aspblayer logs and it clues the source of the error
  - Otherwise logging is the same, additional logs were not added outside of some UI error reporting that now also logs
- Require all imports to use `@project/...` rather than relative for consistency and easier updates when moving files around
- Require type imports when possible through a few eslint rules
  - We will require `import type { X }` instead of `import { type X }` as it has some benefits with compilation (https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/)

Runtime and side effect imports are preserved, there were not many and pretty much all these rule fixes were by eslint through `--fix`.


client/src/components/WebsiteApp.tsx
```diff
@@ -1,13 +1,13 @@
...
 import { useChromeExtension } from '@project/common/app';
 import RootApp from '@project/common/app/components/RootApp';
 import { useEffect, useMemo } from 'react';
 import { AppExtensionDictionaryStorage } from '@project/common/app/services/app-extension-dictionary-storage';
 import { AppExtensionSettingsStorage } from '@project/common/app/services/app-extension-settings-storage';
 import { AppExtensionGlobalStateProvider } from '@project/common/app/services/app-extension-global-state-provider';
 import { SettingsProvider } from '@project/common/settings';
-import { LocalDictionaryStorage } from '../local-dictionary-storage';
-import { LocalSettingsStorage } from '../local-settings-storage';
+import { LocalDictionaryStorage } from '@project/client/src/local-dictionary-storage';
```
killergerbah, line 9: Would it be possible to exclude `src` from these imports, e.g. `@project/client/local-dictionary-storage`? Having `src` in there seems a bit unnecessary since there is nothing above `src` that should be importable. Also, `src` doesn't look pretty when next to the other project imports that don't require it.

common/annotations/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from './dom-annotations';
-export * from './render-annotations';
-export * from './subtitle-annotations';
-export * from './token-collection';
+export * from '@project/common/annotations/dom-annotations';
```
killergerbah: Would it make sense for `index`-like files be excluded from this lint rule? It would be strange in this codebase for an `index` file to export anything that isn't under the same folder, or directly next to the file.

common/util/log.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Optional metadata for an asb logging call.
+ *
+ * When supplied as the first or last argument, `asbLogLabel` is removed from
+ * the message arguments and appended to the `[asbplayer]` console prefix.
+ * The prefixed name avoids colliding with ordinary payload properties such as
+ * `label` when these functions are used as console-compatible replacements.
+ */
+export interface AsbLogOptions {
+    /** The subsystem or operation name to include in the console prefix. */
+    readonly asbLogLabel?: string;
```
killergerbah: Just a thought, I still manually type a lot of code so I would prefer to write something like

```
asbWarn('Anki permission request failed:', e, 'annotations/anki');
```

instead of

```
asbWarn('Anki permission request failed:', e, {asbLogLabel: 'annotations/anki'});
```

or maybe

```
asbWarn('annotations/anki', 'Anki permission request failed:', e)
```

but I'm not sure how the label is supposed to be distinguished from the log in the example above

ShanaryS: We could, but it would break the one-to-one with console.log. I don't think it's a big deal to change it though as we can always just wrap it when we are using ...args.

ShanaryS: I would put the label first so that ...args is easier to use. It also makes label required which I also prefer.

killergerbah: Yeah that makes sense, especially if there aren't any situations where a label could be omitted. I can't think of any, since the logging file name is always a sensible option.

client/src/components/WebsiteApp.tsx
ShanaryS: This doesn't seem possible without creating maintenance burden of specifying new files being added or a complex linting setup with unintuitive rules.

killergerbah (approved):

PR #1122, opened 2026-08-13
redact exports ankiConnectApiKey and dictionaryWaniKaniApiToken

ShanaryS:
These two fields (especially for WaniKani) can leak sensitive information from users when sharing their config. It's best to simply never export these as they can always be easily retrieved from AnkiConnect or a regenerated WaniKani token.

This is implemented through a generalized ignore path which allows ignoring nested keys and through arrays. Even if the field exists in the imported json, we will ignore it and use the current settings.

### DIFF
--- a/common/settings/settings-import-export.ts
+++ b/common/settings/settings-import-export.ts
@@ -727,27 +727,85 @@ const settingsSchema = {
     },
 };
 
-const ignoreKeys: (keyof AsbplayerSettings)[] = [
-    'streamingPages', // Ignored due to security risk (e.g. disable CSP)
-];
+export type IgnorePathSegment = string | '*';
+export type IgnorePath = readonly IgnorePathSegment[];
 
-const withIgnoredKeysRemoved = (settings: any) => {
-    const copy = { ...settings };
-    for (const ignoreKey of ignoreKeys) {
-        delete copy[ignoreKey];
+const ignorePaths: readonly IgnorePath[] = [
+    ['ankiConnectApiKey'],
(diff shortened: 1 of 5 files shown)


common/settings/settings-import-export.ts
```diff
@@ -727,27 +727,85 @@ const settingsSchema = {
...
+    importedSettings: Partial<AsbplayerSettings>,
+    currentSettings: AsbplayerSettings
+): Partial<AsbplayerSettings> => {
+    const mergedSettings = { ...importedSettings };
+
+    if (!('ankiConnectApiKey' in importedSettings)) {
+        mergedSettings.ankiConnectApiKey = currentSettings.ankiConnectApiKey;
+    }
+
+    if (Array.isArray(importedSettings.dictionaryTracks)) {
```
killergerbah: Don't we lose the convenience of `ignorePaths` if we still need to repeat those fields as code in this function?

I wonder if there's a more general way to perform the merge? Maybe a "redacted" sentinel that would indicate the existing setting should be preferred?

ShanaryS: I suppose we would always want this behavior, of ensuring the user's current settings are preserved. In that case I think we can loop over ignorePaths and use it to add instead of remove if necessary.

killergerbah: The only use-case for importing the API keys might be if the user is importing their own file (which I guess is the majority of the time). But we're not exporting them by default anyway so ignoring them on import seems like a sensible default behavior as well if it makes the code nicer.

killergerbah (approved):

PR #1130, opened 2026-08-17
Add subtitle detection for Dreaming (app.dreaming.com)

ravensgulls:
## Summary

Adds subtitle track detection for Dreaming (https://app.dreaming.com — the Dreaming Spanish platform). The site's Shaka-based player renders subtitles internally and exposes no `<track>` elements, so this uses the site's own metadata API instead:

- New `dreaming-page` script: on each sync request it reads the video id from the `?id=` query parameter and fetches the site's `/.netlify/functions/video` API with the bearer token the site keeps in `localStorage` (present for the anonymous ephemeral accounts the site auto-creates, as well as for logged-in users).
- Prefers the inline WebVTT returned by the API (exposed as a `data:` URL, immune to the ~1h expiry of the site's signed CDN URLs); falls back to the signed `subtitles.webvtt` CDN URL when inline content is absent.
- Follows the `youtube-page` pattern (fresh fetch per request plus a video-id guard) rather than `inferTracks`, because all videos share the same pathname (`/spanish/watch`, `/french/watch`) with identity in the query string, and the site soft-navigates/auto-advances between videos.
- Publishes no tracks when the API reports `hasAccess: false`.
- Unit tests cover the API-response→track mapping (`extension/src/pages/dreaming.test.ts`).

## Testing

- Extension, common, and client jest suites pass; eslint and prettier clean.
- Verified end-to-end against the live site: injected the built page script into a free video's watch page and dispatched `asbplayer-get-synced-data` — received basename "Alma's Family" and a `Spanish`/`es`/`vtt` track whose data URL decodes to the correct UTF-8 WebVTT.
- Premium videos use the same API shape (gated by `hasAccess`) but were not tested with a premium account.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


killergerbah:
@ravensgulls Thank you, and sorry it took so long to look at this. It looks good to me.

Before I can merge, can you do these things?

1. Merge main into your branch
2. Read the `CONTRIBUTORS` file and yourself to it if you agree

Issue #1062, opened 2026-07-05
First play of an unrecorded subtitle is silent (tab shows audio playing) and can't be paused, on Firefox


KaiExec:
**What is the issue and what do you do to make it happen**
Mining a subtitle for the first time (i.e. the clip hasn't been recorded yet), using a local video file loaded with two SRT subtitle files (dual subtitle mode): clicking play in the mining dialog's audio preview produces no audible sound, but the Firefox tab's "playing audio" (speaker/mute) indicator lights up. The clip also cannot be paused mid-playback. Once that first playback finishes, clicking play again works normally — audible, and pausable.

I've only tested this with local video + two subtitle files loaded. I haven't verified whether this also happens with streaming video (e.g. mining directly from a YouTube tab) or with only one subtitle file loaded.

**What browser?**
Firefox

**Is it happening with local video or streaming video?**
Local video

**Root cause (from reading `common/audio-clip/audio-clip.ts`)**

- On first play, `FileAudioData.play()` has no cached blob yet, so it falls through to `this._blobClipper.clip(true)` — a live re-play-and-capture pass: it plays the segment audibly while simultaneously recording it via `captureStream()`/`mozCaptureStream()` + `MediaRecorder`. Every subsequent play instead uses the already-produced `Blob` via a much simpler `new Audio()` playback path. This is why the first and second playbacks behave completely differently.
- The `playing` getter used (presumably) by the dialog's pause control only checks `_playClipper`:
  ```js
  get playing() {
      return this._playClipper?.isPlayingAudibly ?? false;
  }
  ```
  `_playClipper` is never created during the initial `_blobClipper.clip(true)` pass, so `playing` incorrectly reports `false` the whole time the first live recording is in progress — which looks like it's why pause doesn't work during that first pass.
- `FileAudioClipper.isPlayingAudibly` has an explicit Firefox carve-out:
  ```js
  get isPlayingAudibly(): boolean {
      return (
          (!isFirefox && this._clippingAudioElement !== undefined && this._clippingAudibly === true) ||
          this._playingAudioElement !== undefined
      );
  }
  ```
  suggesting this pathway is already known to behave differently on Firefox.

**Suggested direction**
Have `FileAudioData.playing` also reflect `_blobClipper`'s in-progress clipping state (not just `_playClipper`), so pause works correctly during the first, live "record it now" pass on Firefox.

I haven't been able to confirm whether the exported/final audio clip itself is actually silent or complete — happy to test if useful.

[...]

PR #1133, opened 2026-08-18
Fix silent audio on first play and pause behavior in Firefox

aramrw:
Fixing #1062 

Assisted-by: Gemini:gemini-3.7-flash

### Summary of Changes

Fixes an issue where playing an unrecorded audio clip for the first time in the popup dialog produces silent audio in Firefox, and mid-playback pause does not work.

### Root Cause
1. **Firefox Mutes Media Elements on Stream Capture**: When `captureStream()` / `mozCaptureStream()` is called on an `<audio>` element in Firefox during the first live clipping pass, Firefox redirects audio into the stream and detaches/mutes the audio output from the speakers.
2. **Audio State Tracking**: `isPlayingAudibly` had a `!isFirefox` check, and `FileAudioData.playing` only inspected `_playClipper?.isPlayingAudibly` without including `_blobClipper.isPlayingAudibly`. This caused `playing` to falsely report `false` during live clipping, preventing the pause action from functioning during the first pass.

### Solution
- Connected `createMediaElementSource(audio)` to `audioContext.destination` in Firefox during audible clipping passes to ensure audio routes to the speakers while being recorded, and closed the `AudioContext` upon stop/completion.
- Removed the `!isFirefox` restriction from `FileAudioClipper.isPlayingAudibly`.
- Updated `FileAudioData.playing` to track `this._blobClipper.isPlayingAudibly || this._playClipper?.isPlayingAudibly`.

Tested on youtube and self hosted the video player to confirm fixes.

### DIFF
--- a/common/audio-clip/audio-clip.ts
+++ b/common/audio-clip/audio-clip.ts
@@ -207,6 +207,7 @@ class FileAudioClipper {
     private readonly _recorderMimeType: string;
     private readonly _trackId?: string;
     private _clippingAudioElement?: HTMLAudioElement;
+    private _clippingAudioContext?: AudioContext;
     private _clippingAudibly?: boolean;
     private _clippingAudioReject?: (error: Error) => void;
     private _stopClippingTimeout?: ReturnType<typeof setTimeout>;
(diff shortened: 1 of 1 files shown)


killergerbah (approved):
Nice work, thank you

PR #1144, opened 2026-08-22
Add opt in generic parser for unsupported pages

ShanaryS:
The goal is to somewhat mitigate the need for specific site implementations by using strategies blindly on pages that have been proved to be useful. Even if this generic parser works though, site specific integrations are still necessary for handling site specific settings or properly parsing titles.

- On the `VideoDataSyncDialog`, we now have a new option to enable generic parsing for this site
  - This is only available if there isn't a specific page script for this site
  - An info message is displayed explaining this new option as well as a warning message for when generic parsing is enabled
- There is also a second option to enable aggressive parsing
  - Base: Conservative inspection of DOM tracks, cues, explicit inline metadata, and known subtitle URLs
  - Aggressive: Hooks into `fetch`, `XHR`, and `JSON.parse` to find anything that resembles subtitles
- The configuration for the sites is stored in global state
  - Keyed by host, right now only if generic is enabled and if aggresive is to be used
- The generic parser works by attaching to videos directly instead of the document. We also have some restrictions in case there is multiple videos as we can't reliably know what corresponds to what since it's for arbitrary sites
- `autoSync` is always disabled for the generic parser. I think this is useful at least intially for safety.
- Since pages without a `pageScript` now use the generic parser, `PageDelegate` is always defined
- Some changes were made to common functions around parsing subtitles (including extensions, m3u8, mpd, etc) to make them more robust against unknown data
- Tests were added for the generic page as we can verify our parsing without needing to test against live pages. Specific site integrations are bespoke and would only be useful to test with the live site but here we can verify the parsing of the formats we are checking.
- Parsing is done with safety at every level, everything is bounded and every value is explicitly checked with a best effort for tolerating failures

Of the currently supported sites, I found this successfully detected subtitles for (delete `pageScript` from their `pages.json`):
- `YleAreena`, `Natural Japanese`, `SVTPlay`, `URPlay`

I'm sure there is more but I just stopped at these since they were easy to check. For sites not currently supported, some that now can be detected are:
- `Tubi.tv`, `TED.com`, `Arte.tv`, `3cat.cat`

- [x] Test on 50 popular video sites to base stated efficacy of the parser

There is obviously thousands more sites that we can autodetect subtitles from now so this should definitely be helpful everyone.


extension/src/pages/m3u8-util.ts
```diff
@@ -46,40 +63,38 @@ export function subtitleTrackSegmentsFromM3U8(url: string): Promise<VideoDataSub
                     }
 
                     for (const label of Object.keys(group)) {
-                        if (label.includes('--forced--')) {
```
killergerbah, line 49: Why was this removed? It looks like the following line implements the same behavior, but with the overhead of defining `fetchTrack` and without the long comment

```
                            if (!label.includes('--forced--')) promises.push(fetchTrack());
```

extension/src/pages/mpd-util.ts
```diff
@@ -12,48 +12,65 @@ export interface Playlist {
...
-    const manifest = await (await originalFetch(mpdUrl)).text();
+    manifest: string,
+    trackExtractor: (
+        playlist: Playlist,
+        language: string,
+        metadata?: MpdTrackMetadata
+    ) => VideoDataSubtitleTrackDef | undefined
+): VideoDataSubtitleTrack[] => {
     const parsedManifest = parse(manifest, { manifestUri: mpdUrl });
+    const metadataByRepresentationId = subtitleMetadataByRepresentationId(manifest);
```
killergerbah: Why go around `parsedManifest`? Isn't this parsing it twice?

extension/src/services/pages.ts
```diff
@@ -103,21 +112,53 @@ async function pageConfigsMergedWithSettingsOverrides(): Promise<PageConfigFile>
...
+    { tutorial, genericSubtitleParserEnabled, aggressiveGenericSubtitleParserEnabled }: PageDelegateOptions
+): PageDelegate {
+    const genericPageConfig = (page?: PageConfig): PageConfig => ({
+        ...page,
+        host: page?.host ?? urlObj.host,
+        pageScript: aggressiveGenericSubtitleParserEnabled ? aggressiveGenericPageScript : genericPageScript,
+        generic: true,
+        searchShadowRootsForVideoElements:
+            page?.searchShadowRootsForVideoElements ?? aggressiveGenericSubtitleParserEnabled,
+        syncAllowedAtPath: page?.syncAllowedAtPath ?? '.*',
```
killergerbah: Why is `.*` better than `undefined`? If it's preferred maybe just make `syncAllowedAtPath` a `string` instead of `string | undefined`?

common/global-state/index.ts
```diff
@@ -20,10 +20,22 @@ export interface OnlineSubtitleSourceConfig {
     jimakuRecentWorks?: JimakuCachedWork[];
 }
 
+export interface GenericSubtitleParserState {
+    pages: {
+        [host: string]: {
+            enabled: boolean;
```
killergerbah: Nitpick, but with two booleans 4 states are possible when you can really only have 3: disabled, enabled, aggressive enabled

extension/src/ui/components/VideoDataSyncUi.tsx
```diff
@@ -281,6 +302,31 @@ export default function VideoDataSyncUi({ bridge }: Props) {
         },
         [bridge]
     );
+    const handleGenericSubtitleParserEnabledChange = useCallback(
```
killergerbah: After enabling I wonder if we should suggest that the user refresh for the setting to take effect

extension/src/pages/subtitle-discovery.ts
```diff
@@ -0,0 +1,299 @@
...
+    if (videoTitle) return videoTitle;
+
+    const openGraphTitle = document.querySelector<HTMLMetaElement>('meta[property="og:title"]')?.content.trim();
+    return openGraphTitle || document.title;
+}
+
+export function bindVideoDataDiscovery(discovery: VideoDataProvider, eventTarget: Document = document): () => void {
+    const requestGenerations = new WeakMap<HTMLVideoElement, number>();
+    const listener = (event: Event) => {
+        const video = event.composedPath().find((target) => target instanceof HTMLVideoElement);
```
killergerbah, line 283: How do you know that this video is the right one? Is the right video the one that has an asbplayer binding?

extension/src/pages/generic-page.ts
```diff
@@ -0,0 +1,307 @@
...
+    }
+}
+
+export class GenericPageDiscovery implements VideoDataProvider {
+    async videoData(video: HTMLVideoElement): Promise<VideoData> {
+        const tracks = nativeSubtitleTracks(video);
+        const videos = Array.from(document.querySelectorAll('video'));
+
+        // Inline metadata is page-scoped. Use it only when its association with the requested video is unambiguous.
+        if (videos.length === 1 && videos[0] === video) {
```
killergerbah: Why does AggressiveGenericPageDiscovery ignore video count but GenericPageDiscovery requires exactly 1? Is there a way to use the video that the user actually bound to?

extension/src/pages/m3u8-util.ts
ShanaryS: Originally it was storing some data for all tracks but I dropped it as it wasn't doing much and I didn't move the check back. I thought the check of excluding forced was trivially clear but now I agree the comment should be kept.

extension/src/pages/mpd-util.ts
ShanaryS: There is some data that's lost that we need for the aggressive parser like the mimeType but I think we can just expose that so it's clearer with a single parse.

common/global-state/index.ts
ShanaryS: Yeah that simplifies things, I'm already setting both at the same time in a lot of places. As a side note if you prefer a different name over aggressive I'm open to it. I didn't spend much time on the naming but I feel aggressive makes it clear that it's slightly dangerous and should only be used when the site needs it.

extension/src/ui/components/VideoDataSyncUi.tsx
ShanaryS: The info and warning message both say that refreshes might be needed, I was trying to cut down on loc entries. I also feel it's fairly obvious a refresh would be needed but that might just be my dev brain. But I think a specific alert is worth the entry.

I also am undecided on exactly where to put the notifications. I notice on some pages the subtitle picker needs to scroll and so the current alerts are off screen.

extension/src/pages/subtitle-discovery.ts
ShanaryS: This function is called directly from Binding with the event so should always be correct.

extension/src/pages/generic-page.ts
ShanaryS: I think this would require more work coordinating as we can't write an logic to identify the current video for arbitrary pages. I think the side effect of ignoring video count only means that we could have unrelated subtitles on any particular video which is part of the warning and so didn't seem necessary to prevent. But in my testing so far (plan to test a lot more websites when finalized and potentially add more generic parsing if possible), I've yet to a see a site with multiple video elements so maybe just limiting to 1 would be fine without trying to figure it out. At the cost of what seems like very few sites.

extension/src/services/pages.ts
ShanaryS: Yeah undefined is clearer, the match everything regex is just unnecessary work I thought made a difference.

extension/src/pages/subtitle-discovery.ts
killergerbah: I see now, I misunderstood with `event.composedPath()...` was doing. It looks like it should be matching the `video` element event target in `VideoDataSyncController`. I also now understand why you switched to using `video` for the event target rather than `document`.

extension/src/pages/generic-page.ts
killergerbah: I think it would be fine to ignore the count and let the user pick the video element that's 'correct' as that should be the only reliable way to pick the video element for generic pages. It seems some sites have a video element for ads so I think this could make a significant difference.

common/global-state/index.ts
killergerbah: I can't think of anything better than 'aggressive' right now but will think about it

extension/src/ui/components/VideoDataSyncUi.tsx
killergerbah: I see what you mean. I guess for one the space between the switches at the bottom can be reduced. And there may be a way to combine the alert with the switch. Maybe doesn't need to be done in this PR.

common/global-state/index.ts
killergerbah: Out of curiosity, are there any sites where the non-aggressive parser works?

ShanaryS: Yes, Natural Japanese, TED, and at least one of the PRs. So I would say like 40-50% of sites that work can detect without it. Maybe Tubi and 3Cat as well.

If you want we can get rid of the base enabled (but still keep the code separation).

killergerbah: Oh no let's keep it, it seems to work quite often although it's more than I thought

extension/src/pages/generic-page.ts
```diff
@@ -0,0 +1,302 @@
...
+    const displaySeconds = totalSeconds % 60;
+    const totalMinutes = Math.floor(totalSeconds / 60);
+    const minutes = totalMinutes % 60;
+    const hours = Math.floor(totalMinutes / 60);
+    return `${hours.toString().padStart(2, '0')}:${minutes.toString().padStart(2, '0')}:${displaySeconds
+        .toString()
+        .padStart(2, '0')}.${milliseconds.toString().padStart(3, '0')}`;
+}
+
+function webVttFromTextTrack(track: TextTrack): string | undefined {
```
killergerbah: Nitpick, you could use a library function to convert to srt instead

extension/src/pages/m3u8-util.ts
```diff
@@ -57,20 +74,25 @@ export function subtitleTrackSegmentsFromM3U8(url: string): Promise<VideoDataSub
...
+                                    (segment: any) => new URL(segment.uri, loadedManifest.url).href
+                                );
+                                const rawExtension = extractExtension(urls[0], 'vtt').toLowerCase();
                                 return trackFromDef({
                                     label: label,
                                     language: track.language,
                                     url: urls,
-                                    extension: extractExtension(subManifest.segments[0].uri, 'vtt'),
+                                    extension:
+                                        rawExtension === 'xml'
```
killergerbah, line 94: Is the `xml` -> `ttml2` mapping a special case for `m3u8` manifests specifically?

extension/src/services/binding.ts
```diff
@@ -786,9 +787,10 @@ export default class Binding {
                 disneyPlus ? 1000 : 0
             );
             this.videoChangeListener = () => {
+                const mediaIdentity = mediaSourceIdentity(this.video);
+                const sourceChanged = !Object.is(mediaIdentity, this._lastLoadedMetadataMediaIdentity);
+                this._lastLoadedMetadataMediaIdentity = mediaIdentity;
```
killergerbah, line 792: Does this change matter for page integrations other than emby/jellyfin?

extension/src/pages/m3u8-util.ts
ShanaryS: Yes, this is a best-effort heuristic as otherwise we will not be able to parse them since there is no xml parser.

I'm not sure if this ever matters but that's the issue with this PR. It's impossible to know the full extent of what decisions affect the efficacy since it's blind. The only way to address that is by seeing when the generic parser fails and why but so far from my testing seems every decision in this PR has been proven to be the right tradeoff.

extension/src/services/binding.ts
ShanaryS: No and I've confirmed it still works for jellyfin. I decided against using src changes means new video for generic parser as that could have disastrous consequences on some sites. I believe the only downside is needing to reload the pages on genuine new videos but maybe it's affected by ads too.

killergerbah (approved):

PR #1145, opened 2026-08-22
Add resuming for auto pause and hiding subtitles during playback

### LINKED ISSUE #832: MvJ Primed Listening Playback Mode
It would be great if asbplayer, as a great FOSS tool, added support for a new [MvJ](https://www.skool.com/mattvsjapan/about) primed listening mode. The idea behind it is to organize video playback as follows:
1. You're watching TL (Target Language) video content with NL (Native Language) subs.
2. Whenever a new line appears, it's shown.
3. Video is paused for 1-4s (depending on a line length) for you to read.
4. You read the NL-sub, which preloads the meaning into your brain, specifically, short-term memory.
5. The subtitle line is hidden.
6. After an extra 300ms pause playback resumes and you hear the TL audio.
7. Brain automatically maps the meaning to the TL-audio and acquires the language.

Originally, a similar idea was implemented as an mpv script by an MvJ community member.  Since it's not publicly available, I've decided to generate an [open-source mpv script](https://github.com/JayXT/MvJPrimedListening/tree/main) that follows the approach outlined above and works fine for offline playback. It seems that MattVsJapan also cooperated with Migaku and they added a similar feature there. All of that made me think that having a good FOSS alternative would be beneficial to community. And unlike Migaku, asbplayer is capable of supporting this mode not only on Netflix and YouTube, but on numerous other services, which makes it another competitive advantage.

JayXT:

Closes #832. Supersedes #1124, which implemented the same request as a dedicated playback mode — this takes @ShanaryS's suggestion from that review and builds the behavior out of generalized settings instead.

## What changes

Auto-pause could only wait for the user, and subtitles were always on screen. Both become settings in their own right:

| Setting | Values |
| --- | --- |
| `autoPauseResumeMode` | `manual` \| `fixed` \| `subtitleLength` |
| `autoPauseResumeDelayMs` | silence after the subtitle is hidden |
| `autoPauseFixedDurationMs` | pause length in `fixed` |
| `autoPauseMinimumDurationMs` / `autoPauseMaximumDurationMs` | bounds in `subtitleLength` (`0` max = no limit) |
| `autoPauseTimePerCharacterMs` | reading time per character |
| `subtitleVisibility` | `always` \| `whilePaused` |

Defaults are `manual` and `always`, which reproduce current behavior exactly.

## Primed listening

The request in #832 is auto-pause at the start of a line, resuming by subtitle length, with subtitles withheld during playback — no mode of its own. A resuming pause runs in two phases: the subtitle is shown for the pause duration, then hidden while playback stays paused for the resume delay, so the audio is heard with nothing on screen.

Visibility therefore follows the pause duration rather than whether the media happens to be paused. A literal reading of `whilePaused` would keep the subtitle up through the resume delay and remove the gap the feature depends on.

`Toggle primed listening` (`shift+L`) switches auto-pause, its pause edge, resume mode and subtitle visibility together, and restores the previous values on the second press — keeping a resume mode you had already configured, and leaving auto-pause on if it already was. It is a convenience over the settings, not a mode: each part is usable alone, so auto-resume without hiding subtitles, hidden subtitles without auto-resume, or `atEnd` for hear-then-read all work.

## Notes on the implementation

- Which edges pause is already encoded in the timeline blocks, so only the policy blocks cannot express reaches `PlaybackPlan`.
- `AutoPauseController` owns the phases. It ends a pause only for a user seek, since auto-pause corrects its position by seeking immediately after pausing.
- Pause length is measured from subtitles on seekable tracks — the same tracks that drive other playback modes — so a translation track loaded beside the target language does not inflate it. Only text that is shown counts; markup and ruby readings do not.

## Testing

`yarn run verify` passes. Adds unit coverage for the pause phases and duration rules, and playback engine tests for the two-phase pause, the shortcut's apply/restore, and a pause surviving an auto-pause correction seek. Documentation under `docs` is updated.


common/playback/controllers/auto-pause-controller.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+export interface AutoPausePolicy {
+    readonly resume?: PlaybackPlanAutoPauseResume;
+    readonly subtitlesWhilePausedOnly: boolean;
+}
+
+// Ruby readings are an annotation above the base text rather than something to read, and markup is
+// only kept in subtitle text when it is rendered rather than stripped at parse time.
+const rubyReadingPattern = /<rt\b[^>]*>[\s\S]*?<\/rt>/gi;
```
ShanaryS: I don't think we should be handling this here. Text shouldn't contain ruby so I think a change where we strip it when loading the subtitles and only storing it in `SubtitleModel.tokenization` I think is the right design.

But I'll just tackle that change later after this PR.

common/playback/playback-engine.ts
```diff
@@ -240,11 +278,15 @@ export default class PlaybackEngine<T extends IndexedSubtitleModel> {
             },
             onDiscontinuity: (currentTimestampMs) => {
                 this.playbackPositionController.discontinuity(currentTimestampMs);
-                this.executor.handleDiscontinuity(currentTimestampMs);
+                const { cause } = this.executor.handleDiscontinuity(currentTimestampMs);
+                // Auto-pause corrects its position by seeking right after pausing, so only a user
+                // seek ends a pause that is still showing its subtitle.
+                if (cause === 'user-seek') this.autoPauseController.cancel();
```
ShanaryS: I don't think we can always cancel on every user seek. We need to pass in the paused state and if paused `this.showSubtitles()` instead of `this.hideSubtitles()`. Or make create a new function `userSeeked()` to use here instead of changing `cancel()`.

common/playback/playback-engine.ts
```diff
@@ -160,13 +185,24 @@ export default class PlaybackEngine<T extends IndexedSubtitleModel> {
...
 
         const executorCallbacks: PlaybackPlanExecutorCallbacks = {
             play: callbacks.play,
             paused: () => this.timingDriver.paused(),
             pause: () => {
                 callbacks.pause();
-                void this.playbackPositionController.savePlaybackPosition(this.timingDriver.currentTimeMs());
+                const timestampMs = this.timingDriver.currentTimeMs();
+                void this.playbackPositionController.savePlaybackPosition(timestampMs);
+                this.autoPauseController.autoPaused(this.playbackModeSubtitlesAt(timestampMs));
```
ShanaryS: I think we need to have `pause()` take in the subtitles that caused the pause rather than reading it after. This seems like it can cause race conditions which should be easily avoidable by passing around explicit state.

ShanaryS (commented):
I like this design much more but I still think it should be more generalized. The settings are but primed listening is still being treated as a special state rather than just a natural consequence of the user's settings.

I don't want the precedent of future configurations getting special logic as that will get quickly complex. I also think maintaining the previous state before toggling primed listening is genuinely convenient but not worth the design complexity.

I think instead we don't have a primed listening keybind or any special logic in playback engine for handling it. We have two keybinds instead, one to cycle through `autoPauseResumeMode` and another to toggle `subtitleVisibility`. I also think these should not have a default keybind as they are fairly niche.

This means that you can still quickly swap to primed listening and regular auto pause with a maximum of 3 presses across two keybinds, or with a single keybind if you never use auto pause for anything else. This also allows using the keybinds for non primed listening usages as well.

The playback mode logic I feel will constantly be expanded with new novel strategies and this ensures reusability and simplicity at the cost of some minor UX hit.

common/playback/controllers/auto-pause-controller.ts
JayXT: FYI, the logic has been updated.

common/playback/playback-engine.ts
JayXT: FYI, the logic has been updated.

common/playback/playback-engine.ts
JayXT: FYI, the logic has been updated.

ShanaryS (approved):

killergerbah (approved):
Tested on Disney Plus, looks good

PR #1157, opened 2026-08-27
Improvements for the generic parser informed by real world tests

ShanaryS:
As mentioned here, https://github.com/asbplayer/asbplayer/pull/1144#issuecomment-5418075326, I went and checked 200+ sites for this feature to get some proper statistics and find any other meaningful improvements. I was able to resolve some of the failures as well as some UX improvements. I think this PR essentially puts this parser in its final state as there are no more generic improvements we can make from my extensive collection.

But first the statistics (as of this PR):

| Result | Sites | Rate |
| --- | ---: | ---: |
| **Base works without selection** | 23 / 86 | **26.7%** |
| **Base works overall** | 54 / 86 | **62.8%** |
| **Base requires subtitle selection** | 31 / 86 | **36.0%** |
| **Aggressive works without selection** | 56 / 86 | **65.1%** |
| **Aggressive works overall** | 73 / 86 | **84.9%** |
| **Aggressive requires subtitle selection** | 17 / 86 | **19.8%** |
| **Aggressive-only rescue** | 19 / 86 | **22.1%** |
| **Fail even with Aggressive** | 13 / 86 | **15.1%** |

The key number is that `~85%` of tested sites (checked 200+) with subtitle detection possible is discovered by the generic parser (far greater than my original estimate of 25%-50%). I was originally hoping for a number ~80% when first developing this feature so I'm happy with where it landed up.

The checked sites covered the most popular video based websites for numerous languages and media types which should be a representative sample of the contexts where asbplayer could be used. The final number for the stats is much lower due to a lot of sites either being hardcoded subs, no subs, or requiring an account or subscription (which shouldn't affect the population distribution for this test).

My intention for this feature still stands from the comment linked above:
> Yes I agree, this should never get in the way of these integrations, if anything that should mean they are relatively easy integration. Even if someone were to request a site with perfect base generic support (for example Natural Japanese), it's still worth adding in my opinion.

> I don't think it's a good value of anyone's time to add integrations without explicit requests, so this serves cater to those cases. I do plan on adding specific integrations for many sites when it becomes the most valuable thing I can do, but right now there are much bigger priorities for asbplayer especially after this change.

Otherwise for the changes:
- Update the docs with the `~85%` efficacy as well as the statement that it doesn't preclude specific integrations
- Added a message to refresh the page when changing the generic parser (at least one user would otherwise miss this)
  - I made this text as generic as possible so it can potentially be reused in the future
- Opening the picker when for a generic page will trigger a reload of the subtitle data
  - If new data is present, the subtitle picker dialog will be refreshed
  - This is necessary for the sites which only work by selected subtitles. Most will preserve the subtitle selection through a page refresh, but some don't and this gives better UX where they can just select the subtitle and open the picker.
- All improvements were pretty much in aggressive mode, both modes gained some improvement in metadata parsing however

### DIFF
--- a/extension/src/controllers/video-data-sync-controller.ts
+++ b/extension/src/controllers/video-data-sync-controller.ts
@@ -375,7 +375,7 @@ export default class VideoDataSyncController {
         const wasLoading = this._syncedData?.subtitles === undefined;
         this._syncedData = data;
 
-        if (this.pickerVisible && !wasLoading) {
+        if (this.pickerVisible && !wasLoading && (await currentPageDelegate()).config.refreshSubtitleDataOnPickerOpen) {
             const subtitleIds = data.subtitles?.map((track) => track.id);
             if (!arrayEquals(previousSubtitleIds, subtitleIds)) {
                 this._frame.clientIfLoaded?.updateState({
--- a/extension/src/controllers/video-data-sync-controller.ts
+++ b/extension/src/controllers/video-data-sync-controller.ts
@@ -93,6 +93,7 @@ export default class VideoDataSyncController {
     private _fullscreenElement?: Element;
     private _activeElement?: Element;
     private _autoSyncAttempted: boolean = false;
+    private _refreshingOpenPicker: boolean = false;
     private _dataReceivedListener?: (event: Event) => void;
     private _dataReceivedEventTarget?: EventTarget;
     private _isTutorial: boolean;
@@ -134,6 +135,7 @@ export default class VideoDataSyncController {
         this._dataReceivedListener = undefined;
         this._dataReceivedEventTarget = undefined;
         this._syncedData = undefined;
+        this._refreshingOpenPicker = false;
         this._cleanupPlayBlocker();
         this._openedLocation = undefined;
     }
@@ -191,9 +193,12 @@ export default class VideoDataSyncController {
             return;
         }
 
-        if (!refreshOpenPicker) {
+        if (refreshOpenPicker) {
+            this._refreshingOpenPicker = true;
+        } else {
             this._syncedData = undefined;
             this._autoSyncAttempted = false;
+            this._refreshingOpenPicker = false;
         }
 
         const eventTarget = pageDelegate.config.generic ? this._context.video : document;
@@ -375,7 +380,7 @@ export default class VideoDataSyncController {
         const wasLoading = this._syncedData?.subtitles === undefined;
         this._syncedData = data;
 
-        if (this.pickerVisible && !wasLoading && (await currentPageDelegate()).config.refreshSubtitleDataOnPickerOpen) {
+        if (this._refreshingOpenPicker && this.pickerVisible && !wasLoading) {
             const subtitleIds = data.subtitles?.map((track) => track.id);
             if (!arrayEquals(previousSubtitleIds, subtitleIds)) {
                 this._frame.clientIfLoaded?.updateState({
@@ -571,6 +576,7 @@ export default class VideoDataSyncController {
     private _hideAndResume() {
         this._cleanupPlayBlocker();
         this._openedLocation = undefined;
+        this._refreshingOpenPicker = false;
         this._context.keyBindings.bind(this._context);
         this._context.subtitleController.forceHideSubtitles = false;
         this._context.mobileVideoOverlayController.forceHide = false;
--- a/extension/src/controllers/video-data-sync-controller.ts
+++ b/extension/src/controllers/video-data-sync-controller.ts
@@ -1,4 +1,4 @@
-import { asbError } from '@project/common/util';
+import { arrayEquals, asbError } from '@project/common/util';
 import type {
     ActiveProfileMessage,
     ConfirmedVideoDataSubtitleTrack,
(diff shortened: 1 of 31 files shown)


extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -363,9 +371,23 @@ export default class VideoDataSyncController {
     }
 
     private async _setSyncedData(data: VideoData) {
+        const previousSubtitleIds = this._syncedData?.subtitles?.map((track) => track.id);
         const wasLoading = this._syncedData?.subtitles === undefined;
         this._syncedData = data;
 
+        if (this.pickerVisible && !wasLoading) {
```
killergerbah: This looks like it will bypass auto-sync for generic pages. Is that intentional?

If we want to keep auto-sync then you could move this block to the bottom of the function:

```

        } else if (!this.pickerVisible || wasLoading) {
            this._frame.clientIfLoaded?.updateState(await this._buildModel({}));
        } else {
           // new block
        }
```

killergerbah (commented):
Great work, how did you come up with the list of sites to look at and how long did it take for you to go through all of them?

extension/src/controllers/video-data-sync-controller.ts
ShanaryS: auto-sync is always disabled for generic pages so it's not a concern currently, I also don't think we should ever enable it since it often gets non-subtitles identified on a few sites. So it would have to have a mechanism to enable auto-sync per site but it doesn't seem worth it.

This check preventing auto-sync I think is fine, since it prevents consuming it for this call. And I don't think auto-sync doesn't seem possible with the picker open anyways. But there is a bug where it would run even when `refreshSubtitleDataOnPickerOpen: false` which could be problematic so I added a check for it.

extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -166,7 +167,7 @@ export default class VideoDataSyncController {
         return this._openedLocation;
     }
 
-    async requestSubtitles({ videoChanged }: RequestSubtitlesOptions) {
+    async requestSubtitles({ videoChanged, refreshOpenPicker = false }: RequestSubtitlesOptions) {
```
killergerbah: `refreshOpenPicker` has the same value as `pageDelegate.config.refreshSubtitleDataOnPickerOpen` but farther down the stack this same class queries `pageDelegate.config.refreshSubtitleDataOnPickerOpen` again. Seems like the 'refresh' behavior should be backed by either the parameter or the page delegate, not both.

I also wonder if there's any way to simplify the logic in this class (specifically this method `requestSubtitles` and `_setSyncedData`) as the more booleans we add the more branchy and difficult to read it becomes.

ShanaryS: Yeah I missed that even though I looked for it. I removed the unnecessary check and simplified the logic so it's much more explicit.

extension/src/controllers/video-data-sync-controller.ts
```diff
@@ -363,31 +375,64 @@ export default class VideoDataSyncController {
...
-                    }
-                } else if (wasLoading) {
-                    // Picker is open in loading state. Populate it now that tracks have arrived.
-                    this._frame.clientIfLoaded?.updateState(await this._buildModel({}));
-                }
+        await this._updatePickerAfterDataReceived(wasLoading);
+    }
+
+    private _updateOpenPickerFromRefresh(previousData: VideoData | undefined, data: VideoData): boolean {
+        if (!this._refreshingOpenPicker || !this.pickerVisible || previousData?.subtitles === undefined) {
```
killergerbah, line 390: Is the purpose of returning on `previousData?.subtitles === undefined` so that the calling code falls through to `_updatePickerAfterDataReceived`?

extension/src/pages/aggressive-generic-page.ts
```diff
@@ -140,19 +487,45 @@ function standaloneHlsTrack(manifestUrl: string, manifest: any): VideoDataSubtit
     );
     if (extensions.size !== 1) return;
     const extension: string | undefined = extensions.values().next().value;
-    return extension === undefined ? undefined : trackFromDef({ label: 'Detected subtitle', url: urls, extension });
+    return extension === undefined ? undefined : trackFromDef({ label: detectedSubtitleLabel, url: urls, extension });
+}
+
+function baseTrackScore(track: VideoDataSubtitleTrack) {
+    if (Array.isArray(track.url)) return trackCandidateScore.segmentedTrack;
+    if (track.label.endsWith(capturedDuringPlaybackLabelSuffix)) return trackCandidateScore.accumulatedTextTrack;
```
killergerbah: AI commented on this and I agree:

> Nit — label-sniffing in scoring aggressive-generic-page.ts:495: baseTrackScore detects accumulated tracks by checking if the label ends with (captured during playback). A site with a literal track label ending in that suffix would be mis-scored. Consider a dedicated flag on the track type instead.

Probably better to be explicit since we control `track`

killergerbah (approved):
Amazing work, this feature is a huge leap for asbplayer

extension/src/controllers/video-data-sync-controller.ts
ShanaryS: Yeah, the refresh is only for subsequent loads for if the user selected the subtitle after `loadedmetadata`.

PR #1158, opened 2026-08-28
Fix inconsistent fullscreen and non-fullscreen subtitle styles

### LINKED ISSUE #432: In some websites the subtitle don't show unless you fullscreen the video
What would be the cause of that?

### LINKED ISSUE #668: Firefox Fullscreen Subtitles Issue
**What is the issue and what do you do to make it happen**
In Firefox when I want to fullscreen a video, the subtitles disappear. I don't know if they are out of screen somewhere or gone, but they are not there. I've tested it with chrome and it works there. Sometimes it works the first time when i go fullscreen (after dragging in subtitles), but after exit and going back into fullscreen one or two times the subtitles are gone.

A fix I found: If Firefox is in window mode it works. So what I am doing at the moment when the subtitles disappear is exiting the video fullscreen mode, then switching Firefox from maximize to window mode and after that I click on the fullscreen button from the video. That works 100% of the time and subtitles will always be there.

**What browser?**
Firefox

It's a streaming website (hianime.to)

Latest version 1.9.1

ShanaryS:
[This was reported in discord.](https://discord.com/channels/962412001810849814/962425139058868234/1539273945356836874)

- The fullscreen subtitle class had a default text shadow that the non-fullscreen didn't have
- The transition between fullscreen and non-fullscreen didn't properly update all the classes
- This means that when subtitles were updated (e.g new annotation data), depending if the user was currently fullscreen or not the subtitle appearance would behave differently
- We will now also set explicit defaults for the outline thickness so that user settings are always applied if they want transparency

The first commit fixed the above issue, the second commit fixes some other inconsistencies with fullscreen subtitles. This speculatively closes #432 and closes #668.
- Rerender subtitles when more of its relevant inputs change
- Image width cache could be stale so we store the ratio and perform the final calculation when needed
- Consistent fallbacks for missing settings or track data
- Ensure subtitles are rerendered immediately when a change is needed rather than wait for the interval
- Consistent css defaults between fullscreen and non fullscreen subtitles
- Account for relative offsets such as scrolling or the video changing

I tested the behaviors of subtitles in fullscreen and non fullscreen modes across Chrome and Firefox and found no issues. It handling various resizing, scrolling, window resizing, fullscreen transitions, zooming, embeded videos, etc all with no issues. I did catch a couple in my testing so I may have missed some interactions.

---

The third commit fixes an unrelated issue I just ran into but seems to fit well with this PR. Chrome apparently sometimes strips newlines when they are by themselves between spans. I noticed that `VideoPlayer` no longer had any newlines when annotations were enabled and this was due to the table css being used for the subtitles. I changed it to `inline-block` which doesn't have this issue but it's kind of unbelievable it just started happening (multiple of my devices but not all). This wasn't always the case and downgrading asbplayer all the way to 1.17 did nothing and older chrome browsers still had the issue.

### DIFF
--- a/extension/src/services/element-overlay.ts
+++ b/extension/src/services/element-overlay.ts
@@ -194,12 +194,16 @@ export class CachingElementOverlay implements ElementOverlay {
         const toggle = () => {
             if (document.fullscreenElement) {
                 container.style.setProperty('display', 'none', 'important');
-                this._transferChildren(container, this._fullscreenContainerElement());
+                this._transferChildren(container, this._fullscreenContainerElement(), this.fullscreenContentClassName);
             } else {
                 container.style.display = '';
 
                 if (this.fullscreenContainerElement) {
-                    this._transferChildren(this.fullscreenContainerElement, container);
+                    this._transferChildren(
+                        this.fullscreenContainerElement,
+                        container,
+                        this.nonFullscreenContentClassName
+                    );
                 }
             }
         };
@@ -235,11 +239,19 @@ export class CachingElementOverlay implements ElementOverlay {
                 }
 
                 if (this.nonFullscreenContainerElement) {
-                    this._transferChildren(this.nonFullscreenContainerElement, container);
+                    this._transferChildren(
+                        this.nonFullscreenContainerElement,
+                        container,
+                        this.fullscreenContentClassName
+                    );
                 }
             } else if (!document.fullscreenElement) {
                 container.style.setProperty('display', 'none', 'important');
-                this._transferChildren(container, this._nonFullscreenContainerElement());
+                this._transferChildren(
+                    container,
+                    this._nonFullscreenContainerElement(),
+                    this.nonFullscreenContentClassName
+                );
             }
         };
 
@@ -291,12 +303,12 @@ export class CachingElementOverlay implements ElementOverlay {
         return document.body;
     }
 
-    private _transferChildren(source: HTMLElement, destination: HTMLElement) {
-        if (!source) {
-            return;
-        }
+    private _transferChildren(source: HTMLElement, destination: HTMLElement, contentClassName: string) {
+        if (!source || !destination) return;
+        if (source === destination) return;
 
         while (source.firstChild) {
+            if (source.firstChild instanceof HTMLDivElement) source.firstChild.className = contentClassName;
             destination.appendChild(source.firstChild);
         }
     }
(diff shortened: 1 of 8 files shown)


extension/src/services/element-overlay.ts
```diff
@@ -384,9 +431,56 @@ export class CachingElementOverlay implements ElementOverlay {
         this.fullscreenContainerElement = undefined;
     }
 
-    private _applyContainerStyles(container: HTMLElement) {
+    private _refreshContainerStylesAfterLayout(container: HTMLElement, options: ApplyContainerStylesOptions) {
+        this._applyContainerStyles(container, options);
+        if (this.layoutAnimationFrame !== undefined) cancelAnimationFrame(this.layoutAnimationFrame);
+        this.layoutAnimationFrame = requestAnimationFrame(() => {
+            this.layoutAnimationFrame = undefined;
+            if (container.isConnected) this._applyContainerStyles(container, options);
```
killergerbah, line 441: Why is this second call in the next frame to `_applyContainerStyles` needed? Should there be a comment with the reason?

extension/src/services/element-overlay.ts
```diff
@@ -384,9 +431,56 @@ export class CachingElementOverlay implements ElementOverlay {
...
+            if (container.style.left === '' || container.style.top === '') {
+                container.style.setProperty('visibility', 'hidden', 'important');
+            }
+            return;
+        }
+
+        container.style.removeProperty('visibility');
+
+        let left = rect.left + rect.width / 2;
+        let top =
```
killergerbah, line 467: This value still needs to be clamped by the boundaryies of the viewport e.g. if the video is larger than the screen (or some other better solution if you can think of any). Without this then the overlay elements appear partially offscreen on Netflix (mobile overlay is completely offscreen and the subtitle container appears close to the bottom edge)

Image

extension/src/services/element-overlay.ts
ShanaryS: It's speculative but it's in case the page itself changes the styles.

extension/src/services/element-overlay.ts
ShanaryS: The new calculation should handle it but I don't have such a scenario to test with at hand.

killergerbah: Tested, it works

killergerbah (approved):

PR #1167, opened 2026-09-01
Dev/migrate to pnpm

mwojick:

This does two main things: 
- Migrates from yarn to pnpm, which is much faster. Clean install for me went from 90 sec to 15 sec, and will get even faster once pnpm 12 is stable. Also more secure and feature rich. Also requires you to specify all direct deps in each respective package.json file. No more hoisting which could create a lot of subtle issues.
- Centralizes all project deps in the monorepo under one top level lock file. Now devs just run a single `pnpm i` at the top level to work on all the js projects.

Few caveats:
- I ended up just deleting the lock files in the `docs` and `scripts/loc` folders without trying to salvage them. Needed to update the docusaurus deps so that all of its transitive deps lined up correctly, but seems to work so far.
- Have tested pretty much everything I could locally (client and ext builds/runs, docs app runs, the top level verify passes), but I haven't been able to test the github action workflows which also got migrated.
- The build on Cloudflare is breaking for some reason but I can't view the logs so not sure why.

Assisted by: Claude Sonnet 5


extension/wxt.config.ts
```diff
@@ -31,6 +31,15 @@ const extName = 'asbplayer';
 export default defineConfig({
     modules: ['@wxt-dev/module-react'],
     srcDir: 'src',
+    // The shared `@project/common` package is framework-agnostic and must not receive WXT's
+    // auto-imports. Without this, unimport mis-scans files like common/settings/settings-provider.ts
+    // (a `storage` constructor param) and injects a spurious `import { storage } from 'wxt/utils/storage'`,
+    // which pnpm's strict node_modules can't resolve from common/. (Yarn's hoisting masked this.)
```
killergerbah: Seems like a Claude-like session-specific comment
- Can it be made clearer what it means to a first time reader? What is "unimport?" Some dependency of WXT?
- Since yarn is no longer in the project it probably doesn't need to be mentioned here (git history will show that this commit removed yarn)

docs/docs/contributing.md
```diff
@@ -41,40 +41,36 @@ Documentation is under the `docs` folder in this repository. When adding new fea
 Below are useful commands for development.
 
 ```bash
-# Install yarn
-npm install --global yarn
-
-# Make sure you're on yarn 3
-yarn set version 3.2.0
+# Install pnpm (if not using nix). Other ways to install: https://pnpm.io/installation
+curl -fsSL https://get.pnpm.io/install.sh | env PNPM_VERSION=11.21.0 sh -
```
killergerbah: Should there be instructions or recommendations for a minimum node version?

extension/package.json
```diff
@@ -44,11 +48,13 @@
...
+        "@jest/globals": "^30.0.5",
         "@types/jest": "^30.0.0",
         "@wxt-dev/module-react": "^1.2.2",
         "jest": "^30.0.5",
         "ts-jest": "^29.4.0",
-        "wxt": "^0.20.19",
-        "yarn-upgrade-all": "^0.7.5"
+        "vite": "^8.2.2",
+        "web-ext": "^10.6.0",
+        "wxt": "^0.21.4"
```
killergerbah: `base-generic-page.test.ts` was failing due to a mock import being ignored. AI recommended adding this line to this file:

```
"@qgustavor/srt-parser": "^2.0.0"
```

which seems to fix the issue after removing all the existing `node_modules` directories

killergerbah (commented):
It's working really well. Are you planning to make the Cloudflare change as well?

PR #1185, opened 2026-09-12
Fixes ASB subtitles going invisible in Crunchyroll fullscreen on Firefox

rajpiskala:
## User impact

This bug is a little specific, so I see why no one caught it. Basically, this only happens if:
- (a) you're using Firefox (not Chrome)
- (b) watching on Crunchyroll (not some other website)
- (c) and watching it specifically on full screen (not the minimized view)

In that scenario, when you use ASB player, the subtitles disappear as soon as they enter fullscreen using Crunchyroll's player controls. The video and audio continue normally, and the subtitles return after leaving fullscreen.

This makes fullscreen viewing unusable for people relying on ASB Player subtitles. For example, my usecase is using Japanese subtitles with Yomitan, so I'm forced to watch it in a smaller window instead

## Known affected environment

The failure has been reproduced with this specific combination:

- Crunchyroll's web player
- Firefox
- ASB Player's subtitle overlay
- Fullscreen entered through Crunchyroll's player

The problem does not occur during ordinary windowed playback. I also do not see this issue with Crunchyroll in Chrome or when using ASB Player on other tested video sites in Firefox.

## Reproduction

1. Open a Crunchyroll episode in Firefox.
2. Load an external subtitle file with ASB Player.
3. Confirm that the ASB Player subtitles are visible during windowed playback.
4. Enter fullscreen using Crunchyroll's fullscreen control.

**Expected:** The ASB Player subtitles remain visible over the fullscreen video.

**Actual:** The subtitles disappear until fullscreen is exited.

## Root cause & fix

Crunchyroll makes its Bitmovin player container the browser's fullscreen element. In this layout, ASB Player's existing parent-selection heuristic can fail to find a suitable element and fall back to attaching the subtitle overlay to `document.body`.

Firefox only displays the fullscreen element and its descendants in true fullscreen, so an overlay attached to `document.body` is outside the visible fullscreen subtree.

This change preserves the existing parent-selection heuristic. Only when that heuristic finds no parent does it fall back to the active fullscreen element, provided that element contains the target video. The subtitle overlay therefore remains inside Crunchyroll's visible fullscreen player.

## Testing

- All linting and workspace tests (191 passed)
- I tested this on Chrome on Crunchyroll to ensure there's no regression in the working environment
- I then tested this on Firefox on Crunchyroll to see that it works - which it does!
- Most of this change is also a regression test for this


ShanaryS (approved):
I can replicate this only on Firefox with the side panel enabled. The first fullscreen has the subtitles but the subsequent fullscreens don't.

PR #1195, opened 2026-09-18
Update license to AGPLv3

ShanaryS:


LICENSE
```diff
@@ -0,0 +1,661 @@
...
+get its source.  For example, if your program is a web application, its
+interface could display a "Source" link that leads users to an archive
+of the code.  There are many ways you could offer source, and different
+solutions will be better for different programs; see section 13 for the
+specific requirements.
+
+  You should also get your employer (if you work as a programmer) or school,
+if any, to sign a "copyright disclaimer" for the program, if necessary.
+For more information on this, and how to apply and follow the GNU AGPL, see
+<https://www.gnu.org/licenses/>.
```
killergerbah, line 661: Does the original MIT notice need to be at the bottom of this file as well?

ShanaryS: I don't think it's necessary. The MIT license doesn't have any restrictions or requirements for when it must be displayed so I kept it in the CONTRIBUTORS and About page as I feel those are the most relevant. I think adding it here may cause more confusion that necessary.

mwojick (approved):

killergerbah:
How should we deal with loc contributions? We may need a similar CI check on the 'pull loc' PRs. Worst case if we don't do anything, I guess we would just lose the translations if we decide to move away from GPL in the future.

ShanaryS:
> How should we deal with loc contributions? We may need a similar CI check on the 'pull loc' PRs.

We do not need anything special for Crowdin. asbplayer is a [Public Project](https://support.crowdin.com/terms/#1-definitions) and so is covered by this [license](https://support.crowdin.com/terms/#125-public-projects).

killergerbah (approved):
Would make sense to merge after 1.21 is fully released

PR #1196, opened 2026-09-19
Add alphanumeric Latin to netflixRubyBaseClass

awelotta:

Sometimes alphanumeric characters are used as ruby bases. Consider the following example
Pasted image 20260918044735

The string "11PM" is given the ruby "イレブンピーエム". In the Netflix subtitles these show up, but are not detected by the current ruby detection.

To fix this, I added `A-Za-z0-9` to the netflixRubyBaseClass variable. The result looks like this.
image

Oh, I'm still failing `yarn run verify` because of missing `extension/src/en.ts`. But I managed to build the chrome extension and get the above screenshot.

Probably we should figure out what other characters Netflix actually accepts as RubyBase. I can't really follow what's been said in other issues, but it seems like just about any character can be used and a special invisible character marks the boundary?


awelotta:
Since I built on Windows, I locally modified `page-file-references-are-correct` to replace some line endings: changed line 9 to `for page in $(jq -r ".pages[].script" $CONFIGPATH | tr -d "\\r"); do`. After which I passed all tests:
```
Test Suites: 15 passed, 15 total
Tests:       192 passed, 192 total
Snapshots:   0 total
Time:        24.454 s
```
I think this is the same issue as here: https://github.com/asbplayer/asbplayer/pull/1179#issue-5389266207

